PR #1, opened 2021-10-25
feat(js): automation test 1

shortcuts:
# Summary

JavaScript API client code generation in TypeScript

### Pros:
- Get working code fast
- Usage seems accurate
### Cons:
- DX is not as good as what we offer today
- If we go with a fully generated client, we will need scripts to post-generation tooling (lint, type check etc). 

### Pros:
- Repository is active
- Good documentation
- Allow custom templates, which would improve the tooling (lint, tsconfig, etc.) and allow us to provide our own client
- Provide built-in post processing lint
### Cons:
- Slow (still faster than writing the whole client)
- Verbose
- Unable to use the retry strategy
- Usage is different that what we provide (require auth on every method)

### Pros:
- Repository is active
- Faster than OpenAPI 
### Cons:
- Less configuration than OpenAPI
- No built-in node http generator (only provide axios, fetch)
- Types are not exported by default (TBC)
- Verbose
- Unable to use the retry strategy
- Usage is different that what we provide (require auth on every method)


shortcuts:
### 🔨 The codegen job will run at the end of the CI.

_Make sure your last commit does not contain generated code, it will be automatically pushed by our CI._

PR #485, opened 2022-05-09
feat(specs): crawler APIC-208

bodinsamuel:
## 🧭 What and Why

🎟 JIRA Ticket: APIC-208

### Changes included:

- Only add the specs for the moment so that we can merge it soon and avoid further conflicts

Not sure about the tag grouping since a lot of endpoints are in `crawler/:id` it was really packed but with tags everything feels disconnected since we can't nest them.
Looking at stripe they do not do a better job, e:g https://stripe.com/docs/api/payment_intents

## 🧪 Test

https://deploy-preview-485--api-clients-automation.netlify.app/

Screenshot 2022-05-09 at 18 44 23

```sh
yarn website

BUNDLE_WITH_DOC=true DOCKER=true yarn cli build specs
```


shortcuts:
kiss @bodinsamuel

PR #819, opened 2022-07-12
chore(deps): use bodinsamuel/renovate-automatic-branch

millotp:
## 🧭 What and Why

Uses https://github.com/bodinsamuel/renovate-automatic-branch to remove logic from renovate, and test the action.

### Changes included:

- Fix the regex error
- Add the action
- Remove the custom code

## 🧪 Test

Wait until friday 16h


bodinsamuel (approved):
👀

millotp:
can everyone approve PR's now that it's public ? I don't think you could do that before

bodinsamuel:
I think I always was able to but not merge. You can double check with @shortcuts

shortcuts:
> can everyone approve PR's now that it's public ? I don't think you could do that before

Yup everyone in the company, it has always been the case

PR #820, opened 2022-07-13
chore(deps): fix renovate and manual trigger

millotp:
## 🧭 What and Why

Maybe fix renovate regex, and add manual trigger for the `renovate` workflow

### DIFF
--- a/.github/workflows/renovate.yml
+++ b/.github/workflows/renovate.yml
@@ -3,6 +3,7 @@ name: Renovate
 on:
   schedule:
     - cron: '0 14 * * 5' # At 14:00 on Friday.
+  workflow_dispatch:
 
 jobs:
   renovate:
--- a/.github/workflows/renovate.yml
+++ b/.github/workflows/renovate.yml
@@ -4,6 +4,10 @@ on:
   schedule:
     - cron: '0 14 * * 5' # At 14:00 on Friday.
   workflow_dispatch:
+    inputs:
+      fake_input:
+        description: input needed to satisfy the yaml linter
+        required: false
 
 jobs:
   renovate:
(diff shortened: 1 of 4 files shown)


renovate.json
```diff
@@ -74,7 +74,7 @@
       "description": "Update yarn version (won't work because we need to download the script but gives an alert at least)",
       "fileMatch": ".yarnrc.yml",
       "matchStrings": [
-        "yarnPath: \\.yarn\\/releases\\/yarn-(?<currentValue>.*?)\\.cjs"
+        "yarnPath: \\.yarn\/releases\/yarn-(?<currentValue>.*?)\\.cjs"
```
shortcuts, line 77: We can make it broader here too just to avoid having to handle escapes

```suggestion
        "yarn\/releases\/yarn-(?<currentValue>.*?)"
```

might be enough wdyt

.github/workflows/renovate.yml
```diff
@@ -3,6 +3,7 @@ name: Renovate
 on:
   schedule:
     - cron: '0 14 * * 5' # At 14:00 on Friday.
+  workflow_dispatch:
```
shortcuts, line 6: can you also add it for the release trigger? so we can recommend it instead of running the CLI

shortcuts: also, does it work without providing inputs?

millotp: On the[doc](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch) it does but I think it's just a linter issue, no big deal

shortcuts (approved):

PR #821, opened 2022-07-13
chore(deps): split regex in two

millotp:
## 🧭 What and Why

We can't have the same named capture group twice in a regex, split it in 2.
`/` seems to be allowed: [doc](https://docs.renovatebot.com/configuration-options/#any)

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -57,7 +57,7 @@
         ".github/actions/setup/action.yml"
       ],
       "matchStrings": [
-        "google-java-format/releases/download/v(?<currentValue>.*?)/google-java-format-(.*?)-all-deps",
+        "google-java-format/releases/download/v(?<currentValue>.*?)/google-java-format-",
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -57,7 +57,8 @@
         ".github/actions/setup/action.yml"
       ],
       "matchStrings": [
-        "google-java-format\/releases\/download\/v(?<currentValue>.*?)\/google-java-format-(?<currentValue>.*?)-all-deps"
+        "google-java-format/releases/download/v(?<currentValue>.*?)/google-java-format-(.*?)-all-deps",
```
shortcuts, line 60:
```suggestion
        "google-java-format/releases/download/v(?<currentValue>.*?)/google-java-format-",
```

We can avoid checking here

shortcuts (commented):
(hopefully) looks good :)

shortcuts (approved):

PR #822, opened 2022-07-13
chore(deps): strict regex for package.json and yarn

millotp:
## 🧭 What and Why

In #532 the regex for `package.mustache` and `.yarnrc.yml` are too broad and detect things that don't have versions,
this should help renovate to find only real deps.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -54,11 +54,6 @@ jobs:
           yarn eslint --ext=json .
           echo 'Use yarn fix:json to fix issues'
 
-      - name: Validate renovate.json
-        uses: rinchsan/renovate-config-validator@v0.0.11
-        with:
-          pattern: renovate.json
-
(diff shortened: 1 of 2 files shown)


.github/workflows/check.yml
```diff
@@ -54,6 +54,11 @@ jobs:
           yarn eslint --ext=json .
           echo 'Use yarn fix:json to fix issues'
 
+      - name: Validate renovate.json
```
millotp, line 57: The validator takes at least 30 seconds tu run, on each PR, either we cache the renovate file or remove this, wdyt ?

shortcuts: (alternatively, this could be in a if with a git diff on the renovate but maybe it's overkill)

millotp: yep I agree, we already some form of validation with the schema anyway

PR #829, opened 2022-07-15
feat(java): generic support for hits

millotp:
## 🧭 What and Why

🎟 JIRA Ticket:  [APIC-575](https://algolia.atlassian.net/browse/APIC-575)

Generic support for all models in theory, only works for `hits` and the java client for now.
The goal was to make it as easy as possible to add a new generic, only add `x-is-generic` to a model in the spec and the property will be propagated to all of it's parent, up to the return type of the operation.

Because java erases type at runtime, we need to pass a additional parameters to inform jackson about the type.

### Changes included:

- Create `GenericPropagator` (name subject to discussion, but it rhymes with alligator so it's cool)
- Add parameter `innerType` to generic endpoint

## 🧪 Test

CI


scripts/ci/githubActions/createMatrix.ts
```diff
@@ -82,7 +86,14 @@ async function getClientMatrix(baseBranch: string): Promise<void> {
     const testsOutputBase = `${testsRootFolder}/${getTestOutputFolder(
       language
     )}`;
-    const testsToDelete = `${testsOutputBase}/client ${testsOutputBase}/methods`;
+    const testsToDelete = matrix[language].toRun
+      .map((client) => {
+        const clientName =
+          language === 'php' ? createClientName(client, language) : client;
```
shortcuts, line 92: is `createClientName` incorrect in this case? how come we can't always call it?

scripts/ci/githubActions/createMatrix.ts
```diff
@@ -82,7 +86,14 @@ async function getClientMatrix(baseBranch: string): Promise<void> {
...
+        const extension = getTestExtension(language);
```
shortcuts, line 93: We can use a wildcard to avoid having extra logic wdyt?

generators/src/main/java/com/algolia/codegen/GenericPropagator.java
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  public static void propagateGenericsToOperations(OperationsMap operations, List<ModelMap> models) {
+    for (CodegenOperation ope : operations.getOperations().getOperation()) {
+      for (ModelMap modelMap : models) {
+        CodegenModel model = modelMap.getModel();
+        if (ope.returnType.equals(model.classname) && hasGeneric(model)) {
+          ope.vendorExtensions.put("x-is-generic", true);
+          // we use {{#optionalParams.0}} to check for optionalParams, so we loose the
+          // vendorExtensions at the operation level
```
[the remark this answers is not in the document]
shortcuts: wdym by `we loose`? added `vendorExtensions` at the spec level are lost?

generators/src/main/java/com/algolia/codegen/GenericPropagator.java
```diff
@@ -0,0 +1,119 @@
+package com.algolia.codegen;
```
shortcuts, line 1: Have you tried to have this logic at the spec bundling level to avoid language-specific handling/depending on the base generator structure?

I think it make sense to have this logic here, since it's not related to the REST API itself, but maybe it would've been less complex to read the tree and propagate values at an higher level, wdyt?

generators/src/main/java/com/algolia/codegen/GenericPropagator.java
```diff
@@ -0,0 +1,119 @@
...
+import org.openapitools.codegen.model.ModelMap;
+import org.openapitools.codegen.model.ModelsMap;
+import org.openapitools.codegen.model.OperationsMap;
+
+public class GenericPropagator {
+
+  // Only static use of this class
+  private GenericPropagator() {}
+
+  private static void setIsTrueGeneric(IJsonSchemaValidationProperties property) {
```
shortcuts, line 15: Could we add some descriptions to each steps, so we know what true generic means/don't forget the flow for future maintenance

generators/src/main/java/com/algolia/codegen/GenericPropagator.java
```diff
@@ -0,0 +1,119 @@
...
+    if (property instanceof CodegenModel) {
+      ((CodegenModel) property).vendorExtensions.put("x-true-generic", true);
+    } else if (property instanceof CodegenProperty) {
+      ((CodegenProperty) property).vendorExtensions.put("x-true-generic", true);
+    }
```
shortcuts, line 20: polymorphism might be overkilled here but it's a bit redundant to have to handle those in each step :(

generators/src/main/java/com/algolia/codegen/GenericPropagator.java
```diff
@@ -0,0 +1,119 @@
...
+        (boolean) ((CodegenProperty) property).vendorExtensions.getOrDefault("x-has-child-generic", false)
+      );
+    }
+    return false;
+  }
+
+  private static CodegenModel propertyToModel(Map<String, ModelsMap> models, CodegenProperty prop) {
+    // openapi generator returns some weird error when looking for primitive type, so we filter them
+    // by hand
+    if (prop == null || prop.openApiType.equals("boolean") || prop.openApiType.equals("object") || !models.containsKey(prop.openApiType)) {
```
shortcuts, line 49: `object` should be replaced by the generic in our case, no?

shortcuts (commented):
first pass, looks like it's working for Java, what's the issue on the JS support side?

generators/src/main/java/com/algolia/codegen/GenericPropagator.java
millotp: in mustache if a variable is not found in the scope it will look in the parents scope, except if we use a `.`, il will only look in the current scope, that mean the `vendorExtensions` of `optionalParams.0`, which is empty.
So we need to set the variable in both scope, or set a global variable but I don't know how to do it

scripts/ci/githubActions/createMatrix.ts
millotp: it's the same as [here](https://github.com/algolia/api-clients-automation/blob/main/generators/src/main/java/com/algolia/codegen/cts/tests/TestsRequest.java#L38), I think the file name matters for php-unit

millotp: you're right we shouldn't need the if

scripts/ci/githubActions/createMatrix.ts
millotp: we already had wildcard before that was the issue, we only want to delete the files we will generate

shortcuts: (I mean for the `extension` only, so we delete e.g. `search*`)

generators/src/main/java/com/algolia/codegen/GenericPropagator.java
millotp: you mean doing it in js inside `buildSpecs.ts` ? I didn't think about it, I'm not sure if we have access to the tree like we do here

generators/src/main/java/com/algolia/codegen/GenericPropagator.java
millotp: there are a lot of `object` in our spec, not all of them should be generic, it's too hard to determine just from the name of the type

scripts/ci/githubActions/createMatrix.ts
millotp: ah yes, I'm scared it will break if we have something similar like `algoliasearch` and `search`, or something else

shortcuts: It's just 2 lines anyway it's not a big deal, just pointing it

generators/src/main/java/com/algolia/codegen/AlgoliaJavaGenerator.java
```diff
@@ -101,12 +105,23 @@ public Map<String, ModelsMap> postProcessAllModels(Map<String, ModelsMap> objs)
...
+    GenericPropagator.propagateGenericsToModels(models);
+
     return models;
   }
 
+  @Override
+  public OperationsMap postProcessOperationsWithModels(OperationsMap objs, List<ModelMap> models) {
+    OperationsMap operations = super.postProcessOperationsWithModels(objs, models);
+
+    GenericPropagator.propagateGenericsToOperations(operations, models);
```
shortcuts, line 117: Is it too late here to move the `propagateGenericsToModels` inside `propagateGenericsToOperations`? We have access to both

shortcuts (commented):
Looks correct on the tree traversing side, I feel like we definitely need more comments otherwise we will forget if something break

generators/src/main/java/com/algolia/codegen/GenericPropagator.java
millotp: I tried it and it's a mess because we get the raw yaml, that mean to get the list of properties in a model you have to loop though everything recursively: `properties, items, allOf, oneOf, anyOf...` and it changes based on if it's a property or a model or a response, it would involve lots of custom code (more than in java)

generators/src/main/java/com/algolia/codegen/GenericPropagator.java
millotp: didn't I add that ??

generators/src/main/java/com/algolia/codegen/AlgoliaJavaGenerator.java
```diff
@@ -101,12 +105,23 @@ public Map<String, ModelsMap> postProcessAllModels(Map<String, ModelsMap> objs)
...
+
+    return operations;
+  }
+
   @Override
   public String toEnumVarName(String value, String datatype) {
-    if ("String".equals(datatype)) {
+    if ("String".equals(datatype) && !value.matches("[A-Z0-9_]+")) {
```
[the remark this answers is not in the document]
millotp: it was a miss of one enum: [diff](https://github.com/algolia/api-clients-automation/commit/0c73c11548239fd0970176a7f138e13fbc3d7ffb#diff-7e6492d7ea101b20bb8ac935dce291f4176f54f5b7d489c1e8286ae01ff4d408R12)

generators/src/main/java/com/algolia/codegen/GenericPropagator.java
shortcuts: I feel like `markTrueGeneric` and some `if`s condition (e.g. [diff](https://github.com/algolia/api-clients-automation/pull/829/files#diff-ba08ab9531c5f89ea4087fe8876f22afe20a4d1808159f098e99a208542b49deR66)) could use more description, we rely a lot on the openapi-generator structure that we don't control so it would be best to easily understand each parts if we ever come back to this logic

generators/src/main/java/com/algolia/codegen/GenericPropagator.java
```diff
@@ -0,0 +1,129 @@
...
+import org.openapitools.codegen.model.ModelsMap;
+import org.openapitools.codegen.model.OperationsMap;
+
+public class GenericPropagator {
+
+  // Only static use of this class
+  private GenericPropagator() {}
+
+  /**
+   * Add the property x-true-generic to a model or property, meaning it should be replaced with T
```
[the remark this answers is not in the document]
millotp: True generic means that it can be replaced with `T` directly, and I needed something different than `x-is-generic` otherwise the tree traversal might eat it's own tail in `markTrueGeneric` and go further than wanted.
I didn't actually try it so `x-is-generic` might be enough but I wanted to be sure not to mix things between spec and generation

generators/src/main/java/com/algolia/codegen/GenericPropagator.java
```diff
@@ -0,0 +1,129 @@
...
+    // so we filter them by hand
+    if (prop == null || prop.openApiType.equals("boolean") || prop.openApiType.equals("object") || !models.containsKey(prop.openApiType)) {
+      return null;
+    }
+    ModelsMap map = models.get(prop.openApiType);
+    return map.getModels().get(0).getModel();
+  }
+
+  private static boolean markTrueGeneric(Map<String, ModelsMap> models, IJsonSchemaValidationProperties model) {
+    CodegenProperty items = model.getItems();
```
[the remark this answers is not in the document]
millotp: items is for when it's an array only I think, like `List<Items>`

generators/src/main/java/com/algolia/codegen/GenericPropagator.java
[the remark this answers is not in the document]
shortcuts: Ok make sense! `true` would imply that we encounter false one so maybe just a naming thing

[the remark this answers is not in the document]
millotp: do you have a better name ? 😁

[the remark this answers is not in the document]
shortcuts: Maybe rename `x-is-generic` to `x-is-base-generic` and all of the propagated ones are `x-is-generic`, or call them `x-is-propagated-generic` so the one in the spec remains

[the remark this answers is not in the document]
millotp: the other one are `propagated` too ahah, maybe `x-is-original-generic`, or `x-is-replaceable-generic` ?

generators/src/main/java/com/algolia/codegen/GenericPropagator.java
shortcuts: hm I see, I'd have expected dereferenced specs parsed to JSON be simpler to work on 🤔

generators/src/main/java/com/algolia/codegen/GenericPropagator.java
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  /** Models and their members will be marked with either x-true-generic or x-has-child-generic */
+  public static void propagateGenericsToModels(Map<String, ModelsMap> models) {
+    // We propagate generics in two phases:
+    // 1. We mark the direct parent of the generic model to replace it with T
+    // 2. We tell each parent with generic properties to pass that generic type all the way down
+
+    for (ModelsMap modelContainer : models.values()) {
+      CodegenModel model = modelContainer.getModels().get(0).getModel();
```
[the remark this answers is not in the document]
millotp: I added an assert and never saw a modelContainer with `size != 1` so I guess the assumption is okay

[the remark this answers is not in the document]
shortcuts: Ok! I'd say it's worth adding a comment so we know it's not because we only handle the first element but because there's not more than 1

generators/src/main/java/com/algolia/codegen/GenericPropagator.java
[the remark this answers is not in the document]
shortcuts: IMO only the method and end object (the one in the spec) should be the "base" so `x-is-generic` is good, the rest should be the propagated ones so `x-propagate-generic` or keep the child logic

It's just for the naming so no big deal pick whatever sounds good to your ears, I just felt like `true` was confusing :D

generators/src/main/java/com/algolia/codegen/AlgoliaJavaGenerator.java
millotp: nevermind I need to do it in both methods because the models file are written before calling the `postProcessOperationsWithModels`, so they won't contain the additional `vendorExtensions`

generators/src/main/java/com/algolia/codegen/GenericPropagator.java
```diff
@@ -0,0 +1,158 @@
...
+    // so we filter them by hand
+    if (prop == null || primitiveModels.contains(prop.openApiType) || !models.containsKey(prop.openApiType)) {
+      return null;
+    }
+    return models.get(prop.openApiType);
+  }
+
+  private static boolean markPropagatedGeneric(IJsonSchemaValidationProperties model) {
+    CodegenProperty items = model.getItems();
+    // if the items itself isn't generic, we recurse on it's items and properties until we reach the
```
shortcuts, line 69:
```suggestion
    // if the item itself isn't generic, we recurse on its items and properties until we reach the
```

generators/src/main/java/com/algolia/codegen/GenericPropagator.java
```diff
@@ -0,0 +1,158 @@
...
+        setPropagatedGeneric(model);
+        return true;
+      }
+    }
+    return false;
+  }
+
+  private static boolean propagateGenericRecursive(Map<String, CodegenModel> models, IJsonSchemaValidationProperties property) {
+    CodegenProperty items = property.getItems();
+    // if the items itself isn't generic, we recurse on it's items and properties (and it's
```
shortcuts, line 87:
```suggestion
    // if the item itself isn't generic, we recurse on its items and properties (and its
```

shortcuts (commented):
Small typos, feel free to ignore the suggestion, looks nice!! :D

generators/src/main/java/com/algolia/codegen/GenericPropagator.java
```diff
@@ -0,0 +1,158 @@
...
+   * Add the property x-has-child-generic to a model or property, meaning one of its members is
+   * generic and it should propagate the T
+   */
+  private static void setHasChildGeneric(IJsonSchemaValidationProperties property) {
+    if (property instanceof CodegenModel) {
+      ((CodegenModel) property).vendorExtensions.put("x-has-child-generic", true);
+    } else if (property instanceof CodegenProperty) {
+      ((CodegenProperty) property).vendorExtensions.put("x-has-child-generic", true);
+    }
+  }
```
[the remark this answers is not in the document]
shortcuts: (just thinking of reducing the openapi-generator logic we have here but it's not necessary)

generators/src/main/java/com/algolia/codegen/GenericPropagator.java
millotp: `items` is the term used by openapi, I'll leave it as is

generators/src/main/java/com/algolia/codegen/GenericPropagator.java
millotp: damn I have lots of issues with `it's` and `its`, sorry

generators/src/main/java/com/algolia/codegen/GenericPropagator.java
```diff
@@ -0,0 +1,158 @@
...
+    setVendorExtension(property, "x-propagated-generic", true);
+  }
+
+  /**
+   * Add the property x-has-child-generic to a model or property, meaning one of its members is
+   * generic and it should propagate the T
+   */
+  private static void setHasChildGeneric(IJsonSchemaValidationProperties property) {
+    setVendorExtension(property, "x-has-child-generic", true);
+  }
```
shortcuts, line 37: We don't even need them but I can understand to make it clearer

PR #846, opened 2022-07-18
chore(deps): precise version for yarn and fix java formatter

millotp:
## 🧭 What and Why

Using the recursive regex matcher we can make sure that only version inside `devDependencies` are updated.
Also remove the `v` from java formatter version.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -46,8 +46,10 @@
         "package.mustache"
       ],
       "matchStrings": [
-        "\\s\"(?<depName>.*?)\": \"(?<currentValue>\\d.*?)\",?\\s"
+        "\"devDependencies\":\\s*{[^}]*}",
+        "\\s\"(?<depName>.*?)\": \"(?<currentValue>.*?)\",?\\s"
       ],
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -46,8 +46,10 @@
         "package.mustache"
       ],
       "matchStrings": [
-        "\\s\"(?<depName>.*?)\": \"(?<currentValue>\\d.*?)\",?\\s"
+        "\"devDependencies\":\\s*{[^}]*}",
+        "\\s\"(?<depName>.*?)\": \"(?<currentValue>.*?)\",?\\s"
```
shortcuts, line 50: aren't those regex independent? Or does the array order defines the match nest level or something?

shortcuts (approved):
Ignore my comment if it's not the case

nice :)

renovate.json
millotp: [recursive match string](https://docs.renovatebot.com/configuration-options/#recursive) ftw !

shortcuts: that's sick

PR #823, opened 2022-07-13
fix(javascript): ensure requesters work as in v4

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/APIC-588

### Changes included:

The goal of this PR is to import requester related tests implemented in v4 to ensure we did not broke it when refactoring.

Changes in this PR:
- Import requester tests from v4, they are 1:1 the same, except some syntax changes
  - I've kept the server logic in the test files, but it will be moved at the monorepo level later to be able to do similar tests on the CTS
  - Some of those tests should also disappear once we have the e2e CTS in place, but can be done in parallel
- Fix wrong requester.send signature that was made compliant for the `echoRequester`
- Add some type definition + re-use types

## Next

Ensure body is correct in POST and DELETE methods
  - Done in https://github.com/algolia/api-clients-automation/pull/849

## 🧪 Test

CI :D

PR #849, opened 2022-07-18
fix(clients): make POST body related to the endpoint, assert body in tests

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-321

The issue was discovered in https://github.com/algolia/api-clients-automation/pull/823, but I've moved the fix for the clients here.

### Changes included:

The goal of this PR is to assert the body of each requests in the CTS even when we don't expect to have one.

On some `POST` requests, the body can be optional (e.g. `browse`), but needs to be a valid JSON object, when no body is expected at all (e.g. `clearAllSynonyms`), an empty string is considered as valid. ~In any case, an empty object (`'{}'`) is considered as valid.~ Some endpoints **do not expect a body at all**, so we need to move that logic at the method level since it can't be factorized.

### Issue

1. Calling the `browse` method with only an `indexName` throws an empty body error
2. Calling the `clearRules` method with a body throws an unexpected body error

### Next

Fix `null` body for `GET` and `DELETE` methods on the PHP client, once @damcou is back: https://algolia.atlassian.net/browse/APIC-591

## 🧪 Test

- CI
- Playground:
  - `deleteApiKey` for the `DELETE` method change
  - `browse` with empty object only for the `POST` method change

### DIFF
--- a/clients/algoliasearch-client-java-2/algoliasearch-core/src/main/java/com/algolia/utils/JSONBuilder.java
+++ b/clients/algoliasearch-client-java-2/algoliasearch-core/src/main/java/com/algolia/utils/JSONBuilder.java
@@ -23,6 +23,9 @@ public class JSONBuilder {
     mapper.disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
     mapper.enable(SerializationFeature.WRITE_ENUMS_USING_TO_STRING);
     mapper.enable(DeserializationFeature.READ_ENUMS_USING_TO_STRING);
+
+    // Allow defaulting a POST body property to an empty Object instance
+    mapper.configure(SerializationFeature.FAIL_ON_EMPTY_BEANS, false);
     return mapper;
   }
 }
--- a/clients/algoliasearch-client-java-2/algoliasearch-core/src/main/java/com/algolia/utils/JSONBuilder.java
+++ b/clients/algoliasearch-client-java-2/algoliasearch-core/src/main/java/com/algolia/utils/JSONBuilder.java
@@ -23,9 +23,6 @@ public class JSONBuilder {
     mapper.disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
     mapper.enable(SerializationFeature.WRITE_ENUMS_USING_TO_STRING);
     mapper.enable(DeserializationFeature.READ_ENUMS_USING_TO_STRING);
-
-    // Allow defaulting a POST body property to an empty Object instance
-    mapper.configure(SerializationFeature.FAIL_ON_EMPTY_BEANS, false);
     return mapper;
   }
 }
--- a/clients/algoliasearch-client-javascript/packages/client-common/src/transporter/createTransporter.ts
+++ b/clients/algoliasearch-client-javascript/packages/client-common/src/transporter/createTransporter.ts
@@ -85,17 +85,20 @@ export function createTransporter({
 
   async function retryableRequest<TResponse>(
     request: Request,
-    requestOptions: RequestOptions
+    requestOptions: RequestOptions,
+    isRead = true
   ): Promise<TResponse> {
     const stackTrace: StackFrame[] = [];
-    const isRead = request.useReadTransporter || request.method === 'GET';
 
     /**
      * First we prepare the payload that do not depend from hosts.
      */
     const data = serializeData(request, requestOptions);
-    const headers = serializeHeaders(baseHeaders, requestOptions);
-    const method = request.method;
+    const headers = serializeHeaders(
+      baseHeaders,
+      request.headers,
+      requestOptions.headers
+    );
 
     // On `GET`, the data is proxied to query parameters.
     const dataQueryParameters: QueryParameters =
@@ -109,6 +112,7 @@ export function createTransporter({
     const queryParameters: QueryParameters = {
       'x-algolia-agent': algoliaAgent.value,
       ...baseQueryParameters,
+      ...request.queryParameters,
       ...dataQueryParameters,
     };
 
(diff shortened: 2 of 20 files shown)


shortcuts:
> Looks good ! Hopefully all endpoints accept that !

Doing more tests on the `DELETE` side before merging

templates/java/tests/requests/requests.mustache
```diff
@@ -66,12 +66,14 @@ class {{client}}RequestsTests {
...
         assertEquals(req.path, "{{{request.path}}}");
         assertEquals(req.method, "{{{request.method}}}");
-
         {{#request.body}}
         assertDoesNotThrow(() -> {
             JSONAssert.assertEquals("{{#lambda.escapeQuotes}}{{{request.body}}}{{/lambda.escapeQuotes}}", req.body, JSONCompareMode.STRICT);
         });
         {{/request.body}}
+        {{^request.body}}
+        assertEquals(req.body, null);
```
millotp, line 75: you can use `assertNull` :)

millotp (commented):
Looks good for java and php ! But you are still getting undefined for js

shortcuts:
> Looks good for java and php ! But you are still getting undefined for js

Yup I'm importing the echoRequester fixes from the other PR, just making sure it works properly

templates/java/tests/requests/requests.mustache
shortcuts: done!

clients/algoliasearch-client-javascript/packages/client-common/src/transporter/createTransporter.ts
```diff
@@ -236,41 +240,23 @@ export function createTransporter({
...
-        ...baseRequestOptions.headers,
-      },
-    };
-
-    const isRead = request.useReadTransporter || request.method === 'GET';
+    /**
+     * A read request is either a `GET` request, or a request that we make
+     * via the `read` transporter (e.g. `search`).
+     */
+    const isRead = Boolean(
```
millotp, line 250: is `Boolean()` needed ? can it be anything other than a boolean ?

clients/algoliasearch-client-php/lib/RequestOptions/RequestOptionsFactory.php
```diff
@@ -50,7 +50,8 @@ private function normalize($options)
             'headers' => [
                 'x-algolia-application-id' => $this->config->getAppId(),
                 'x-algolia-api-key' => $this->config->getAlgoliaApiKey(),
-                'User-Agent' => $this->config->getAlgoliaAgent() !== null
+                'User-Agent' =>
```
millotp, line 53: I think the php changes can be moved to another PR until @damcou is back

tests/output/javascript/yarn.lock
```diff
@@ -35,12 +35,6 @@ __metadata:
   languageName: node
```
millotp, line 35: weird that this wasn't generated by the CTS, maybe you can include this to the file to push from the CI ?

clients/algoliasearch-client-java-2/algoliasearch-core/src/main/java/com/algolia/utils/JSONBuilder.java
```diff
@@ -23,6 +23,9 @@ public ObjectMapper build() {
     mapper.disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
     mapper.enable(SerializationFeature.WRITE_ENUMS_USING_TO_STRING);
     mapper.enable(DeserializationFeature.READ_ENUMS_USING_TO_STRING);
+
+    // Allow defaulting a POST body property to an empty Object instance
+    mapper.configure(SerializationFeature.FAIL_ON_EMPTY_BEANS, false);
```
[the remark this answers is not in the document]
shortcuts: Yep that's the part I'm not really sure about, it's what they recommend when trying to provide an empty object but I'd rather about it.

It's needed for the `new Object()`, do you have an alternative in mind?

clients/algoliasearch-client-javascript/packages/client-common/src/transporter/createTransporter.ts
shortcuts: Good catch it's a leftover!

clients/algoliasearch-client-java-2/algoliasearch-core/src/main/java/com/algolia/utils/JSONBuilder.java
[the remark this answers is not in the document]
millotp: maybe have a special case for this in the `serialize` method, like `if(obj.getClass().getName().equals("Object")) return "{}";`

clients/algoliasearch-client-php/lib/RequestOptions/RequestOptionsFactory.php
shortcuts: The changes for the PHP client here still fixes the `POST` methods that were not working previously (e.g. `browse`). However, I've removed the transporter/CTS changes in the meantime

clients/algoliasearch-client-java-2/algoliasearch-core/src/main/java/com/algolia/utils/JSONBuilder.java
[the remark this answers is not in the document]
shortcuts: seems to work here: https://github.com/algolia/api-clients-automation/pull/849/commits/8ee4c28f1ba792f31a171524a5d2c0f17777687d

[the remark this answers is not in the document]
millotp: very hacky but as least it's safer than the flag !

millotp (approved):
looks good !

Issue #853, opened 2022-07-20
javascript: certain types aren't generic


Haroenv:
## Description

- SearchResponse, SearchResponses should take a single generic THit
- HighlightResult and SnippetResults should take a single generic THit

SnippetResults and HighlightResult should also be recursive as in v4

```ts
declare type SnippetMatch = {
    readonly value: string;
    readonly matchLevel: 'none' | 'partial' | 'full';
};

export declare type SnippetResult<THit> = THit extends string | number ? SnippetMatch : {
    [KAttribute in keyof THit]: SnippetResult<THit[KAttribute]>;
};


declare type HighlightMatch = {
    readonly value: string;
    readonly matchLevel: 'none' | 'partial' | 'full';
    readonly matchedWords: readonly string[];
    readonly fullyHighlighted?: boolean;
};

export declare type HighlightResult<THit> = THit extends string | number ? HighlightMatch : {
    [KAttribute in keyof THit]?: HighlightResult<THit[KAttribute]>;
};

export declare type Hit<THit> = THit & {
    readonly objectID: string;
    readonly _highlightResult?: HighlightResult<THit>;
    readonly _snippetResult?: SnippetResult<THit>;
    readonly _rankingInfo?: RankingInfo;
    readonly _distinctSeqID?: number;
};
```

shortcuts:
Thanks! This is being worked on for Java and JavaScript here: https://github.com/algolia/api-clients-automation/pull/829

millotp:
Should the hit be overridden by the generic ? like `hits: T[]` and it's up to the user to extend its type with `Hit` ?

shortcuts:
> Should the hit be overridden by the generic ?

It should be `type Hits<T> = Array<Hit<T>>` and `type Hit<T = Record<string, any>> = T & { ...baseProps }`

millotp:
Okay so the generic logic will be exclusively for hits, it would be hard to make to work with other generic type

Haroenv:
Is there a plan for Highlight and Snippet to be correct as well?

shortcuts:
Not for now as we have other priorities, but if you have an implementation in mind feel free to contribute!

Haroenv:
customer found this too: https://github.com/algolia/algoliasearch-client-javascript/issues/1534

PR #854, opened 2022-07-21
feat(javascript): support generic for hits

millotp:
## 🧭 What and Why

🎟 JIRA Ticket: [APIC-544](https://algolia.atlassian.net/browse/APIC-544)

Solves the first point of #853.

Add support for hits in js, it's a bit too custom for hits, hope it works for `hightlightResult` too

### Changes included:

- Reuse most of the code from java generic

## 🧪 Test

Playground


generators/src/main/java/com/algolia/codegen/Utils.java
```diff
@@ -190,4 +190,8 @@ public static void setGenerationBanner(Map<String, Object> additionalProperties)
       "This file is generated, manual changes will be lost - read more on" + " https://github.com/algolia/api-clients-automation."
     );
   }
+
+  public static void prettyPrint(Object o) {
+    Json.prettyPrint(o);
+  }
```
shortcuts, line 196: best addition of this repo

templates/javascript/clients/model.mustache
```diff
@@ -11,18 +11,24 @@ import { {{classname}} } from '{{filename}}';
 /**
 * {{{description}}}
 */{{/description}}
-export type {{classname}} = {{#oneOf}}{{{.}}} {{^-last}}|{{/-last}} {{/oneOf}}{{#allOf}}{{{.}}} {{^-last}}&{{/-last}} {{/allOf}};
+export type {{classname}}{{#vendorExtensions.x-has-child-generic}}<T>{{/vendorExtensions.x-has-child-generic}} = {{#oneOf}}{{{.}}} {{^-last}}|{{/-last}} {{/oneOf}}
```
shortcuts, line 14: How come we don't propagate it in the `oneOf`s?

generators/src/main/java/com/algolia/codegen/GenericPropagator.java
```diff
@@ -138,6 +159,10 @@ public static void propagateGenericsToModels(Map<String, ModelsMap> modelsMap) {
     for (CodegenModel model : models.values()) {
       propagateGenericRecursive(models, model);
     }
+
+    for (CodegenModel model : models.values()) {
+      propagateToComposedSchema(models, model);
+    }
```
shortcuts, line 165: Does the order matter here?

shortcuts (commented):
first pass brb after meal

templates/javascript/clients/model.mustache
millotp: We don't have any generic oneOf yet, i can add it but it won't be used

generators/src/main/java/com/algolia/codegen/GenericPropagator.java
millotp: yes because we need the generic to be set on all models before adding it to the schemas, because we use the models look up table to find them and we need all of them to already have generic propagated.

templates/javascript/clients/model.mustache
shortcuts: Yep might not be necessary for now but if we add it we won't be surprised

templates/javascript/clients/model.mustache
```diff
@@ -11,18 +11,24 @@ import { {{classname}} } from '{{filename}}';
...
 */{{/description}}{{^isEnum}}
-export type {{classname}} = {{#parent}} {{{.}}} & {{/parent}} {
+export type {{classname}}{{#vendorExtensions.x-has-child-generic}}<T>{{/vendorExtensions.x-has-child-generic}} = {{#parent}} {{{.}}} & {{/parent}} {
 {{#vars}}{{#description}}
     /**
     * {{{description}}}
     */{{/description}}
-    {{name}}{{^required}}?{{/required}}: {{#isEnum}}{{classname}}{{{nameInCamelCase}}}{{#isArray}}[]{{/isArray}}{{/isEnum}}{{^isEnum}}{{{dataType}}}{{#isNullable}} | null{{/isNullable}}{{/isEnum}};{{/vars}}
+    {{name}}{{^required}}?{{/required}}: {{#isEnum}}{{classname}}{{{nameInCamelCase}}}{{#isArray}}[]{{/isArray}}{{/isEnum}}
+    {{^isEnum}}{{#vendorExtensions.x-propagated-generic}}(T & {{{complexType}}}){{#isArray}}[]{{/isArray}}{{/vendorExtensions.x-propagated-generic}}
```
shortcuts, line 27: How come TS does not throw with this syntax? We should to `Array<Hit<T>>` instead of `Array<Hit & T>`?

shortcuts: Ah but the `Hit` object is not generic, so it works but not really correct: https://github.com/algolia/api-clients-automation/commit/e149520897e9075fef07c5b667f08411f7405e71

millotp: it's a different behavior than in java indeed, in java we cannot make the `Hit` generic, instead people have to extend it themselves, I'll modify it for js

shortcuts: Yep I just checked the Java output and guessed it make sense

shortcuts (approved):
Looks great, nice addition!

PR #855, opened 2022-07-21
feat(javascript): add `requester-fetch`

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/APIC-572

### Changes included:

Import `fetch` requester from the community implementation: https://github.com/algolia/algoliasearch-client-javascript/pull/1411

Only changes that were added from the first implementation are:
- Use ms for timeouts
- Allow overriding default `requesterOptions`

### Side changes

I've also moved the test utils to their own folder as it was becoming redundant

## 🧪 Test

CI :D

### DIFF
--- a/clients/algoliasearch-client-javascript/packages/requester-fetch/src/createFetchRequester.ts
+++ b/clients/algoliasearch-client-javascript/packages/requester-fetch/src/createFetchRequester.ts
@@ -1,0 +1,81 @@
+import type {
+  EndRequest,
+  Requester,
+  Response as AlgoliaResponse,
+} from '@algolia/client-common';
+
+function isAbortError(error: unknown): boolean {
+  return error instanceof Error && error.name === 'AbortError';
+}
+
+function getErrorMessage(error: unknown, abortContent: string): string {
+  if (isAbortError(error)) {
+    return abortContent;
+  }
+  return error instanceof Error ? error.message : 'Network request failed';
+}
+
+export type FetchRequesterOptions = {
+  readonly requesterOptions?: RequestInit;
+};
+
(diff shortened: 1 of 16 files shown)


clients/algoliasearch-client-javascript/packages/requester-fetch/src/createFetchRequester.ts
```diff
@@ -0,0 +1,81 @@
...
+        redirect: 'manual',
+        signal,
+        ...requesterOptions,
+        headers: {
+          ...requesterOptions.headers,
+          ...request.headers,
+        },
+      });
+    } catch (error) {
+      return {
```
millotp, line 52: doesn't 400 errors also end up here ? In that case we should have a status code

millotp (approved):
Everything looks good ! Just a small question about fetch

clients/algoliasearch-client-javascript/packages/requester-fetch/src/createFetchRequester.ts
shortcuts: > doesn't 400 errors also end up here ?

Those errors are actually forwarded [a bit below](https://github.com/algolia/api-clients-automation/pull/855/files/ecc86bc37bac0a720429975e20b82cb9ef8c2863#diff-e60fa9cc9f808f42c204640f1c5e6870f57c2e173e9238c66d2b24d6ba52c1b3R66-R70), in each requester there is this notion of global error that returns a `0` error, I don't have more context on this part but it seems to correctly forward 400s if you check the tests

PR #868, opened 2022-07-26
fix(specs): make `hits` required, `facets` count as an integer, `searchParams` as optional

shortcuts:
## 🧭 What and Why

### Changes included:

- The `required` field for `hits` was at the wrong spot
  - `hits` is required in a `SearchResponse`, but also a `RecommendationsResponse`
- The `facets` returned are a `Record<string, Record<string, number>>`
- All parameters of `searchParams` are optional, therefore we don't need to make it required.

## 🧪 Test

CI :D

### DIFF
--- a/specs/search/common/schemas/SearchResponse.yml
+++ b/specs/search/common/schemas/SearchResponse.yml
@@ -48,7 +48,7 @@ baseSearchResponse:
       description: If a search encounters an index that is being A/B tested, abTestID reports the ongoing A/B test ID.
     abTestVariantID:
       type: integer
-      description: If a search encounters an index that is being A/B tested, abTestVariantID reports the variant ID of the index used.
+      description: If a search encounters an index that is being A/B tested, abTestVariantID reports the variant ID of the index used (starting at 1).
     aroundLatLng:
       type: string
       description: The computed geo location.
--- a/specs/search/common/schemas/SearchResponse.yml
+++ b/specs/search/common/schemas/SearchResponse.yml
@@ -26,12 +26,13 @@ searchHits:
       type: array
       items:
         $ref: 'Hit.yml#/hit'
+  required:
+    - hits
 
 baseSearchResponse:
   type: object
   additionalProperties: false
   required:
-    - hits
     - nbHits
     - page
     - nbPages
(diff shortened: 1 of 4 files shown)


specs/search/common/schemas/SearchResponse.yml
```diff
@@ -26,12 +26,13 @@ searchHits:
...
       items:
         $ref: 'Hit.yml#/hit'
+  required:
+    - hits
 
 baseSearchResponse:
   type: object
   additionalProperties: false
   required:
-    - hits
```
bodinsamuel, line 34: hits can be undefined in a search response? 🤔 
this api is great

bodinsamuel (dismissed):

specs/search/common/schemas/SearchResponse.yml
shortcuts: No no it can't but `hits` is not defined in the `baseSearchResponse`, it's in the `searchHits`

bodinsamuel: what's the point of baseSearchResponse if it's invalid?

shortcuts: So the end object is `SearchResponse`, which is a `allOf` of:
- `baseSearchResponse` (which is re-used in `recommend` and `browse`)
- `searchHits`

The reason why `searchHits` is excluded from the `baseSearchResponse` is because search hits are different than recommend hits, doing so allow us to have a sharable definition instead of re-writing all props/having many refs

shortcuts: but good remark as I had to re-watch the recommend type that was also wrong :D https://github.com/algolia/api-clients-automation/pull/868/commits/c843d6e71a69c2f8fd1dbc1a25c32f4759213935

bodinsamuel: okay I see, is it possible to have no hits and a nbHits property?

shortcuts: Yep, in that case `hits` would be an empty array with `nbHits` set at 0

bodinsamuel: ah yes, hits is actually required but the shape is changing so it's actually always there 👍🏻

shortcuts: Sorry if my explanation was not clear!

bodinsamuel (approved):

PR #869, opened 2022-07-26
fix(ci): prevent deleting all tests, correctly run JS tasks

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/APIC-595

### Changes included:

- JS ran all jobs instead of only the ones that changed
  - see [build step](https://github.com/algolia/api-clients-automation/runs/7521374907?check_suite_focus=true) running everything instead of `algoliasearch search` clients
- All tests are currently deleted, while we always run the CI over all specs
  - see [generation commit](https://github.com/algolia/api-clients-automation/commit/c5f14fd5502b1ee626d67895ef5573387098414c), only `search` related should run, but we've deleted everything.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -236,7 +236,7 @@ jobs:
 
       - name: Build clients
         if: ${{ steps.cache.outputs.cache-hit != 'true' && matrix.client.language != 'php' }}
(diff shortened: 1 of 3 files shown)


bodinsamuel (commented):
Please wait for review 😬

PR #871, opened 2022-07-26
fix(ci): make CTS only run generated tests

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

Follow up of https://github.com/algolia/api-clients-automation/pull/869

The CTS runs everything present in the client CTS folder, as our CI only run tasks for clients that have changed, we need to:
1. delete every tests present
2. generate tests for clients that have changed
3. store tests of clients that have changed

## 🧪 Test

CI :D

### DIFF
--- a/.github/actions/restore-artifacts/action.yml
+++ b/.github/actions/restore-artifacts/action.yml
@@ -71,8 +71,6 @@ runs:
       shell: bash
       run: |
         rm -rf clients/algoliasearch-client-javascript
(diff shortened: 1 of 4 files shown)


bodinsamuel (approved):
Hard to tell if it's not breaking anything 🥶

shortcuts:
> Hard to tell if it's not breaking anything 🥶

True :/ Those parts are really hard to debug as it runs in only certain scenarios

PR #872, opened 2022-07-26
feat(javascript): allow providing options for node requester

shortcuts:
## 🧭 What and Why

### Changes included:

I discovered during the Crawler implementation that it was already possible in v4 to override those options, but was not introduced here.

## 🧪 Test

CI :D


bodinsamuel (approved):
ah yes very important !

PR #873, opened 2022-07-27
fix(javascript): expose sub `algoliasearch` client's type

shortcuts:
## 🧭 What and Why

### Changes included:

Sub `algoliasearch` client's type are not exposed, which can be useful in complex implementations like in the Crawler. It also help to differentiate which client does what.

## 🧪 Test

CI :D

### DIFF
--- a/templates/javascript/clients/algoliasearch/builds/imports.mustache
+++ b/templates/javascript/clients/algoliasearch/builds/imports.mustache
@@ -16,5 +16,8 @@ export * from './models';
 
 export const apiClientVersion = searchClientVersion;
 
+/**
+ * The client type.
+ */
 export type Algoliasearch = ReturnType<typeof algoliasearch>;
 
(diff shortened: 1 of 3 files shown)


bodinsamuel (commented):
I think we did it before it was published as multiple packages but it's no longer useful imo,
e.g: you can import SearchClient with `@algolia/client-search` instead of `{ SearchClient }  from 'algoliasearch'`
wdyt?

shortcuts:
> I think we did it before it was published as multiple packages but it's no longer useful imo, e.g: you can import SearchClient with `@algolia/client-search` instead of `{ SearchClient } from 'algoliasearch'` wdyt?

Wouldn't typescript complain until you've added `@algolia/client-search` as a dependency if so? IIRC it was something people were complaining about

shortcuts:
There was a bump on the open-issue for v4: https://github.com/algolia/algoliasearch-client-javascript/issues/1080#issuecomment-1154125872

bodinsamuel:
yeah okay for this issue I get it 👌🏻

bodinsamuel (approved):
I should forbid you to force-push 😬

PR #876, opened 2022-07-28
fix(specs): `rules` and `objects` parameters/responses

shortcuts:
## 🧭 What and Why

### Changes included:

Some more specs fixes found during the Crawler implementation.

- make `getObjects` generic as it was in v4
- `distinct` should accept `boolean` as some other parameter of the same shape
- more parameter naming similar to v4
- js generics:
  - use partials for some redundant case, which also helps with readability
  - add default to `Record<string, any>` 

## 🧪 Test

CI :D


bodinsamuel (approved):
At this point of customisation I wonder if we could have chose a different strategy.
E.g: keep the generation simple and wrap with custom code where we would inherit the simple generation and override the methods when needed.

shortcuts:
> At this point of customisation I wonder if we could have chose a different strategy. E.g: keep the generation simple and wrap with custom code where we would inherit the simple generation and override the methods when needed.

Do you mean for the generic part? I admit that it's a bit cumbersome in some parts.

Also, it seems that openapi-generator supports handlebars, which could reduce the logic of those cases

shortcuts:
> we would inherit the simple generation and override the methods when needed.

I think wrapping would have a size impact also :/ which is not ideal for js

PR #887, opened 2022-08-01
feat(javascript): provide `browseObjects`, `browseRules` and `browseSynonyms` helper methods

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/APIC-601

### Motivations:

(See previous motivations in the edit PR section)

We initially thought methods like `browseObjects` were ops related, but I discovered it was pretty used in some codebases, like 3 times in the Crawler.

The goal of this PR is to import the 3 `browse*` methods that were available in the version, to ease the user journey of iterating over API calls.

### Changes included:

- Move `createRetryablePromise` logic to a more generic `createIterablePromise`
  - idk about the name, I'll keep it exposed in the `@algolia/client-common` package as an alternative for users to do whatever they want with it
- Add tests for new options
- Add doc about usage breaking change

## 🧪 Test

- CI :D 
- `yarn docker playground javascript search` or `cd cd playground/javascript/node/ && yarn start:search`


clients/algoliasearch-client-javascript/packages/client-common/src/types/CreateIterablePromise.ts
```diff
@@ -0,0 +1,40 @@
+export type IterableOptions<TResponse> = Partial<{
```
bodinsamuel, line 1: I wonder: why the types files are Uppercased and not the implementation?

templates/javascript/clients/client/api/helpers.mustache
```diff
@@ -1,49 +1,82 @@
 /**
```
bodinsamuel, line 1: Remind me why the helpers are in a mustache?

playground/javascript/node/search.ts
```diff
@@ -1,5 +1,11 @@
-import { searchClient } from '@algolia/client-search';
-import { ApiError } from '@algolia/client-common';
+import {
```
bodinsamuel, line 1: This file change should probably not be commit to keep consistency with the rest of them

bodinsamuel (commented):
Looks good, minor comments 👍🏻 

NB: the "how to test" part of the PR desc is there to help the reviewer, please put actual test command to run and/or repro ☺️

damcou (commented):
Looks good and I'm ok with the way it's done as we already discussed. I'll let someone else having a look on the typescript side of things ;)

playground/javascript/node/search.ts
shortcuts: Yep indeed I've committed it here for you to run the playground without having to update it, but I'll remove it

templates/javascript/clients/client/api/helpers.mustache
shortcuts: Main reason is that those helpers are only present in the `client-search` package, so to avoid having a manually written file of helpers hidden in all the generated ones, I've made it a template

playground/javascript/node/search.ts
```diff
@@ -17,9 +23,15 @@ client.addAlgoliaAgent('Node playground', '0.0.1');
 
 async function testSearch() {
   try {
-    const res = await client.search<{ name: string }>({ requests: [{ indexName: searchIndex, query: searchQuery }] });
+    const records: Record<string, any> = [];
 
-    console.log(`[OK]`, res.results[0].hits![0].name);
+    await client.browseObjects({
+      indexName: 'gatsbyjs',
```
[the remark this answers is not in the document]
shortcuts: I'll remove the playground from the PR it was mostly to ease your review

website/docs/clients/migration-guides/index.mdx
```diff
@@ -415,6 +415,81 @@ client.batch("<YOUR_INDEX_NAME>", new BatchWriteParams().setRequests(batch));
...
+</td><td width="600px">
+
+<TabsLanguage>
+<TabItem value="javascript">
+
+```js
+// browse records
+const records: Record<string, any> = [];
+
+await client.browseObjects({
```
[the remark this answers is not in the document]
shortcuts: The parameter names have been renamed since it matches what we've used for other helpers, but the way it work is the same

clients/algoliasearch-client-javascript/packages/client-common/src/types/CreateIterablePromise.ts
shortcuts: Good point, we use camel case in the generated files but since most of those are imported from the current client, I've kept the same shape. I'll make an other PR with just the renames

templates/javascript/clients/client/api/helpers.mustache
```diff
@@ -1,49 +1,82 @@
...
-  ...createRetryablePromiseOptions
-}: WaitForTaskOptions, requestOptions?: RequestOptions): Promise<GetTaskResponse> {
-  return createRetryablePromise({
-    ...createRetryablePromiseOptions,
+waitForTask(
+  {
+    indexName,
+    taskID,
+    maxRetries = 50,
+    timeout = (retryCount: number): number =>
```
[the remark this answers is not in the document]
shortcuts: It's still an object with the same shape, just that instead of spreading we default the value here

damcou (approved):

PR #888, opened 2022-08-03
feat(php): Restore Guzzle as a suggestion

damcou:
## 🧭 What and Why

🎟 JIRA Ticket: APIC-600

### Changes included:

OpenAPI default generated client introduced Guzzle as a dependency in the composer file, which is different from the current PHP client where this library is just suggested.
The goal of this PR is to mimic the current behaviour where Guzzle is suggested and where the logic fallbacks on a custom HTTP Client in case Guzzle is not installed on the project.

(PR also contains a fix on the Husky exclusion pattern)

## 🧪 Test
- CI / playground

### DIFF
--- a/clients/algoliasearch-client-php/composer.json
+++ b/clients/algoliasearch-client-php/composer.json
@@ -24,7 +24,6 @@
     "ext-curl": "*",
     "ext-json": "*",
     "ext-mbstring": "*",
-    "guzzlehttp/guzzle": "^7.3",
     "guzzlehttp/psr7": "^2.0",
     "psr/http-message": "^1.0",
     "psr/log": "^1.0 || ^2.0 || ^3.0",
@@ -47,5 +46,8 @@
     "psr-4": {
       "Algolia\\AlgoliaSearch\\Test\\": "test/"
     }
+  },
+  "suggest": {
+    "guzzlehttp/guzzle": "If you prefer to use Guzzle HTTP client instead of the Http Client implementation provided by the package"
   }
 }
--- a/clients/algoliasearch-client-php/composer.json
+++ b/clients/algoliasearch-client-php/composer.json
@@ -24,6 +24,7 @@
     "ext-curl": "*",
     "ext-json": "*",
     "ext-mbstring": "*",
+    "guzzlehttp/guzzle": "^7.3",
     "guzzlehttp/psr7": "^2.0",
     "psr/http-message": "^1.0",
     "psr/log": "^1.0 || ^2.0 || ^3.0",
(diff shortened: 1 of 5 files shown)


clients/algoliasearch-client-php/composer.json
```diff
@@ -24,7 +24,6 @@
     "ext-curl": "*",
```
shortcuts, line 24: This shouldn't be committed if it's from a mustache template

clients/algoliasearch-client-php/lib/Algolia.php
```diff
@@ -86,10 +86,29 @@ public static function setLogger(LoggerInterface $logger)
...
+            } else {
+                $guzzleVersion = \GuzzleHttp\ClientInterface::MAJOR_VERSION;
+            }
+        }
+
         if (null === self::$httpClient) {
-            self::setHttpClient(
-                new \Algolia\AlgoliaSearch\Http\GuzzleHttpClient()
-            );
+            if (class_exists('\GuzzleHttp\Client') && 6 <= $guzzleVersion) {
```
shortcuts, line 103: Can't the `suggest` parameter provide a minimum version?

clients/algoliasearch-client-php/lib/Http/CurlHttpClient.php
```diff
@@ -0,0 +1,151 @@
+<?php
```
shortcuts, line 1: It makes me think that our CTS should include tests for all of the requesters we expose, could you please add a task for it?

shortcuts: Also for the JS client, I've imported [tests from the previous major](https://github.com/algolia/api-clients-automation/blob/main/clients/algoliasearch-client-javascript/packages/requester-browser-xhr/src/__tests__/browser-xhr-requester.test.ts) to ensure it still work properly, do we have tests for this one? It could be a great addition in the meantime of the CTS

shortcuts (commented):
Noice

clients/algoliasearch-client-php/lib/Algolia.php
damcou: Apparently not (they didn't do it in the current client so I guess this is supposed to work this way).

https://getcomposer.org/doc/04-schema.md#suggest

"The format is like package links above, except that the values are free text and not version constraints."

shortcuts: could be a great addition to their format :D

clients/algoliasearch-client-php/lib/Http/CurlHttpClient.php
damcou: Created task : APIC-607

shortcuts (commented):
Since it introduces a new HTTP client I'd mark it as a `feat` to be clearer in the changelog

shortcuts (approved):

PR #890, opened 2022-08-03
fix(javascript): make type filename consistent

shortcuts:
## 🧭 What and Why

### Changes included:

TypeScript filenames of the `client-common` package are not consistent with what we exposed in the other clients, in this PR the files are renamed to camelCase.

## 🧪 Test

-


damcou (approved):
Looks good, thanks for this 👍

damcou (approved):

PR #892, opened 2022-08-03
fix(javascript): expose `clearCache` and `appId`

shortcuts:
## 🧭 What and Why

### Changes included:

Those two properties were missing from the root of the client, which introduced unwanted breaking change.

Those properties were already present in v4, I've just imported it here.

## 🧪 Test

CI :D


damcou (approved):
🚀

PR #908, opened 2022-08-08
fix(javascript): utils tests run time

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

The tests can take 0ms to run, so we need to take that into account.

## 🧪 Test

CI :D


damcou (approved):
🕐

PR #909, opened 2022-08-08
fix(php): Test post empty body

damcou:
## 🧭 What and Why

🎟 JIRA Ticket: APIC-591

### Changes included:

- Add tests for POST requests with empty body according to what was done for js and java.

## 🧪 Test

- CI


shortcuts (approved):
idk how I missed that, but glad to see it working!

PR #911, opened 2022-08-09
fix(specs): correct `searchParams` with `cursor` for `browse` method

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/APIC-617

### Changes included:

Not only query strings are accepted for the `browse` methods, but search params too.

## 🧪 Test

- playground with `browse` method implementation


damcou (approved):
👍

PR #915, opened 2022-08-10
fix(php): composer dependency

damcou:
### Changes included:

Added a small change on the dependencies of the composer.json since there's an issue when I try to install the PHP client on a raw laravel app.

## 🧪 Test

CI

### DIFF
--- a/templates/php/composer.mustache
+++ b/templates/php/composer.mustache
@@ -34,7 +34,7 @@
         "psr/simple-cache": "^1.0 || ^2.0 || ^3.0"
     },
     "require-dev": {
-        "friendsofphp/php-cs-fixer": "^3.5.0",
+        "friendsofphp/php-cs-fixer": "^2.0 || ^3.5.0",
         "phpunit/phpunit": "^9.3",
         "vlucas/phpdotenv": "^5.4"
     },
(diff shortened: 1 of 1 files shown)


templates/php/composer.mustache
```diff
@@ -34,7 +34,7 @@
         "psr/simple-cache": "^1.0 || ^2.0 || ^3.0"
     },
     "require-dev": {
-        "friendsofphp/php-cs-fixer": "^3.5.0",
+        "friendsofphp/php-cs-fixer": "^2.0 || ^3.5.0",
```
shortcuts, line 37: Shouldn't we require at least 3.5.0? I believe the syntax can change a lot from a major to an other

damcou: I tested locally, and the v3.x is installed (and no changes are generated in the files), I just want to see if I can install the new php client in a laravel app (currently it's possible with the current php client with this dependency: https://github.com/algolia/algoliasearch-client-php/blob/master/composer.json#L23)

shortcuts: Ah ok if it was like this it make sense to keep it indeed

shortcuts (approved):
nice

PR #919, opened 2022-08-10
fix(javascript): do not send user-agent for Predict

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/APIC-621

See https://algolia.atlassian.net/browse/PRED-523 for context

### Changes included:

The Predict API does not yet support user-agent query parameters, which makes every requests being rejected for unknown/non-parsable query parameter.

This is a not-really-pretty patch until it gets implemented, I've tried to minimize the footprint on the other clients but they all use the same requester. LMK if you have alternatives.

- JavaScript: We now send an empty string to the factory, which is not forwarded by the transporter.
- Tests: Small condition to skip the generation of the user-agent test for the Predict client.

I've added the Jira ticket in a comment near the patches until it gets resolved, to avoid forgetting to remove it.

The fix is only applied for JavaScript, as it's the only client that is recommended from the Predict team.

## 🧪 Test

`yarn docker playground javascript predict` should show an empty `algoliaAgent`, which gets removed if you do a real request (screenshot is with the `echoRequester`)

Screenshot 2022-08-10 at 18 46 48

### DIFF
--- a/clients/algoliasearch-client-javascript/packages/client-common/src/transporter/createTransporter.ts
+++ b/clients/algoliasearch-client-javascript/packages/client-common/src/transporter/createTransporter.ts
@@ -110,11 +110,14 @@ export function createTransporter({
         : {};
 
     const queryParameters: QueryParameters = {
-      'x-algolia-agent': algoliaAgent.value,
       ...baseQueryParameters,
       ...request.queryParameters,
       ...dataQueryParameters,
     };
+
+    if (algoliaAgent.value) {
+      queryParameters['x-algolia-agent'] = algoliaAgent.value;
+    }
 
     if (requestOptions && requestOptions.queryParameters) {
       for (const key of Object.keys(requestOptions.queryParameters)) {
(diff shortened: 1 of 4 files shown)


clients/algoliasearch-client-javascript/packages/client-common/src/transporter/createTransporter.ts
```diff
@@ -110,12 +110,15 @@ export function createTransporter({
...
-      'x-algolia-agent': algoliaAgent.value,
       ...baseQueryParameters,
       ...request.queryParameters,
       ...dataQueryParameters,
     };
 
+    if (algoliaAgent.value) {
+      queryParameters['x-algolia-agent'] = algoliaAgent.value;
+    }
+
```
shortcuts, line 121: This is the only part that impacts the other client, which is something we are in control of as it's only reached by the factory of each clients.

damcou (approved):
LGTM, do we know when the Predict client will accept user-agent ?

shortcuts:
> LGTM, do we know when the Predict client will accept user-agent ?

their engineers are OOO so the issue should be tackled once they come back

PR #922, opened 2022-08-12
fix(javascript): allow undefined object when all parameters are optional

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: 
- https://algolia.atlassian.net/browse/APIC-590
- https://algolia.atlassian.net/browse/APIC-622

### Changes included:

Discovered during Crawler implementation.

Some methods only have optional parameters, but still require an empty object as a parameter, which is not an ideal DX.

## 🧪 Test

CI :D


damcou (approved):
👍

PR #926, opened 2022-08-12
chore: benchmark test

bodinsamuel:
just a test

```sh
cd tests/benchmark && APP_ID= API_KEY= yarn bench
```


bodinsamuel:
For the moment I just tried to put something in place however I encountered a first "issue", the v5 client does not kill itself in a timely manner.
Screenshot 2022-08-12 at 19 15 57

you can compare by running both test manually
```
cd tests/search/js/v5
APP_ID= API_KEY= node ./index.js
```
```
cd tests/search/js/v4
APP_ID= API_KEY= node ./index.js
```

You will see v4 terminating quickly and v5 being idle for 1-2sec. That usually highlight a setTimeout/interval hanging in the event loop or connection not closed somewhere.

bodinsamuel:
Thought I found an issue but just missed an await 👍🏻 

--- 

Here is the result, v5 seems consistently slower but not dramatically slower (oscillate from 6% to 25% percent).
Keeping in mind it's a very dumb test, it might just be a diff of bootstrap more than indexing itself.

Screenshot 2022-08-12 at 19 24 51

shortcuts:
I often get v5 being faster on bigger sample size, I'll try with large batches instead of simple `saveObject` to see if it changes.

(We can however still observe that it varies a lot as most of the operations depends on external factors)

bodinsamuel:
Closing as it's not a prio and hard to interpret as-is

PR #936, opened 2022-08-18
feat(php): add iterator helper methods

damcou:
## 🧭 What and Why

🎟 JIRA Ticket: APIC-599

### Changes included:

⚠️ 
**I know that this is not exactly what was done for the javascript client, but I think it was faster and more efficient to take inspiration of the current PHP client for this task since we have an already working solution (+ the way it's done for js is probably not as friendly for PHP)**

- Imported existing iterators from the current PHP client: https://github.com/algolia/algoliasearch-client-php/tree/master/src/Iterators (instead of making the API calls directly in the iterators, we now pass the `SearchClient` in the constructor so we only use the generated methods in the client)
- Added `browseObjects`, `browseRules` and `browseSynonyms` in the `Helper` class, they call the iterators directly.
- Updated PHP search playground
- Updated migration guide related to these methods.

## 🧪 Test
- `yarn docker cts run php`
- `yarn docker playground php search`
- Netlify preview for the guide.


millotp (approved):
Looks good ! I didn't test it but if it comes from the previous client there should be no issue :)

clients/algoliasearch-client-php/lib/Support/Helpers.php
```diff
@@ -221,4 +224,46 @@ private static function isKeyUpdated($key, $keyParams)
...
+     *
+     * @return RuleIterator
+     */
+    public static function browseRules($indexName, $searchClient, $requestOptions = [])
+    {
+        return new RuleIterator($indexName, $searchClient, $requestOptions);
+    }
+
+    /**
+     * Helper: Iterate on the `searchSynonyms` method of the client to allow aggregating rules of an index.
```
shortcuts, line 257:
```suggestion
     * Helper: Iterate on the `searchSynonyms` method of the client to allow aggregating synonyms of an index.
```

website/docs/clients/migration-guides/index.mdx
```diff
@@ -479,7 +479,44 @@ console.log(synonyms, synonyms.length);
 <TabItem value="php">
 
 ```php
-// WIP
+
+use Algolia\AlgoliaSearch\Support\Helpers;
```
shortcuts, line 483: I'd say it's worth exposing them at the client level instead of an other package, it removes a parameter and make usage more consistent between languages

up to you!

shortcuts (commented):
Nice!

shortcuts (approved):
Looks great!

PR #947, opened 2022-08-22
fix(javascript): use `exports` field to pick correct bundle

francoischalifour:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/PRED-530 https://algolia.atlassian.net/browse/APIC-625
Slack thread: https://algolia.slack.com/archives/C0341QDM3EG/p1660811425419539

This updates the clients exports to use the [`exports`](https://nodejs.org/api/packages.html#packages_main_entry_point_export) field.

There were issues in how clients were exported because the `module` assumed that it would get executed in a Node environment. This threw in multiple scenarios: Vite, Rollup, standard `<script type="module">`, etc.

We now rely on `exports` with two branches: `"node"` to load the Node ESM and `"default"` to load the browser ESM.

> When using environment branches, always include a `"default"` condition where possible. Providing a `"default"` condition ensures that any unknown JS environments are able to use this universal implementation, which helps avoid these JS environments from having to pretend to be existing environments in order to support packages with conditional exports. For this reason, using `"node"` and `"default"` condition branches is usually preferable to using `"node"` and `"browser"` condition branches.
>
> — From [Conditional exports in Node.js](https://nodejs.org/api/packages.html#packages_conditional_exports)

Using `"browser"` instead of `"default"` also fails with Rollup and [`@rollup/plugin-node-resolve`](https://github.com/rollup/plugins/tree/master/packages/node-resolve):

> (!) Plugin node-resolve: Could not resolve import "@algolia/predict" in /packages/predict-tools/dist/esm/PredictToolsSettings.js using exports defined in /node_modules/@algolia/predict/package.json.

For consistency, and to follow the Node recommendation, I also updated the `algoliasearch` browser export to use `"default"`.

### Changes included:

- Rely on `exports` with `"node"` and `"default"` for all API clients
- Rely on `"default"` instead of `"browser"` for `algoliasearch`

## 🧪 Test

No tests added, I believe the CI will do the work?

### DIFF
--- a/templates/javascript/clients/package.mustache
+++ b/templates/javascript/clients/package.mustache
@@ -25,7 +25,8 @@
         "import": "./dist/{{packageName}}.esm.browser.js",
         "default": "./dist/{{packageName}}.umd.js"
       }
-    }
+    },
+    "./src/*": "./src/*"
   },
   "files": [
     "dist",
--- a/templates/javascript/clients/package.mustache
+++ b/templates/javascript/clients/package.mustache
@@ -6,11 +6,27 @@
   {{^isAlgoliasearchClient}}
   "name": "{{{npmNamespace}}}/{{packageName}}",
   "description": "JavaScript client for {{packageName}}",
-  "main": "index.js",
-  "jsdelivr": "dist/{{packageName}}.umd.js",
-  "unpkg": "dist/{{packageName}}.umd.js",
-  "module": "dist/{{packageName}}.esm.node.js",
-  "browser": "dist/{{packageName}}.umd.js",
+  "main": "./index.js",
+  "umd:main": "./dist/{{packageName}}.umd.js",
+  "jsdelivr": "./dist/{{packageName}}.umd.js",
+  "unpkg": "./dist/{{packageName}}.umd.js",
+  "types": "./index.d.ts",
+  "exports": {
+    ".": {
+      "node": {
+        "import": "./dist/{{packageName}}.esm.node.js",
+        "module": "./dist/{{packageName}}.esm.node.js",
+        "require": "./dist/{{packageName}}.cjs.js",
+        "default": "./dist/{{packageName}}.cjs.js"
+      },
+      "default": {
+        "umd": "./dist/{{packageName}}.umd.js",
+        "module": "./dist/{{packageName}}.esm.browser.js",
+        "import": "./dist/{{packageName}}.esm.browser.js",
+        "default": "./dist/{{packageName}}.umd.js"
+      }
+    }
+  },
   "files": [
     "dist",
     "model",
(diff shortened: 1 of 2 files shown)


templates/javascript/clients/package.mustache
```diff
@@ -6,11 +6,27 @@
...
-  "main": "index.js",
-  "jsdelivr": "dist/{{packageName}}.umd.js",
-  "unpkg": "dist/{{packageName}}.umd.js",
-  "module": "dist/{{packageName}}.esm.node.js",
-  "browser": "dist/{{packageName}}.umd.js",
+  "main": "./index.js",
+  "umd:main": "./dist/{{packageName}}.umd.js",
+  "jsdelivr": "./dist/{{packageName}}.umd.js",
+  "unpkg": "./dist/{{packageName}}.umd.js",
+  "types": "./index.d.ts",
```
shortcuts, line 13: shouldn't we only rely on the `exports` field like in the `algoliasearch` clause?

shortcuts: or at least make definition consistent between the two so if one is incorrect we know they all are

shortcuts (commented):
Thanks! Small question :D

templates/javascript/clients/package.mustache
francoischalifour: I kept it top-level because as opposed to `algoliasearch`, there's a single type definition for both environments for other packages. I'm fine either way though.

shortcuts (approved):
Should be good after this fix!

templates/javascript/clients/package.mustache
```diff
@@ -3,14 +3,31 @@
...
+        "default": "./dist/{{packageName}}.cjs.js"
+      },
+      "default": {
+        "umd": "./dist/{{packageName}}.umd.js",
+        "module": "./dist/{{packageName}}.esm.browser.js",
+        "import": "./dist/{{packageName}}.esm.browser.js",
+        "default": "./dist/{{packageName}}.umd.js"
+      }
+    },
+    "./src/*": "./src/*.ts",
```
francoischalifour, line 28: I think this should rather either both one of those:

```json
{
    "./src/*": "./src/*",
    "./src/*.ts": "./src/*.ts",
}
```

Otherwise if there's another extension in the folder, it would incorrectly get imported as TypeScript.

shortcuts: Those clients all follow the same pattern with a single API file in their `src` folder, we could be more precise on the name instead of using globs but it shouldn't change for now at least

francoischalifour: I see, but it's less error prone and reduce the public API if we went for `"./src/*.ts": "./src/*.ts"`. It would get tricky to debug if you encounter this bug. So we better address it now imo.

PR #951, opened 2022-08-24
chore(docker): rename the container to api-client to be less generic

millotp:
## 🧭 What and Why

The conteneur name `dev` was a bit vague and could clash with other containers, renaming it to `api-client` avoids that

## 🧪 Test

`yarn docker:setup`


.lintstagedrc
```diff
@@ -1,5 +1,5 @@
 {
-  "generators/**/*.java": "docker exec -t dev yarn cli format java generators",
+  "generators/**/*.java": "docker exec -t api-client yarn cli format java generators",
```
shortcuts, line 2: I'd say let's make it extra explicit and name it after the repository name

PR #958, opened 2022-08-25
chore: prepare release 2022-08-25

millotp:
## Summary

This PR has been created using the `yarn release` script. Once merged, the clients will try to release their new version if their version has changed.

## Version Changes

- ~javascript: 5.0.0-alpha.11 (no commit)~
- java: 4.0.0-SNAPSHOT -> **`minor` _(e.g. 4.0.0-SNAPSHOT)_**
- ~php: 4.0.0-alpha.15 (no commit)~

### Skipped Commits

<p>It doesn't mean these commits are being excluded from the release. It means they're not taken into account when the release process figured out the next version number, and updated the changelog.</p>

<details>
  <summary>
    <i>Commits without language scope:</i>
  </summary>

  
</details>

<details>
  <summary>
    <i>Commits with unknown language scope:</i>
  </summary>

  - chore(docker): rename the container to api-client to be less generic (#951)
</details>


shortcuts:
> Does it matter to make 2 releases ? Here only java will be released, I want to release before leaving

I'm not sure if commits between prepare and release trigger are not skipped D: never tried it that's why I ask

I'll release later this evening you can leave freely ☮️

PR #972, opened 2022-08-29
chore(deps): use latest version of renovate-automatic-branch action

millotp:
## 🧭 What and Why

Github seems to use a cached version when using tag like `@v1`, and never updates.
We want to use the latest feature (like `pull-request-title`) so we use the `latest` tag, as long as @bodinsamuel doesn't sabotage his action

Wait until next friday to try it out !

### DIFF
--- a/.github/workflows/renovate.yml
+++ b/.github/workflows/renovate.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Renovate Automatic Branch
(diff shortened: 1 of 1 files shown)


shortcuts (approved):

bodinsamuel:
It's not cached it's based on tags, I just always forget to retag the v1.

millotp:
oh okay, do you also have to tag `latest` manually ?

bodinsamuel:
no latest is an automatic in "docker" and every container service

PR #974, opened 2022-08-30
feat(specs): create new methods

bengreenbank:
## 🧭 What and Why

I am adding 2 new methods to the Predict API clients:

- Delete a user
- View all users

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/PRED-545?atlOrigin=eyJpIjoiMjdmZmRhZjBiZTg3NGU3YWI3Y2UzZDQ2Y2Q2YzhmOTYiLCJwIjoiaiJ9

### Changes included:

- Create both specs
- Move 'user profile' to common
- Add pagination parameters
- Move some errors to common

## 🧪 Test

Use new endpoints with credentials to test for a response.

PR #975, opened 2022-08-30
chore: fix JavaScript build

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

See build error in https://github.com/algolia/api-clients-automation/pull/974

Glob pattern match does not work with single match (unless we add a trailing comma). We can use regular scope syntax when there's a single element to prevent this error.

This throw:
```
cd clients/algoliasearch-client-javascript && yarn build:many '{@algolia/predict}'
```

This does not:
```
cd clients/algoliasearch-client-javascript && yarn build:many @algolia/predict
// OR
cd clients/algoliasearch-client-javascript && yarn build:many '{@algolia/predict,}'
```

### DIFF
--- a/scripts/buildClients.ts
+++ b/scripts/buildClients.ts
@@ -24,15 +24,18 @@ async function buildClient(
       break;
     case 'javascript':
       const npmNamespace = getClientsConfigField('javascript', 'npmNamespace');
-      const toRun = gens
-        .map(({ additionalProperties: { packageName } }) =>
+      const packageNames = gens.map(
+        ({ additionalProperties: { packageName } }) =>
           packageName === 'algoliasearch'
             ? packageName
             : `${npmNamespace}/${packageName}`
-        )
-        .join(',');
+      );
+      const toRun =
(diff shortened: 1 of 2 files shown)


scripts/buildClients.ts
```diff
@@ -24,15 +24,18 @@ async function buildClient(
...
-        .map(({ additionalProperties: { packageName } }) =>
+      const packageNames = gens.map(
+        ({ additionalProperties: { packageName } }) =>
           packageName === 'algoliasearch'
             ? packageName
             : `${npmNamespace}/${packageName}`
-        )
-        .join(',');
+      );
+      const toRun =
```
millotp, line 33: interesting bug, maybe it's easier to always use the `{name,}` syntax in both cases, that way you don't need the if (unless `{name1,name2,}` isn't valid syntax)

shortcuts: ah it should work yes

millotp (approved):
nice !

PR #976, opened 2022-08-30
fix(cts): friendly error message when test is missing

millotp:
## 🧭 What and Why

When a new method is added to the spec, we check that there is a corresponding request test, but the error message was backward.
Now it's: 
`operationId 'getMachin' does not exist in the tests suite, please create the file 'tests/CTS/methods/requests/client/getMachin.json'`

## 🧪 Test

Remove a test from the CTS

### DIFF
--- a/generators/src/main/java/com/algolia/codegen/cts/tests/TestsRequest.java
+++ b/generators/src/main/java/com/algolia/codegen/cts/tests/TestsRequest.java
@@ -54,7 +54,18 @@ public class TestsRequest extends TestsGenerator {
     for (Map.Entry<String, CodegenOperation> entry : operations.entrySet()) {
       String operationId = entry.getKey();
       if (!cts.containsKey(operationId)) {
-        throw new CTSException("operationId " + operationId + " does not exist in the spec");
+        throw new CTSException(
+          "operationId '" +
+          operationId +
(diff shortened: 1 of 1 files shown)


generators/src/main/java/com/algolia/codegen/cts/tests/TestsRequest.java
```diff
@@ -54,7 +54,16 @@ public void run(Map<String, CodegenModel> models, Map<String, CodegenOperation>
     for (Map.Entry<String, CodegenOperation> entry : operations.entrySet()) {
       String operationId = entry.getKey();
       if (!cts.containsKey(operationId)) {
-        throw new CTSException("operationId " + operationId + " does not exist in the spec");
+        throw new CTSException(
```
shortcuts, line 57: can we add a newline with a link to the doc? https://api-clients-automation.netlify.app/docs/contributing/testing/common-test-suite

shortcuts (commented):
nice ty!

generators/src/main/java/com/algolia/codegen/cts/tests/TestsRequest.java
millotp: good idea, should be good

shortcuts (approved):
Great!

PR #978, opened 2022-08-31
fix(cts): friendly error message when a paremeter is not recognized

millotp:
## 🧭 What and Why

Friendly error message when a parameter is not recognized in the CTS, proposing what's accepted.
It can be confusing when it comes from a oneOf because we infer it's type before, and then display all the possibilities, it could also be a wrong match.
Also warning message for list oneOf, which are not well supported.

## 🧪 Test

Change the name of a param.


shortcuts (approved):
Nice!

PR #979, opened 2022-09-01
chore: add CODEOWNERS

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

Seems like the CODEOWNERS file was at the wrong spot, it should now require reviews from us automatically? 

## 🧪 Test

-


millotp (approved):
The file seemed correct to me before according to the [doc](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners), but let's try like that !

PR #980, opened 2022-09-01
fix(javascript): update Predict models error spec

francoischalifour:
## 🧭 What and Why

🎟 JIRA Ticket: [PRED-544](https://algolia.atlassian.net/browse/PRED-544)

### Changes included

This adapts the new error format of the Predict backend when a model is requested but that it's not available.

Here's the new response format:

```json
{
  "user": "user_1",
  "predictions": {
     "funnel_stage": {
        "error": "The funnel_stage prediction is not available."
     }

  }
}
```

Changes:

- Added the new optional `error` key in each model
- Made `value` and `lastUpdatedAt` optional now, because they may not exist if there's an error

Maybe that we could use smarter conditional types to infer that `value` and `lastUpdatedAt` do exist when there's no `error`.

## 🧪 Test

None added.


millotp (dismissed):
Looks good,
To keep the `required` it could be done with a `oneOf`, however it would also create intermediary types that are not very friendly, this works fine for me

shortcuts:
Yep I'd personally go with a oneOf that does something like

```yml
funnel_stage:
  oneOf:
    - '#/funnelStage'
    - '#/error'

funnelStage:
  type: object
  description: Prediction for the **funnel_stage** model.
  properties:
    value:
      type: array
      items:
        title: funnel_stage
        type: object
        properties:
          name:
            type: string
          probability:
            type: number
            format: double
            minimum: 0
            maximum: 1
        required:
          - name
          - probability
    lastUpdatedAt:
      type: string
  required:
    - value
    - lastUpdatedAt

error:
  type: object
  description: The error when the requested model is not available
  properties:
    error:
      type: string
```

And similar for other objects to re-use the `error` thing

bengreenbank (dismissed):

shortcuts (commented):
_tbh I did not knew my suggestion would work gg_

It seems that there's duplicate names (https://github.com/algolia/api-clients-automation/commit/ef67c7e2759cb4b1d4ed9b6ab6abb4cda091d4f0#diff-ff4658ef4480ca1020bc6ddcd1e8b6fe043d35d4d28f18d5831f9df2dc53bb66R9), you might need to find an other name

bengreenbank (approved):

shortcuts (approved):
Looks great!

PR #1016, opened 2022-09-14
feat(specs): deleteByParams

### LINKED ISSUE #1015: deleteBy endpoint should only accept a subset of searchParams
## Describe the problem

`deleteBy` endpoint should only accept a subset of `searchParams`, cf: https://www.algolia.com/doc/rest-api/search/#delete-by
> This endpoint doesn’t support all the options of a query, only its filters (numeric, facet, or tag) and geo queries. It also doesn’t accept empty filters or query.

## Describe the solution

Create a subset of the `searchParam` (but it would involve quite a lot of changes I guess?)

clemfromspace:
close #1015


shortcuts (approved):
Looks good! We can add some more tests in https://github.com/algolia/api-clients-automation/blob/main/tests/CTS/methods/requests/search/deleteBy.json to ensure the parameters are correctly sent but it shouldn't change from the current behavior!

shortcuts:
Ah yes you still need to update the test file with `deleteByParams` now anyway

clemfromspace:
@shortcuts Done ✅

shortcuts (approved):
thanks!

PR #1042, opened 2022-09-26
fix(generators): change default connection timeouts

bengreenbank:
## 🧭 What and Why

Removing the default connection timeout from Predict client to see if this resolves an issue we are experiencing on the dashboard.

The issue is where the client times out instantly first time, but then works OK if the page is refreshed.

🎟 JIRA Ticket: n/a

### Changes included:

- Override connection timeout within the Predict client

## 🧪 Test
n/a

### DIFF
--- a/templates/javascript/clients/client/builds/browser.mustache
+++ b/templates/javascript/clients/client/builds/browser.mustache
@@ -14,15 +14,15 @@ export function {{apiName}}(
   return create{{capitalizedApiName}}({
     appId,
     apiKey,{{#hasRegionalHost}}region,{{/hasRegionalHost}}
-    timeouts: isPredictClient ? {
+    timeouts: {{#isPredictClient}}{
       connect: 10000,
       read: DEFAULT_READ_TIMEOUT_BROWSER,
       write: DEFAULT_WRITE_TIMEOUT_BROWSER,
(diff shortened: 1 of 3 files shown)


templates/javascript/clients/client/builds/browser.mustache
```diff
@@ -14,7 +14,11 @@ export function {{apiName}}(
   return create{{capitalizedApiName}}({
     appId,
     apiKey,{{#hasRegionalHost}}region,{{/hasRegionalHost}}
-    timeouts: {
+    timeouts: isPredictClient ? {
```
shortcuts, line 17: 😢 if only it was possible in mustache templates 😢 @millotp would sleep better

francoischalifour:
@bengreenbank If we just want to try this fix on the dashboard, we can provide these as options on the dashboard:

```js
algoliapredict(
  application.applicationId,
  application.adminApiKey ?? '',
  application.logRegion === 'de' ? 'eu' : 'us',
  {
    timeouts: {
      connect: 10000,
      read: 2000,
      write: 30000,
    },
  },
)
```

bengreenbank:
Resolved by: https://github.com/algolia/AlgoliaWeb/pull/15270

PR #1178, opened 2022-12-14
fix(clients): replace sources client with ingestion

millotp:
## 🧭 What and Why

Remove `sources` client and replace it with it's big brother `ingestion`.


millotp:
@shortcuts should I force push the deleted files ?

shortcuts:
> @shortcuts should I force push the deleted files ?

🤔 weird that the artefact restores them, feel free to if needed!

PR #1191, opened 2022-12-23
fix(javascript): pass with no tests

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

Seems like it exit 1 when no tests are found


millotp (approved):
weird we never had that before

PR #1202, opened 2022-12-28
refactor(specs): new predict segment condition syntax

bengreenbank:
## 🧭 What and Why

Changing the Predict segment syntax from a `string` to an `object` that follows the [proposed specification](https://codesandbox.io/s/segment-filter-syntax-forked-j4vhwo?file=/src/types/option-filters-syntax.ts).

🎟 JIRA Ticket: [PRED-921](https://algolia.atlassian.net/browse/PRED-921)
https://algolia.atlassian.net/browse/DI-764

### Changes included:

- Add `version` to segment object
- Add new `conditions` format to segment object

## Dependencies

**Do not merge until the API supports the new format. Even if the PR below is merged the changes may not have been deployed to production.**

https://github.com/algolia/predict-api/pull/63

[PRED-921]: https://algolia.atlassian.net/browse/PRED-921?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bengreenbank:
Do not merge yet. See 'Dependencies' heading in the PR description

Fluf22 (dismissed):
Looks good to me!

specs/predict/responses/Segment.yml
```diff
@@ -86,25 +89,265 @@ segment:
...
+  description: The operator used on the numerical filter value.
+  type: string
+  default: =
+  enum:
+    - =
+    - '!='
+    - '>'
+    - '>='
+    - <
+    - <=
```
shortcuts, line 338: not related to this PR but wouldn't you want your API to have the same convention for operators? I see that in some places you use `AND`/`OR`, here we use mathematical operators but below (in `segmentFilterProbability`) we use an other form

shortcuts (commented):
great for the spec, mostly wondering something for the API consistency

specs/predict/responses/Segment.yml
```diff
@@ -86,25 +89,265 @@ segment:
...
+  type: object
+  required:
+    - value
+  properties:
+    operator:
+      $ref: '#/segmentFilterOperatorNumerical'
+    value:
+      $ref: '#/segmentPropertyFilterValue'
+
+segmentPropertyFilterValue:
```
millotp, line 310: this is exactly the same as `segmentAffinityFilterValue`, could you reuse it or try with:
```
segmentPropertyFilterValue:
  allOf:
    - $ref: 'segmentAffinityFilterValue'
```

shortcuts (approved):
my comment is non blocking I'm just curious

millotp (commented):
Nice !

specs/predict/responses/Segment.yml
millotp: unless you want to modify one of them some day you could keep them split

specs/predict/responses/Segment.yml
bengreenbank: We tried to use `<` and `>` as object keys for `segmentFilterProbability` but the spec didn't allow it. So here we use some aliases.

Is there a symbol operator for `OR`? Of course `AND` can be `+`.

specs/predict/responses/Segment.yml
bengreenbank: https://github.com/algolia/api-clients-automation/pull/1202/commits/3060f006374db97638566df1ff5dd588a34622d1

specs/predict/responses/Segment.yml
francoischalifour: `segmentFilterProbability` is out of date, no? https://github.com/algolia/api-clients-automation/blob/f6de2c0f82d8669d0cc701d0b47e630d19bb4dbd/specs/predict/responses/Segment.yml#L345-L353

bengreenbank: These textual representations (`lt`, `lte` etc) actually resolve as symbols `<`, `<=`. See https://github.com/algolia/api-clients-automation/pull/1202/commits/e1c61bea44a33ed1c5f54ba444682300cac56345

Is this what you are referring to?

shortcuts: I was not speaking for the generation, indeed the workaround make it work, but for the consistency of the API it might be great to consider having the same operator convention for each parameters. For example, every comparison could be made with: lt lte gt gte eq neq and or

shortcuts: (btw I don't think it's a discussion for this PR it can still move on!)

shortcuts (approved):

tests/CTS/methods/requests/predict/updateSegment.json
```diff
@@ -20,14 +20,40 @@
...
       "updateSegmentParams": {
-        "conditions": "predictions.order_value.value > 100 AND predictions.funnel_stage.score < 0.9"
+        "conditions": {
+          "operator": "AND",
+          "operands": [
+            {
+              "name": "predictions.order_value",
+              "filters": [
+                {
+                  "operator": ">",
```
shortcuts, line 30: you'd need to update those tests

generators/src/main/java/com/algolia/codegen/AlgoliaJavaGenerator.java
```diff
@@ -120,10 +120,33 @@ public OperationsMap postProcessOperationsWithModels(OperationsMap objs, List<Mo
 
   @Override
   public String toEnumVarName(String value, String datatype) {
-    if ("String".equals(datatype) && !value.matches("[A-Z0-9_]+")) {
+    // when it's not a string, we don't want to change the name of the variable generated
```
shortcuts, line 123: you can revert this commit if there's no more symbols!

shortcuts (approved):

PR #1203, opened 2022-12-29
fix(specs): authenticationID is not required in source and destination

morganleroi:
## 🧭 What and Why
The field `authenticationID` is not required in `Source`. For example, we can setup a new Source, without authentication setup.

### Changes included:
- Remove field from required


shortcuts (commented):
nice! and same for destination now, can you also add the change to this PR please?

millotp (approved):

PR #1210, opened 2023-01-03
fix(specs): example should be integer

kai687:
## 🧭 What and Why

This PR fixes the type of the example of the `taskID` path parameter. It should be an integer, not a string.

🎟 JIRA Ticket: N/A

### Changes included:

- Change type of example of `taskID` to integer


millotp (approved):
good catch, thanks !

PR #1211, opened 2023-01-03
fix(specs): numberOfPendingTasks is plural

kai687:
## 🧭 What and Why

This PR aims to improve the specs for the `listIndicesResponse` schema.

Testing the `/1/indexes` endpoint, I found that `numberOfPendingTasks` should be plural.

This PR also adds default values for `pendingTask` (`false`) and `numberOfPendingTasks` (`0`) as both values always seem to be part of the response.

🎟 JIRA Ticket: N/A

### Changes included:

- fix name of response parameter `numberOfPendingTasks`
- Add default values for `numberOfPendingTasks` and `pendingTask`
- Make `numberOfPendingTasks` required for `listIndicesResponse`
- Make `primary` and `replicas` nullable

### DIFF
--- a/specs/search/common/schemas/listIndicesResponse.yml
+++ b/specs/search/common/schemas/listIndicesResponse.yml
@@ -48,9 +48,13 @@ fetchedIndex:
       description: A boolean which says whether the index has pending tasks. This value is deprecated and should not be used.
     primary:
       type: string
+      nullable: true
+      default: null
       description: Only present if the index is a replica. Contains the name of the related primary index.
     replicas:
       type: array
+      nullable: true
+      default: null
       items:
         type: string
       description: Only present if the index is a primary index with replicas. Contains the names of all linked replicas.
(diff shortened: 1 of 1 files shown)


shortcuts (dismissed):
Great catch, thanks!

shortcuts:
> If numberOfPendingTasks is always present in the response, should this be marked as a required property?

Yep indeed

> Should primary and replicas be marked as nullable with default values of null if they're optional parameters?

As long as they are not required it should be interpreted correctly by the language, but it might make sense in some cases 🤔

shortcuts (approved):
looks great!

shortcuts:
@kai687 will you try your 2nd question in this PR or can I merge?

kai687:
@shortcuts sorry, got side-tracked.

shortcuts:
> @shortcuts sorry, got side-tracked.

no problem!

kai687:
So, I guess making `replicas` and `primary` nullable with `null` as default values broke the code gen.

shortcuts:
> So, I guess making `replicas` and `primary` nullable with `null` as default values broke the code gen.

oops D: yep indeed it looks like

shortcuts (approved):

shortcuts:
Thanks again!

PR #1212, opened 2023-01-04
fix(specs): make enabled nullable

kai687:
## 🧭 What and Why

The `enabled` property has a default of `null` but wasn't declared `nullable`.
This PR addresses that.

🎟 JIRA Ticket: N/A

### Changes included:

- Make `enabled` nullable


shortcuts (approved):
so good! thanks :D

PR #1228, opened 2023-01-11
fix(specs): move lastRun and nextRun to task.trigger

millotp:
## 🧭 What and Why

Put the `lastRun` and `nextRun` in the `trigger` object on tasks, and update `frequency` to `cron`


damcou (approved):
🪨

PR #1245, opened 2023-01-19
feat(specs): add query param to filter list endpoint results

Fluf22:
## 🧭 What and Why

🎟 JIRA Ticket: [DI-837](https://algolia.atlassian.net/browse/DI-837)

### Changes included:

Add query param filtering to lists endpoint in the ingestion API client:
- sources
- destinations
- authentications

[DI-837]: https://algolia.atlassian.net/browse/DI-837?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/specs/ingestion/common/authenticationParameters.yml
+++ b/specs/ingestion/common/authenticationParameters.yml
@@ -1,0 +1,30 @@
+type:
+  name: type
+  in: query
+  description: Which type the returned authentication should have. Can be a list of string separated with commas.
+  required: false
+  style: form
+  explode: false
+  schema:
+    type: array
+    items:
+      $ref: '#/authenticationTypeWithNone'
+    example: basic,oauth,none
+
+platform:
+  name: platform
+  in: query
+  description: Which platform the returned authentication should have. Can be a list of string separated with commas.
+  required: false
+  style: form
+  explode: false
+  schema:
+    type: array
+    items:
+      $ref: './schemas/authentication.yml#/PlatformType'
+    example: commercetools,bigcommerce
+
+authenticationTypeWithNone:
+  oneOf:
+    - $ref: './schemas/authentication.yml#/AuthenticationType'
+    - $ref: './schemas/authentication.yml#/AuthenticationNone'
--- a/specs/ingestion/common/schemas/authentication.yml
+++ b/specs/ingestion/common/schemas/authentication.yml
@@ -97,6 +97,10 @@ AuthenticationType:
   type: string
   enum: ['googleServiceAccount', 'basic', 'apiKey', 'oauth', 'algolia']
 
+AuthenticationNone:
+  type: string
+  enum: ['none']
+
 PlatformType:
   type: string
   enum: ['bigcommerce', 'commercetools']
--- a/specs/ingestion/common/destinationParameters.yml
+++ b/specs/ingestion/common/destinationParameters.yml
@@ -1,0 +1,24 @@
+type:
+  name: type
+  in: query
+  description: Which type the returned destination should have. Can be a list of string separated with commas.
+  required: false
+  style: form
+  explode: false
+  schema:
+    type: array
+    items:
+      $ref: './schemas/destination.yml#/DestinationType'
+    example: search,predict
+
+authenticationID:
+  name: authenticationID
+  in: query
+  description: Which authenticationID the returned destination should have. Can be a list of string separated with commas.
+  required: false
+  style: form
+  explode: false
+  schema:
+    type: array
+    items:
+      type: string
--- a/specs/ingestion/common/sourceParameters.yml
+++ b/specs/ingestion/common/sourceParameters.yml
@@ -1,0 +1,24 @@
+type:
+  name: type
+  in: query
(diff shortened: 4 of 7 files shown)


specs/ingestion/common/schemas/authentication.yml
```diff
@@ -97,6 +97,10 @@ AuthenticationType:
   type: string
   enum: ['googleServiceAccount', 'basic', 'apiKey', 'oauth', 'algolia']
 
+AuthenticationNone:
+  type: string
+  enum: ['none']
+
```
shortcuts, line 103: we don't support none IIRC? does it mean it should be optional?

specs/ingestion/common/authenticationParameters.yml
```diff
@@ -0,0 +1,30 @@
...
+  style: form
+  explode: false
+  schema:
+    type: array
+    items:
+      $ref: './schemas/authentication.yml#/PlatformType'
+    example: commercetools,bigcommerce
+
+authenticationTypeExtended:
+  anyOf:
```
shortcuts, line 28: I'm not sure about our anyOf support but for this case `oneOf` will work otherwise

Fluf22: Ah yes! I'll try with `oneOf`

specs/ingestion/common/schemas/authentication.yml
Fluf22: It will have to be supported by the query param algo, since we allow not having an authentication, and it's a different meaning than not having the param in the query param list

shortcuts: ah indeed that make sense if we want to consider that an empty list should match all authentication type

specs/ingestion/common/authenticationParameters.yml
```diff
@@ -0,0 +1,30 @@
...
+  required: false
+  style: form
+  explode: false
+  schema:
+    type: array
+    items:
+      $ref: './schemas/authentication.yml#/PlatformType'
+    example: commercetools,bigcommerce
+
+authenticationTypeExtended:
```
shortcuts, line 27: Since this should be the principal type the user uses, I'd rename this to `AuthenticationType` and make `AuthenticationType` named something like `AllAuthenticationType` or something

specs/ingestion/common/schemas/authentication.yml
```diff
@@ -97,6 +97,10 @@ AuthenticationType:
...
 PlatformType:
```
shortcuts, line 104: wdyt of Platform instead? So the exported type have a precise name

shortcuts (commented):
looks great :)

specs/ingestion/common/schemas/authentication.yml
shortcuts: (same motivations as the other comment)

Fluf22: I do agree that a precise name would be clearer.
Would you be okay with me creating a ticket for later? I feel like it would pollute the scope of this PR.

shortcuts: the client is not shipped for now so I don't mind but I believe if we sell it with our APIs at GA it should be done

specs/ingestion/common/authenticationParameters.yml
Fluf22: Hmm... I don't know for this one... In my mind, `AuthenticationType` should be the main type. `none` is an extended value that help to filter like `NOT authenticationType`.
Wdyt?

shortcuts: It's common to have the type name as close as possible as the property name. My naming suggestion concerns the `oneOf` which is the one exported at the method level, here Extended is the `main`

specs/ingestion/common/schemas/authentication.yml
Fluf22: Let me create a new PR real quick right now so we don't forget

shortcuts: :kermittyping:

Fluf22: Available here: https://github.com/algolia/api-clients-automation/pull/1246 ☺️

specs/ingestion/common/authenticationParameters.yml
Fluf22: I agree: this extended version is only used in the query parameters. The real `AuthenticationType` matching the API object is used in the original `Authentication` object 😁

shortcuts: I'd at least rename the QP one to properly state that it includes a `none` value in that case

Fluf22: I agree, I'll change the name to reflect that

shortcuts (approved):
wonderful

makes me think we still name list endpoints with get prefix here, idk if we want to make it match our internal naming 🤔

PR #1250, opened 2023-01-20
feat(specs): add sourceID query param to task list endpoint

Fluf22:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:
- Add `sourceID` query param to help task filtering
- Remove `AuthenticationNone` type
- Add `PlatformNone` type

### DIFF
--- a/specs/ingestion/common/authenticationParameters.yml
+++ b/specs/ingestion/common/authenticationParameters.yml
@@ -21,5 +21,10 @@ platform:
   schema:
     type: array
     items:
-      $ref: './schemas/authentication.yml#/Platform'
-    example: commercetools,bigcommerce
+      $ref: '#/platformWithNone'
+    example: commercetools,none
+
+platformWithNone:
+  oneOf:
+    - $ref: './schemas/authentication.yml#/Platform'
+    - $ref: './schemas/authentication.yml#/PlatformNone'
--- a/specs/ingestion/common/authenticationParameters.yml
+++ b/specs/ingestion/common/authenticationParameters.yml
@@ -8,8 +8,8 @@ type:
   schema:
     type: array
     items:
-      $ref: '#/authenticationTypeWithNone'
-    example: basic,oauth,none
+      $ref: './schemas/authentication.yml#/AuthenticationType'
+    example: basic,oauth
 
 platform:
   name: platform
@@ -23,8 +23,3 @@ platform:
     items:
       $ref: './schemas/authentication.yml#/Platform'
     example: commercetools,bigcommerce
-
-authenticationTypeWithNone:
-  oneOf:
-    - $ref: './schemas/authentication.yml#/AuthenticationType'
-    - $ref: './schemas/authentication.yml#/AuthenticationNone'
--- a/specs/ingestion/common/schemas/authentication.yml
+++ b/specs/ingestion/common/schemas/authentication.yml
@@ -101,6 +101,10 @@ Platform:
   type: string
   enum: ['bigcommerce', 'commercetools']
 
+PlatformNone:
+  type: string
+  enum: ['none']
+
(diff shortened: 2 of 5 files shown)


specs/ingestion/common/authenticationParameters.yml
```diff
@@ -1,5 +1,5 @@
-type:
-  name: type
+authenticationType:
```
millotp, line 1: why use specific name here ? we don't support that in our API

millotp (commented):
nice ! but this doesn't reflect our api 🤔

specs/ingestion/common/authenticationParameters.yml
Fluf22: The PR is in review 😬
And it's to be consistent to what is being done on the front end. because it's easier to handle all types together if their names are different ☺️

Fluf22: This has been reverted and I added a mapping record in the frontend to have matching naming between API and query param ☺️

Fluf22:
@millotp and Netlify fixed itself 🙌🏻

specs/ingestion/common/schemas/authentication.yml
```diff
@@ -97,14 +97,14 @@ AuthenticationType:
   type: string
   enum: ['googleServiceAccount', 'basic', 'apiKey', 'oauth', 'algolia']
 
-AuthenticationNone:
```
millotp, line 100: why do you remove this filter ? is it not available anymore ?

millotp (commented):
looks good ! small question for auth

specs/ingestion/common/schemas/authentication.yml
Fluf22: It doesn't exist 😅 It's a mistake I made a few days ago: if you don't want to link an auth to a source, you have to create a source without an auth. But you don't create an auth with a `none` type, that wouldn't make any sense 😛

millotp: oh right that make sense
However `authenticationID` can still be used with `none` in list sources but it's not explicit in the doc, should we had it wherever it's possible ?

Fluf22: I thought about it, but didn't know how I could make it? 🤔 
Maybe a oneOf that validates either a classic `string`, or an enum with just `none` as valid value? But I'm not sure it would make it clearer for the user?

millotp: just the doc is enough I think, or in the examples

millotp: or we make a strongly typed UUID

Fluf22: Just the doc and examples should be enough as a first step, I think. I'll make the change

specs/ingestion/common/sourceParameters.yml
```diff
@@ -14,11 +14,14 @@ type:
 authenticationID:
   name: authenticationID
   in: query
-  description: Which authenticationID the returned source should have. Can be a list of string separated with commas.
+  description: >-
+    Which authenticationID the returned source should have. Can be a list of string separated with commas.
+    Also supports 'none' as a value to return sources that don't bring any authentication.
```
millotp, line 19:
```suggestion
    Also supports 'none' as a value to return sources that don't have any authentication.
```

millotp (approved):
looks good ! non-blocking comment

specs/ingestion/common/sourceParameters.yml
Fluf22: I wasn't sure how to word it so thanks! 😁

millotp (approved):

PR #1279, opened 2023-02-03
fix(specs): correct types for Tasks trigger

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

This PR fixes some wrong types on the Ingestion specs, and makes the definition closer to the real API.

## 🧪 Test

- there's no test for this client.


Fluf22 (approved):
Seems perfect 👍🏻

PR #1286, opened 2023-02-03
feat(javascript): improve on task trigger type guards

Fluf22:
## 🧭 What and Why

🎟 JIRA Ticket: Related to [DI-873](https://algolia.atlassian.net/browse/DI-873)

### Changes included:
- export type guard functions directly
- handle undefined trigger value

[DI-873]: https://algolia.atlassian.net/browse/DI-873?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/templates/javascript/clients/client/api/guards.mustache
+++ b/templates/javascript/clients/client/api/guards.mustache
@@ -5,11 +5,11 @@
  * @summary Guard method that returns a strongly typed specific OnDemandTrigger for a given Trigger.
  * @param trigger - The given Task Trigger.
  */
-isOnDemandTrigger(
-  trigger: Trigger
+export function isOnDemandTrigger(
+  trigger?: Trigger
 ): trigger is OnDemandTrigger {
-  return trigger.type === 'on_demand';
-},
+  return trigger !== undefined && trigger.type === 'on_demand';
+}
 
 /**
  * Guard: Return strongly typed specific ScheduleTrigger for a given Trigger.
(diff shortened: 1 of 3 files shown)


shortcuts (dismissed):
looks good!

shortcuts (approved):
<3 nice catch!

templates/javascript/clients/client/api/guards.mustache
```diff
@@ -5,33 +5,33 @@
...
  * @param trigger - The given Task Trigger.
  */
-isSubscriptionTrigger(
+export function isSubscriptionTrigger(
   trigger: Trigger
 ): trigger is SubscriptionTrigger {
   return trigger.type === 'subscription';
-},
+}
 {{/isIngestionClient}}
```
shortcuts, line 37: just realized the check here but you don't need it in the main file in that case

shortcuts (approved):

templates/javascript/clients/client/api/guards.mustache
Fluf22: Well, I kept it for safety: in the future, if we want to add type guards for another client, we will surely remove the one in the main file, so this functions will already be protected.
Should I still remove it from one of the file?

shortcuts: > Well, I kept it for safety: in the future, if we want to add type guards for another client, we will surely remove the one in the main file, so this functions will already be protected.

yup make sense it's a non blocking comment anyway there's no impact at all

feel free to remove it or not but it won't change anything in the end

PR #1316, opened 2023-02-15
fix(specs): update Predict Segment operators to textual format

francoischalifour:
## 🧭 What and Why

Since the API client generator tool doesn't support `>` symbol as keys, we decided to use the `GT` syntax both in the API client and in the Predict REST API.

This PR updates symbol operators like `"="` to textual operators like `"EQ"`.

### Changes included:

- Updated the probability filters to use textual operators like `"EQ"`
- Updated all operators to uppercase
- Remove operator converter introduced in [`e1c61be` (#1202)](https://github.com/algolia/api-clients-automation/pull/1202/commits/e1c61bea44a33ed1c5f54ba444682300cac56345)

## 🧪 Test

- Added test with filter probability

### Related

- https://github.com/algolia/predict-api/pull/67

### DIFF
--- a/generators/src/main/java/com/algolia/codegen/AlgoliaJavaGenerator.java
+++ b/generators/src/main/java/com/algolia/codegen/AlgoliaJavaGenerator.java
@@ -125,23 +125,6 @@ public class AlgoliaJavaGenerator extends JavaClientCodegen {
       return super.toEnumVarName(value, datatype);
     }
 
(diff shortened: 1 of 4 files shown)


bengreenbank (dismissed):

bengreenbank (approved):

shortcuts (approved):
Love it!

PR #1332, opened 2023-02-20
chore(java): workflow to test using different jdk versions

aallam:
## 🧭 What and Why

Validate compatibility with different environments: 
- Java LTS versions: 8, 11 and 17
- Java distributions: temurin and zulu

🎟 JIRA Ticket: APIC-643

### Changes included:

Add `javacompat` job to `check` workflow. 
On each Java version/distribution, the job builds Java API client and runs CTS tests.


.github/workflows/check.yml
```diff
@@ -273,12 +273,55 @@ jobs:
...
+
+    strategy:
+      fail-fast: false
+      matrix:
+        jdk: [temurin, zulu]
+        java: [8, 11, 17]
+    name: Java ${{ matrix.java }}/${{ matrix.jdk }} Build
+
+    steps:
+      - name: Checkout
```
shortcuts, line 295: shouldn't it restore the artifacts outputted from client-gen? otherwise it tests the previously generated code (e.g. `main`)

shortcuts (commented):
awesome!

Fluf22:
Nothing more to add

.github/workflows/check.yml
aallam: Yes nice suggestion! I've added that :D

.github/workflows/check.yml
```diff
@@ -273,12 +273,60 @@ jobs:
           name: clients-${{matrix.client.language }}
           path: clients-${{matrix.client.language }}.zip
 
+  javacompat:
```
millotp, line 276: could you use a full name pls ? We try to avoid abbreviation on this repository, as people outside of the team are suppose to work on it. Something like `java_compatibility_check` ? or smaller

.github/workflows/check.yml
```diff
@@ -273,12 +273,60 @@ jobs:
...
+        with:
+          type: minimal
+
+      - name: Download java artifacts
+        uses: ./.github/actions/restore-artifacts
+        with:
+          java: true   
+
+      - name: Build
+        run: cd clients/algoliasearch-client-java-2 && ./gradlew build --stacktrace
```
millotp, line 318: could you add a small if in bash to verify the java version ? In case we touch the `setup` action and override the java version

aallam: There is maybe a better way to do this from [gradle](https://docs.gradle.org/current/userguide/toolchains.html#toolchains) :D I will give it a try!

aallam: I've added the following:
* Explicitly declare the java toolchain. In case installed java version changed, the build will fail.
```groovy
java {
    toolchain {
        languageVersion = JavaLanguageVersion.of(11)
    }
}
```
* Set source and target (binary) compatibility to Java 8
```groovy
compileJava {
    sourceCompatibility = '1.8'
    targetCompatibility = '1.8'
}
```

Fluf22 (approved):

shortcuts (approved):
**GG**

shortcuts:
converted to draft until we come back to this

shortcuts:
closing since we choose a different implem

PR #1335, opened 2023-02-23
chore(java): add api compat check

aallam:
## 🧭 What and Why

Compares the current generated code with the latest published version (snapshots included).

To public API check is done using [japicmp-gradle-plugin](https://github.com/melix/japicmp-gradle-plugin). 
In case on an incompatibility, the build will fail. The report should be available under `api/build/reports`.

🎟 JIRA Ticket: APIC-645

### Changes included:

- Add a `api` package to Java API client
- Add a job to run the `:api:japicmp`


shortcuts (commented):
😮 nice!

btw should be gitignore the bin for the playground/tests folders?

.github/workflows/check.yml
```diff
@@ -232,6 +232,10 @@ jobs:
         if: ${{ steps.cache.outputs.cache-hit != 'true' && matrix.client.language == 'javascript' && contains(matrix.client.toRun,'algoliasearch') }}
         run: cd ${{ matrix.client.path }} && yarn workspace algoliasearch test
 
+      - name: Run Java 'algoliasearch' public API validation
```
shortcuts, line 235: cache will hit since there's no code change in this PR, you can bump the version of https://github.com/algolia/api-clients-automation/blob/main/.github/.cache_version to invalidate all caches

aallam:
> btw should be gitignore the bin for the playground/tests folders?

Yes, it shouldn't be included indeed, gradle should build the folder at runtime

.github/workflows/check.yml
aallam: Good to know! thanks! 🙌

clients/algoliasearch-client-java-2/api/build.gradle
```diff
@@ -0,0 +1,37 @@
+import me.champeau.gradle.japicmp.JapicmpTask
+
+plugins {
```
millotp, line 3: could you add a small comment in this file about what this does pls ? it compares with the latest deployed version right ?

millotp (commented):
Nice ! Would be very useful for other languages too, but I have never heard of this library before, not sure if it's available elsewhere

clients/algoliasearch-client-java-2/api/build.gradle
aallam: Thanks for the suggestion! Indeed, it compares the latest published version public API with the current generated code. If a breaking change has been detected, the gradle task (and the job) will fail. 

PS: For now, this checks also the snapshot versions, but we should remove it after version 4 GA.

aallam:
> Nice ! Would be very useful for other languages too, but I have never heard of this library before, not sure if it's available elsewhere

Yes indeed! especially for strongly typed languages! I know that similar ideas exists for other languages such as [kotlin](https://github.com/Kotlin/binary-compatibility-validator), [go](https://pkg.go.dev/golang.org/x/exp/apidiff) and [rust](https://github.com/rust-lang/rust-semverver). It is definitely worth to look more into this!

millotp (approved):
Nice ! 
Looks like this is pretty fast too run also :)

PR #1345, opened 2023-02-27
chore(scripts): catch up on deprecated github actions

millotp:
## 🧭 What and Why

🎟 JIRA Ticket: [DI-923](https://algolia.atlassian.net/browse/DI-923)

We have a lot of deprecation warning, for node 12 and for the `set-output` thing, but it's not supported in js so we have to do a [workaround](https://github.com/actions/toolkit/issues/1218#issuecomment-1288890856) until this is fixed.

[DI-923]: https://algolia.atlassian.net/browse/DI-923?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -6,9 +6,6 @@ inputs:
   type:
     description: Type of setup, `minimal` will only setup the JavaScript monorepo, `matrix` will set the run variables.
     required: false
-  workflow_name:
-    description: Name of the workflow that is executing this action.
-    required: false
   language:
     description: the language for which to install deps
     required: false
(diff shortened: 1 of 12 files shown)


shortcuts (dismissed):
looks perfect!

millotp:
it's only superficially unfortunately, the matrix is empty but should be full :(

shortcuts:
> it's only superficially unfortunately, the matrix is empty but should be full :(

oh wow indeed, was I looking at the wrong run?

millotp:
the `setOutput` in js only works for the first call and then nothing

shortcuts (approved):
:okidoki: gg

PR #1346, opened 2023-02-28
chore(scripts): update gradle to v8

millotp:
## 🧭 What and Why

gradle 8 is required by some of our dependencies, like #1342.


shortcuts (dismissed):
does it have an impact on the generated client?

millotp:
shouldn't have, it's just tooling and we don't use many custom gradle plugin yet (wait until @aallam changes everything ahah)

aallam (dismissed):
Nice! :D this should also improve build speed and `.kts` support! :D

millotp:
rebasing to get the debian image, and now it works locally and on the CI :)

shortcuts (approved):
_GG_

PR #1365, opened 2023-03-02
chore(scripts): use spinnies for multi spinner

millotp:
## 🧭 What and Why

Finally fix the long lasting bug of `ora` handling 1 spinner only, using [spinnies](https://github.com/jbcarpanelli/spinnies) although a bit outdated, it works very well !

Screenshot 2023-03-02 at 11 08 36

### Changes included:

- Make `verbose` global
- Remove deprecated `skipUtils` option

## 🧪 Test

`yarn docker build specs -s`


scripts/buildClients.ts
```diff
@@ -1,24 +1,21 @@
 /* eslint-disable no-case-declarations */
```
shortcuts, line 1: I'm thinking that we should always go verbose mode now that we are "out" of the hard testing phase, wdyt?

shortcuts (commented):
much cleaner and easier to understand! GG

shortcuts (approved):
nothing blocking actually

scripts/buildClients.ts
millotp: you mean on the CI ?

shortcuts: nope everywhere, I feel like people doesn't know it can be verbose and actually most of the time you want to see what's going on

PR #1366, opened 2023-03-02
chore(scripts): remove double error log

millotp:
## 🧭 What and Why

When an error occurs in `execa` we printed the error twice, there is no need for that and we can print a better error instead.


Fluf22 (approved):
👌🏻

shortcuts (approved):
YES TY

PR #1374, opened 2023-03-03
fix(all): release independently each language

millotp:
## 🧭 What and Why

Make the release script independent for each language to avoid cascade failure.
Fix the java version in the java repo.
Always spread for each language and then let git nb diff decide.


shortcuts (approved):
**G**~OO~ _D_

PR #1382, opened 2023-03-07
fix(scripts): released tag needs to be deleted

millotp:
## 🧭 What and Why

`released` tag needs to be deleted both locally and on origin apparently.


morganleroi (approved):
🍳

PR #1396, opened 2023-03-13
feat(specs): make the `eu` region available for the ingestion client

Fluf22:
## 🧭 What and Why

🎟 JIRA Ticket:

### Changes included:

- make the `eu` region available for the ingestion client


shortcuts (approved):
Indeed!

PR #1417, opened 2023-03-22
fix(scripts): report testName in error

millotp:
## 🧭 What and Why

Add the name of the CTS test in the error message to ease debugging.
The diff is misleading, the only change is the indentation.

### DIFF
--- a/generators/src/main/java/com/algolia/codegen/exceptions/CTSException.java
+++ b/generators/src/main/java/com/algolia/codegen/exceptions/CTSException.java
@@ -3,9 +3,15 @@ package com.algolia.codegen.exceptions;
 public class CTSException extends GeneratorException {
 
   private boolean skipable;
+  private String testName;
 
   public CTSException(String message) {
     super(message);
+  }
+
+  public CTSException(String message, String testName) {
+    super(message);
+    this.testName = testName;
   }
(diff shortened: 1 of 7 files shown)


generators/src/main/java/com/algolia/codegen/exceptions/CTSException.java
```diff
@@ -20,4 +26,17 @@ public CTSException(String message, boolean skipable) {
...
   }
+
+  public void setTestName(String testName) {
+    this.testName = testName;
+  }
+
+  @Override
+  public String getMessage() {
+    if (testName != null) {
+      return "Error in " + testName + ": " + super.getMessage();
```
shortcuts, line 37: wdyt of newlines or decorated before the message? it's pretty verbose sometimes

shortcuts (commented):
looks great!!

generators/src/main/java/com/algolia/codegen/exceptions/CTSException.java
millotp: yep good idea

shortcuts (approved):
greatooo

Fluf22 (approved):
👍🏻

PR #1418, opened 2023-03-22
fix(specs): add customRequest to ingestion and small fix

millotp:
## 🧭 What and Why

- Fix the `tasks.trigger` property
- Rename `on_demand` to `onDemand`
- Add the 4 custom requests

### DIFF
--- a/specs/ingestion/spec.yml
+++ b/specs/ingestion/spec.yml
@@ -35,8 +35,14 @@ tags:
     description: A source describe the platform informations.
   - name: tasks
     x-displayName: Tasks
-    description: A Task describe how we fetch the data (reindex, on-demand, subscription) from a source and how we index it on the Algolia side.
+    description: A Task describe how we fetch the data (onDemand, schedule, subscription) from a source and how we index it on the Algolia side.
 paths:
+  # ######################
+  # ### Custom request ###
+  # ######################
+  /1{path}:
+    $ref: '../common/paths/customRequest.yml'
+
   # authentications API.
   /1/authentications:
     $ref: 'paths/authentications/authentications.yml'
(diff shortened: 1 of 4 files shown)


shortcuts (approved):
great find!

specs/ingestion/spec.yml
```diff
@@ -35,8 +35,14 @@ tags:
...
     x-displayName: Tasks
-    description: A Task describe how we fetch the data (reindex, on-demand, subscription) from a source and how we index it on the Algolia side.
+    description: A Task describe how we fetch the data (onDemand, schedule, subscription) from a source and how we index it on the Algolia side.
 paths:
+  # ######################
+  # ### Custom request ###
+  # ######################
+  /1{path}:
+    $ref: '../common/paths/customRequest.yml'
+
```
mehmetaligok, line 45: I guess this is a thing that is used in all the other specs. 

But what is the use-case of this? I am just trying to understand

mehmetaligok (approved):
Looks fine but I asked a question to understand a use-case.

specs/ingestion/spec.yml
shortcuts: exactly!

those endpoints are used as an escape hatch so that the user can't still use a feature even if it's not yet propagated to the API clients, or we can leverage the client while doing internal testing

mehmetaligok: understood, thank you

millotp:
sorry for the double review I never know how to handle stack PR

shortcuts (approved):

shortcuts:
> sorry for the double review I never know how to handle stack PR

the rebase requires a force push so a new review, but you can merge -Xours when on the branch and it _should_ resolve conflicts

PR #1102, opened 2022-10-23
feat(go): bootstrap go ingestion client

millotp:
## 🧭 What and Why

Bootstrap the GO api client generation, with the ingestion API as a first test.
This is very basic for now, but it works good enough for the ingestion API, it's not ready for the other Algolia API yet.

There is no CI support for now, maybe in the next PR.

The go client is compatible with golang `1.13`, like the current `algoliasearch-client-go` package.

### Changes included:

- Add GO to the docker image
- Create the `go-ingestion` client
- Create custom GO generator

## 🧪 Test

The CTS is being made in #1115.

You can try the playground by running `yarn docker playground go ingestion`.

PR #1423, opened 2023-03-24
feat(go): initial setup for go client

mehmetaligok:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/APIC-656

Go API client generation initial setup to prepare the groundwork is implemented. It is heavily inspired by https://github.com/algolia/api-clients-automation/pull/1102. As decided in the RFC Go 1.19 is set as the version to go for now. 

### Changes included:

- CI pipeline updated to not break current flows. 
- Dockerfile updated to support go. 
- Client configs are created according to https://github.com/algolia/algoliasearch-client-go. 
- Go code generator implemented in `generators/src/main/java/com/algolia/codegen/AlgoliaGoGenerator.java` to follow a similar folder/packaging as the original client.
- Basic playground setup for ingestion client is created. 
- Initial Go templates are created. It is mostly copied from https://github.com/algolia/api-clients-automation/pull/1102 and some parts are altered to cover other cases. Most of the creation problems are fixed but still, there are some problems to be handled in the future. Since it is already far from ready. It is not logical to block groundwork development for now. 

### Future Steps:
 - Go mustache templates still need more improvements.
     - Optional body parameters are not properly handled yet. In some cases, code generation is not working as expected.
     - All the [client requirements](https://api-clients-automation.netlify.app/docs/contributing/add-new-language/#algolia-requirements) should be implemented in the generated code. 
 - CTS development should be done. 
 - The release pipeline for the generated code should be set to create automated releases on the API client. 

## 🧪 Test

A small playground example was added. Initial linting with the generated code is fine except for some cases mentioned on `Future Steps`.


config/openapitools.json
```diff
@@ -106,6 +106,48 @@
...
+          "packageName": "personalization"
+        }
+      },
+      "go-recommend": {
+        "output": "#{cwd}/clients/algoliasearch-client-go",
+        "additionalProperties": {
+          "packageName": "recommend"
+        }
+      },
+      "go-search": {
```
shortcuts, line 146: could you add predict to the list? I feel like the team would be happy about it, cc @mahsa-mk @francoischalifour for visibility

scripts/types.ts
```diff
@@ -15,7 +15,7 @@ export type LanguageConfig = {
   utilsPackageVersion?: string;
   tests: {
     extension: string;
-    outputFolder: string;
+    outputFolder?: string;
```
shortcuts, line 18: how come it's optional? we always want the folder even if it's "." or ""

templates/go/partial_header.mustache
```diff
@@ -0,0 +1,18 @@
+/*
```
shortcuts, line 1: for the generation banner we just have this https://github.com/algolia/api-clients-automation/blob/a5ba9841550f2ef828fb719ead537a5d8d038c29/generators/src/main/java/com/algolia/codegen/Utils.java#L187, which is spread to every clients by default (you can print it by doing `{{{generationBanner}}}`)

if you want to improve it I don't mind, it would be great that every clients have a similar one

shortcuts (commented):
LOVELY

basically nothing to add, generation looks great, CI works with the base stuff so I'm good to go with this for a first iteration

config/openapitools.json
mehmetaligok: sure, let me try to add it

shortcuts: doesn't have to be in this PR if it doesn't work, but I'd expect it to be generated, they have some polymorphic endpoint that can help with generating a new client :D

Dockerfile
```diff
@@ -24,6 +24,12 @@ RUN apt-get update && apt-get install -y \
     python3 \
     && rm -rf /var/lib/apt/lists/*
 
+# Go
+COPY --from=golang:1.19-bullseye /usr/local/go/ /usr/local/go/
```
millotp, line 28: why is this line doubled ? Is it possible to use [gvm](https://github.com/moovweb/gvm) or something else to install the specific version ?

config/openapitools.json
mehmetaligok: Okay, I added clients according to the existing go repo's clients. I can try to add if something brakes I will take a note to add it in another PR.

templates/go/client.mustache
```diff
@@ -0,0 +1,332 @@
...
+		return fmt.Errorf("Expected %s to be of type %s but received %s.", name, expected, reflect.TypeOf(obj).String())
+	}
+	return nil
+}
+
+// parameterToString convert interface{} parameters to string, using a delimiter if format is provided.
+func parameterToString(obj interface{}, collectionFormat string) string {
+	var delimiter string
+
+	switch collectionFormat {
```
millotp, line 119: no sure we will ever use something custom like that, I think it can be hardcoded and remove the param `collectionFormat`

templates/go/api.mustache
```diff
@@ -0,0 +1,232 @@
...
+
+func WithItemsPerPage(itemsPerPage int32) Option {
+	return Option{
+		Type: "query",
+		Name: "itemsPerPage",
+		Value: parameterToString(itemsPerPage, ""),
+	}
+}
+
+func WithBody(body interface{}) Option {
```
millotp, line 36: because you use `go > 1.18` you can replace all `interface{}` by `any`

templates/go/api.mustache
```diff
@@ -0,0 +1,232 @@
...
+		{{#is5xx}}
+		if res.StatusCode >= 500
+		{{/is5xx}}
+		{{/range}}
+		{{^range}}
+		{{^wildcard}}
+		if res.StatusCode == {{{code}}} {
+		{{/wildcard}}
+		{{/range}}
+			var v {{{dataType}}}
```
millotp, line 203: apparently we use the same error type for all error, so all those blocks generate the same code, they could be collapsed into a single one, or figure out how to get the correct error type (but we don't do it in other langages as well)

templates/go/client.mustache
```diff
@@ -0,0 +1,332 @@
...
+var jsonCheck = regexp.MustCompile(`(?i:(?:application|text)/(?:vnd\.[^;]+\+)?json)`)
+
+// APIClient manages communication with the {{appName}} API v{{version}}
+// In most cases there should be only one, shared, APIClient.
+type APIClient struct {
+  appID string
+	cfg   *Configuration
+}
+
+// NewAPIClient creates a new API client. Requires a userAgent string describing your application.
```
millotp, line 32: this comment is no longer correct

templates/go/partial_header.mustache
mehmetaligok: I didn't notice that but I couldn't use this as it is. Because in go convention generated file header mostly follows this;  `^// Code generated .* DO NOT EDIT\.$`(https://pkg.go.dev/cmd/go#hdr-Generate_Go_files_by_processing_source). 

So I have made a minor addition in utils, please let me know if it is not appropriate, https://github.com/algolia/api-clients-automation/pull/1423/commits/7393a28198e6060c8605ab91fd2a4a77609a01d0. Also, I keep `partial_header.mustache`. It seemed useful to have that in case we need to add something more in the future.

shortcuts: > I didn't notice that but I couldn't use this as it is. Because in go convention generated file header mostly follows this;

That make sense, I did non looked for convention when adding the header, so IMO you can move the util to the global one, it impacts all client but it's not a huge deal as it seems better.

> Also, I keep partial_header.mustache. It seemed useful to have that in case we need to add something more in the future.

I'd just remove the info related stuff, or redirect to the docs just, so that we don't have to handle this here. For the client name, API versions etc. it's fine to keep them if you want

templates/go/client.mustache
mehmetaligok: yes, it looks like `collectionFormat` is not sent in any place. So I just removed it for simplicity as you suggested.

templates/go/api.mustache
```diff
@@ -0,0 +1,232 @@
...
+  "strings"
+)
+
+type Option struct {
+	Type string
+	Name string
+	Value string
+}
+
+func WithPage(page int32) Option {
```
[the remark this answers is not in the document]
mehmetaligok: Yes, the optional parameter thing is a real problem. I tried to figure it out but couldn't find an easy solution. Probably request sending part should be refactored heavily to properly manage this. 

I already noted this in the description as a future step and I will open up a ticket for this one.

templates/go/api.mustache
mehmetaligok: I replaced our usage. 

But there are still some `interface{}` usage in `openapi-generator` and it is not easy to override because they used some places as hardcoded. I believe it is not a huge deal as of now.

templates/go/api.mustache
mehmetaligok: Unfortunately, I am not sure about this part. But let me add a note about improving error handling in the api template. Since this file already needs some changes I think this shouldn't be a blocker and we can discuss/optimize this in more detail.

templates/go/client.mustache
```diff
@@ -0,0 +1,332 @@
...
+  }
+
+  return &APIClient{
+    appID: cfg.AppID,
+    cfg: &cfg,
+  }
+}
+
+func getDefaultHost({{#hasRegionalHost}}region Region{{/hasRegionalHost}}) string {
+  {{#hasRegionalHost}}return strings.Replace("https://{{{host}}}", "{region}", string(region), -1){{/hasRegionalHost}}
```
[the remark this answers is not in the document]
mehmetaligok: yes, that is not correct. But there are more improvements here as well. Not all client requirements are met yet. So I noted this in the description as well. And I will handle it with a different ticket/PR.

--

btw I said a similar thing for some other comments but my initial purpose for this PR was to set up the groundwork for the go client and clean obvious bugs without affecting other clients' pipelines. Because I don't want to include too many things in one place it is hard to develop and review. With this PR, at least we can merge a part of the go client and iterating over a merged piece would be easier.

templates/go/partial_header.mustache
mehmetaligok: sure, I don't mind removing the info related things.

---

> That make sense, I did non looked for convention when adding the header, so IMO you can move the util to the global one, it impacts all client but it's not a huge deal as it seems better.

I am sorry, I couldn't fully understand. Do you want me to update all the client's headers as the go client? (`// Code generated by OpenAPI Generator (https://openapi-generator.tech), manual changes will be lost - read more on https://github.com/algolia/api-clients-automation. DO NOT EDIT.`) I am just trying to confirm without changing it :D

shortcuts: yup so they all rely on https://github.com/algolia/api-clients-automation/blob/a5ba9841550f2ef828fb719ead537a5d8d038c29/generators/src/main/java/com/algolia/codegen/Utils.java#L187

mehmetaligok: alright, it is updated with all the other changes. And it is ready to be reviewed.

Dockerfile
shortcuts: we should use the same logic as the other clients like, leverage the arg (https://github.com/algolia/api-clients-automation/blob/71d805f2b9a31a3ef149389dae992bc501b48a4d/Dockerfile#L3) and a `-version` file https://github.com/algolia/api-clients-automation/tree/main/config

shortcuts (commented):
looks really nice :)

Dockerfile
mehmetaligok: hmm, I missed this detail. Let me try to achieve something similar. Looks like it is also used in CI as well.

templates/go/partial_header.mustache
```diff
@@ -0,0 +1 @@
+// {{{generationBanner}}}
```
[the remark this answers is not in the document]
mehmetaligok: sure, if needed we can add back later. Updated in https://github.com/algolia/api-clients-automation/pull/1423/commits/b57d3b2d00386466870f1ed26ff87ea0572f51a2

Dockerfile
mehmetaligok: should be done with https://github.com/algolia/api-clients-automation/pull/1423/commits/125e81a978cf456d3d94608cffd5021fd9488486

shortcuts (approved):
awesome, great work! nothing to add on my side, this PR is a great starting point for the later improvements :)

PR #1444, opened 2023-04-04
feat(go): template improvements

mehmetaligok:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/APIC-660

With https://github.com/algolia/api-clients-automation/pull/1423, the initial setup merged but as explained in that PR we need to make more improvements to finalize the generation. 

This PR is meant the close the requirement gap on the go client. There is still some improvement room for the implementation but the needed [requirements](https://api-clients-automation.netlify.app/docs/contributing/add-new-language/#algolia-requirements) should be fulfilled with this PR. 

### Changes included:

- `templates/go/api.mustache` updated to handle parameters properly. 
    - Optional parameters, precisely the optional body parameters, created problems with the generated code.
    - It was converted to a struct approach to creating the correct structure for all request param. It is similar to the [default go template](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/go/api.mustache) but it is altered according to our needs.
- `templates/go/utils.mustache` added to prevent code generation errors.
    -  The nullable primitive types are not generated automatically, with this file, all needed nullable types are added. As of now, we only need `NullableBool` but I didn't remove the rest since it may be needed in the future. 
- transport layer is added as the [original client](https://github.com/algolia/algoliasearch-client-go) and the templates are updated to use the transport layer.
    - It is mostly moved from [algoliasearch-client-go](https://github.com/algolia/algoliasearch-client-go) and removed unnecessary pieces with preserving the functionality. 
    - There are some improvement points on the moved codes but it is better to do these improvements after we set the tests and when we are sure that the client is working as expected. There are some complex operations so moving and updating at the same time could bring some unnoticed bugs. This can prolong the development time which is needed to make this client ready. We can improve this internal layer without breaking changes later on. 

## 🧪 Test

The playground example was briefly updated to ensure the new request structure is working. Also, I couldn't find any static check errors on the code after the changes.


generators/src/main/java/com/algolia/codegen/AlgoliaGoGenerator.java
```diff
@@ -34,14 +38,21 @@ public void processOpts() {
...
     supportingFiles.add(new SupportingFile("client.mustache", "", "client.go"));
     supportingFiles.add(new SupportingFile("response.mustache", "", "response.go"));
+    supportingFiles.add(new SupportingFile("utils.mustache", "", "utils.go"));
 
     try {
       Utils.generateServer(client, additionalProperties);
-
       additionalProperties.put("packageVersion", Utils.getClientConfigField("go", "packageVersion"));
+      // using hostForGo to avoid overriding host in supporting files
+      additionalProperties.put("hostForGo", (String) additionalProperties.get("host"));
```
shortcuts, line 47: does the go generator requires one more variable than the others? If not, I'd rather change the util so that all templates uses the same naming/logic than having a special case

clients/algoliasearch-client-go/algolia/internal/errs/no_more_host_to_try_err.go
```diff
@@ -0,0 +1,21 @@
...
+		intermediateNetworkErrors: errs,
+	}
+}
+
+func (e *NoMoreHostToTryErr) IntermediateNetworkErrors() []error {
+	return e.intermediateNetworkErrors
+}
+
+func (e *NoMoreHostToTryErr) Error() string {
+	return "all hosts have been contacted unsuccessfully, it can either be a server or a network error or wrong appID/key credentials were used. You can use opt.ExposeIntermediateNetworkErrors(true) to investigate."
```
millotp, line 20: do we support `ExposeIntermediateNetworkErrors` ?

clients/algoliasearch-client-go/algolia/internal/transport/requester.go
```diff
@@ -0,0 +1,58 @@
...
+// the Algolia API client (by wrapping it in a user-defined instance of
+// transport.Requester) but would like to wrap it into a middleware layer or
+// pass it to an HTTP interceptor.
+func DefaultHTTPClient() *http.Client {
+	return &http.Client{
+		Transport: defaultTransport,
+	}
+}
+
+var defaultTransport http.RoundTripper = &http.Transport{
```
millotp, line 31: is it necessary to specify the type ?

clients/algoliasearch-client-go/algolia/internal/transport/transport.go
```diff
@@ -0,0 +1,139 @@
...
+
+	return &Transport{
+		requester:     requester,
+		retryStrategy: newRetryStrategy(hosts, readTimeout, writeTimeout),
+		compression:   compression,
+	}
+}
+
+func (t *Transport) Request(ctx context.Context, req *http.Request, k call.Kind) (*http.Response, error) {
+	exposeIntermediateNetworkErrors := false // todo: expose this option to the user
```
millotp, line 42: ah I didn't so the todo forget my comment above

clients/algoliasearch-client-go/algolia/internal/transport/transport.go
```diff
@@ -0,0 +1,139 @@
...
+	var intermediateNetworkErrors []error
+
+	// Add Content-Encoding header, if needed
+	if shouldCompress(t.compression, req.Method, req.Body) {
+		switch t.compression {
+		case compression.GZIP:
```
millotp, line 48: do we plan on supporting other compression algorithm ? if not this could be an if.
also I don't think we do anything about gzip in other languages, maybe it's something we miss

templates/go/api.mustache
```diff
@@ -9,170 +9,198 @@ import (
...
-	Type string
-	Name string
-	Value string
+	optionType  string
+	name  string
+	value string
 }
 
-func WithPage(page int32) Option {
+func QueryParamOption(name string, val any) Option {
```
millotp, line 22: how is the DX with this change ? I know we shouldn't focus on that for the first iterations of the client, just wondering

templates/go/api.mustache
```diff
@@ -9,170 +9,198 @@ import (
...
 
-{{/headerParams}}
 {{#bodyParams}}
   // body params
-  if {{paramName}} != nil {
-  localVarPostBody = {{paramName}}
-  } else {
-  localVarPostBody = body
-  }
+  localVarPostBody = r.{{paramName}}
```
millotp, line 184: same here, it could just be `postBody`

millotp (commented):
looks good ! GG on bringing the retry strategy !

generators/src/main/java/com/algolia/codegen/AlgoliaGoGenerator.java
mehmetaligok: for now it only requires this. My initial thought was also changing the util with other templates but I didn't want to do that directly. But if you are okay to do this change on other templates as well, I can do it. It should be fairly simple

clients/algoliasearch-client-go/algolia/call/kind.go
```diff
@@ -0,0 +1,12 @@
+package call
+
+type Kind int
+
+const (
+	Read Kind = iota
+	Write
```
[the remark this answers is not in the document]
mehmetaligok: these constants are used with the operations, not for the hosts. They are used to call the API endpoints. We don't need another const as `ReadWrite` because there will be no usage, now. We are setting it as `Read` depending on the `x-use-read-transporter` and the rest of the time it is just `Write`. So there is no other option for now. 

when defining hosts we are using the functions below and there are three of them as you mentioned. `IsRead`, `IsWrite`, and `IsReadWrite`.

clients/algoliasearch-client-go/algolia/debug/debug.go
```diff
@@ -0,0 +1,57 @@
...
+// debug logging is enabled should not happen anymore.
+func Disable() {
+	debug = false
+}
+
+// Display displays the given parameter on the standard output in a custom way,
+// depending on the given input type. This function is internally used by the
+// Algolia API client to display, for instance, HTTP requests and responses when
+// debug logging is enabled.
+func Display(itf any) {
```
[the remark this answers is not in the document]
mehmetaligok: make sense, it was copied from the current client. I am renaming it.

clients/algoliasearch-client-go/algolia/internal/transport/requester.go
mehmetaligok: I guess it is not necessary. removed it.

generators/src/main/java/com/algolia/codegen/AlgoliaGoGenerator.java
shortcuts: Yep it's basically changing the variable we use, we should force consistency between generators where applicable so that we don't struggle maintaining

clients/algoliasearch-client-go/algolia/internal/transport/transport.go
mehmetaligok: Let me simplify this as you suggested. 

--

tbh I am not sure about this compression stuff. It was supported by the original client and I thought adding it here will not break anything. But I couldn't see anything specific in the other generated clients.

templates/go/api.mustache
mehmetaligok: At the current stage, this shouldn't be necessary at all because we are proving all optional and required parameters to the users. But in case someone sends a very specific header or query param I add this. Also, I added `opts ...Option` in case if it is needed. 

Normally it will not affect the user but if it will be used in this stage it will look like this; 

```
auths, err := ingestionClient.GetAuthentications(
    ingestionClient.NewApiGetAuthenticationsRequest().WithItemsPerPage(2),
    ingestion.QueryParamOption("myQueryParam1", "myQueryParamValue1"),
    ingestion.HeaderParamOption("myHeaderParam1", "myHeaderParamValue2"), 
    )
```

millotp: the issue is that the user shouldn't have to worry if it's a query param or a header, it should be the job of the API client.
or is this just a custom fallback if the param is not in the api client ?

mehmetaligok:
> oh also I noticed that the go client is using the `default` values in the models, we shouldn't use it and use `nil` instead, to leverage the default value of the API directly (because the API client might not be up-to-date)

@millotp Sorry, I couldn't fully understand. So this means we shouldn't utilize the default value on the specs in any of the models and if the value is not set by the user, we should marshal it as empty?

generators/src/main/java/com/algolia/codegen/AlgoliaGoGenerator.java
mehmetaligok: sure, I will update this with the next commit

templates/go/api.mustache
mehmetaligok: exactly this `is this just a custom fallback if the param is not in the api client`

all the query, header, or body params are provided to the users with the given specs. So normally they shouldn't worry about those but this is just a fallback functionality if there are some rare cases.

millotp:
> we shouldn't utilize the default value on the specs in any of the models and if the value is not set by the user, we should marshal it as empty

exactly, this is what we do in other languages, the default is purely informative for the user, but we don't rely on it

mehmetaligok:
> > > we shouldn't utilize the default value on the specs in any of the models and if the value is not set by the user, we should marshal it as empty
> > 
> > 
> > exactly, this is what we do in other languages, the default is purely informative for the user, but we don't rely on it
> 
> @millotp Oh alright, I haven't noticed this at all. I am not sure if the changes will be small or not. If you don't mind, I would like to tackle this with another PR. So we will not make this PR bigger.

You can ignore this comment, it turned out pretty simple to do so I added it in this PR. When it is removed from the initialization function, the rest is automatically handled. If the user is not using specifically `NewI{{param}}WithDefaults`, we will not set the default values.

clients/algoliasearch-client-go/algolia/debug/utils.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	data, err := io.ReadAll(r)
+	_ = r.Close()
+	if err != nil {
+		return nil, ""
+	}
+	return io.NopCloser(bytes.NewReader(data)), string(data)
+}
+
+func decodeGzipContent(in string) (string, error) {
```
shortcuts, line 27: do we expect to use this?

clients/algoliasearch-client-go/algolia/internal/errs/no_more_host_to_try_err.go
shortcuts, line 1: We might see more and more custom error, it might be worth to align on some names in out contributing documentation so that other clients can benefit it

shortcuts: doesn't have to be tackled in this PR, but worth adding a ticket to tackle it

clients/algoliasearch-client-go/algolia/internal/transport/requester.go
```diff
@@ -0,0 +1,58 @@
+package transport
+
+import (
+	"net"
+	"net/http"
+	"time"
+)
+
+const (
```
shortcuts, line 9: just wondering but are they the same as the current client?

shortcuts (dismissed):
this looks super clean, I don't have anything blocking to add actually

shortcuts:
just for context about the default values, they are set at the engine level so we don't need them in the API clients, that's why they are only in the specs

templates/go/api.mustache
```diff
@@ -9,170 +9,198 @@ import (
...
 // Deprecated
 {{/isDeprecated}}
-func (c *APIClient) {{nickname}}({{#requiredParams}}{{paramName}} {{^isArray}}{{^isPrimitiveType}}*{{/isPrimitiveType}}{{/isArray}}{{{dataType}}}{{^-last}},{{/-last}}{{/requiredParams}}{{#hasOptionalParams}}{{#hasRequiredParams}}, {{/hasRequiredParams}}opts ...Option{{/hasOptionalParams}}) ({{#returnType}}{{^isArray}}{{^returnTypeIsPrimitive}}*{{/returnTypeIsPrimitive}}{{/isArray}}{{{.}}}, {{/returnType}}error) {
+func (c *APIClient) {{nickname}}(r {{#structPrefix}}{{&classname}}{{/structPrefix}}{{^structPrefix}}Api{{/structPrefix}}{{operationId}}Request, opts ...Option) ({{#returnType}}{{^isArray}}{{^returnTypeIsPrimitive}}*{{/returnTypeIsPrimitive}}{{/isArray}}{{{.}}}, {{/returnType}}error) {
 	var (
     localVarPostBody     any
-    {{#hasBodyParam}}body                 any{{/hasBodyParam}}
 		{{#returnType}}
-		returnValue  {{^isArray}}{{^returnTypeIsPrimitive}}*{{/returnTypeIsPrimitive}}{{/isArray}}{{{.}}}
+		localVarReturnValue  {{^isArray}}{{^returnTypeIsPrimitive}}*{{/returnTypeIsPrimitive}}{{/isArray}}{{{.}}}
```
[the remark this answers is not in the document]
millotp: I still think they hurt readability, if we know there is no collision with global variable we should remove it IMO

millotp (commented):
looks good ! nothing more to add except the `localVar`

clients/algoliasearch-client-go/algolia/internal/errs/no_more_host_to_try_err.go
mehmetaligok: Yes, I will open up a ticket in general about error handling to make it more standard.

clients/algoliasearch-client-go/algolia/internal/transport/requester.go
mehmetaligok: yeap, they are the same as the current client, ref: https://github.com/algolia/algoliasearch-client-go/blob/master/algolia/transport/requester.go#L9-L14

templates/go/api.mustache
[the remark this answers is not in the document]
mehmetaligok: okay no huge deal for me. 
I believe there is no collision as of now for `returnValue ` and `postBody`. But let me double-check with generating the code to see.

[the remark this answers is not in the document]
mehmetaligok: alright, updated with https://github.com/algolia/api-clients-automation/pull/1444/commits/feef2aa30200c0ef550a3953170801628449c045

shortcuts (approved):
Gg!!

PR #1447, opened 2023-04-06
feat(go): add more playground examples

mehmetaligok:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/APIC-664

More structured playground examples were added for the go client to give more examples to test.

### Changes included:

- Playground examples added for generated go packages. 

## 🧪 Test

can be tested with `yarn docker playground go {client}`

### DIFF
--- a/playground/go/analytics.go
+++ b/playground/go/analytics.go
@@ -1,0 +1,24 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/algolia/algoliasearch-client-go/v4/algolia/analytics"
+)
+
+func testAnalytics(appID, apiKey string) int {
+	indexName := getEnvWithDefault("ANALYTICS_INDEX", "test_index")
+	analyticsClient := analytics.NewClient(appID, apiKey, analytics.US)
+
+	getTopFilterForAttributeResponse, err := analyticsClient.GetTopFilterForAttribute(
+		analyticsClient.NewApiGetTopFilterForAttributeRequest("myAttribute1,myAttribute2").WithIndex(indexName),
+	)
+	if err != nil {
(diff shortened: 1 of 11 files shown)


Fluf22 (approved):
Niiiiice 👌🏻

playground/go/analytics.go
```diff
@@ -0,0 +1,24 @@
...
+
+	"github.com/algolia/algoliasearch-client-go/v4/algolia/analytics"
+)
+
+func testAnalytics(appID, apiKey string) int {
+	indexName := getEnvWithDefault("ANALYTICS_INDEX", "test_index")
+	analyticsClient := analytics.NewClient(appID, apiKey, analytics.US)
+
+	getTopFilterForAttributeResponse, err := analyticsClient.GetTopFilterForAttribute(
+		analyticsClient.NewApiGetTopFilterForAttributeRequest("myAttribute1,myAttribute2").WithIndex(indexName),
```
millotp, line 14: For a future improvement for DX, what about putting only the optional parameters inside an struct ? And keep the required ones at the first level

millotp (approved):
Looks nice ! Thanks for adding them 😀

playground/go/analytics.go
mehmetaligok: tbh I also noticed that the required param can be handled in a better position, and already took a private note about it when creating these examples 😀 
image

--

But directly removing from the struct may bring some other problems, I think more to add in the to `New` function for the request struct.  
 For example; using like `analyticsClient.NewApiGetTopFilterForAttributeRequest("myAttribute1,myAttribute2", indexName)` if the index name is required. So it would be more clear to user what is required when creating the request struct.

PR #1457, opened 2023-04-12
fix(specs): task trigger guard update

Fluf22:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

Update task trigger guard so we can use TaskCreateTrigger with them, since its signature changed


millotp (approved):
good point !

Fluf22:
Oops!

Sorry, I thought I would be able to push without checking first locally 😭 Doing it right now...

PR #1458, opened 2023-04-12
fix(specs): task trigger guard import

Fluf22:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:
Fix forgotten type import for updated type guard


shortcuts (approved):
The linter swaps the types order, you can do it in the template directly if you want so it's faster to process

PR #1459, opened 2023-04-12
fix(specs): task trigger guard import order

Fluf22:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

Update task trigger guard import order so it matches the linter's


shortcuts (changes requested):
all good but waiting for release to be merged as it might break it otherwise

shortcuts (approved):
noiiiiiiice

PR #1475, opened 2023-04-19
fix(javascript): remove `mode` from Fetch requester

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-1210

### Changes included:

Motivations can be found in https://github.com/algolia/algoliasearch-client-javascript/pull/1452

TLDR: We can remove this parameter from the defaulted values, as it's not compatible with Cloudflare workers. If the user wants to specify it, they can by providing request options.


millotp (approved):
Interesting !

PR #1474, opened 2023-04-19
feat(specs): add `neuralSearch` options

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-1210

### Changes included:

closes https://github.com/algolia/api-clients-automation/issues/1456

this PR implements the changes made on the JavaScript client to leverage the new NeuralSearch parameters.

PR #1476, opened 2023-04-19
chore: make Java breaking changes non blocking

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

It's not possible to easily debug issues like https://github.com/algolia/api-clients-automation/pull/1474 as the artifact is not generated when the CI fails, and therefore the file not accessible. This PR changes the CI check to be non blocking but to print the file if it exists

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -239,7 +239,7 @@ jobs:
         if: ${{ steps.cache.outputs.cache-hit != 'true' && matrix.client.language == 'java' }}
         run: |
           cd ${{ matrix.client.path }}
-          ./gradlew :api:japicmp
+          ./gradlew :api:japicmp || exit 0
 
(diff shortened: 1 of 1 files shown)


shortcuts:
I've ported the change to https://github.com/algolia/api-clients-automation/pull/1474 to see if it works

shortcuts:
> I've ported the change to #1474 to see if it works

all good

millotp (commented):
I'm not sure that's the point of the check, it should fail if the public API changed, unless the PR title starts with `feat!` for example

shortcuts:
> I'm not sure that's the point of the check, it should fail if the public API changed, unless the PR title starts with `feat!` for example

yes it make sense but the clients are not yet GA, we can keep it and print it but changes are expected

BTW the changes I've made are additions and shouldn't be considered as breaking changes, which is weird

PR #1492, opened 2023-04-28
fix(specs): update required for kotlin tests to pass

aallam:
## 🧭 What and Why

Fixes necessary API/CTS specifications for Kotlin API client to pass

### Changes included:

- CTS specs
- API specs

### DIFF
--- a/tests/CTS/methods/requests/ingestion/createTask.json
+++ b/tests/CTS/methods/requests/ingestion/createTask.json
@@ -53,7 +53,7 @@
       "sourceID": "search",
       "destinationID": "destinationName",
       "trigger": {
-        "type": "subscription"
+        "type": "onDemand"
       },
       "action": "replace"
     },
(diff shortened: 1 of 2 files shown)


tests/CTS/methods/requests/ingestion/createTask.json
```diff
@@ -53,7 +53,7 @@
       "sourceID": "search",
       "destinationID": "destinationName",
       "trigger": {
-        "type": "subscription"
```
shortcuts, line 56: Both exist, it doesn't work?

aallam: Nope :/ The CTS generator consistently [chooses](https://github.com/algolia/api-clients-automation/blob/7032ca41177d911c8dacba664a9bd79e22661bfc/generators/src/main/java/com/algolia/codegen/cts/tests/ParametersWithDataType.java#L269) the wrong type! Ideally, we should fix the CTS generator to properly handle this case, but that might be time-consuming. So, it's a problem to tackle another day!

shortcuts: Make sense ty for the explanation!

shortcuts (approved):
Gg!!

PR #1535, opened 2023-05-10
feat(go): parameter arguments for request creation improved with all …

mehmetaligok:
…required params

## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/APIC-676

When creating a request instance to pass we are expecting only path params and some required params was needed set with ".With{{Param}}()" syntax.  

### Changes included:

- Instead of just path parameters we are expecting all required params when instantiating.
- For some rare requests, there is no param at all. It is handled to not expect an empty request instance.
- Playground examples are updated with the new generation. 

## 🧪 Test

CI should pass and playground examples can be tried without an issue.


mehmetaligok:
PS: The PR is delayed because of yesterday's GitHub incident since the actions were not running.

Fluf22 (approved):
Looks good to me!

PR #1563, opened 2023-05-23
fix(go): go client tag duplication resolved

mehmetaligok:
## 🧭 What and Why

We have created a new action on go client repo but we are also releasing a tag from this repo as well. So we have duplicate tags on the go client. This needs to be resolved.

### Changes included:

* Action from generated client removed.
* spreadGeneration.ts updated to release go tags with the `v` prefix.

## 🧪 Test
Will be tested with the next release.


millotp (approved):
that's nice ! release without a release action, so powerful

PR #1564, opened 2023-05-23
feat(specs): ingestion partial update method changed to patch

mehmetaligok:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-1280

Ingestion partial updates will use `PATCH` instead of `PUT`.

### Changes included:

- `PUT` method changed to `PATCH` for partial updates.

### DIFF
--- a/specs/ingestion/paths/authentications/authenticationID.yml
+++ b/specs/ingestion/paths/authentications/authenticationID.yml
@@ -16,7 +16,7 @@ get:
     '400':
       $ref: '../../../common/responses/BadRequest.yml'
 
-put:
+patch:
   tags:
     - authentications
   summary: Update a authentication.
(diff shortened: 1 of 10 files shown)


shortcuts (changes requested):
we should have both methods available, right? otherwise it's a breaking change

mehmetaligok:
> we should have both methods available, right? otherwise it's a breaking change

@shortcuts 

For the generated client-wise, it shouldn't be a breaking change for now. Since we are only changing the HTTP method, the name and all the parameters will be the same. For example, usage will stay as `updateAuthentication` and nothing will change from the outer perspective. 

for the documentation part we will only show `PATCH` since we want the user to use `PATCH` for partial updates and currently, we are allowing both `PUT` and `PATCH` and that shouldn't be a breaking change yet.

shortcuts (approved):
oh I thought it was changing the method names, then it should be :okidoki:

PR #1594, opened 2023-06-05
fix(specs): add docker source input

millotp:
## 🧭 What and Why

Add support for `docker` source.

### DIFF
--- a/specs/ingestion/common/schemas/source.yml
+++ b/specs/ingestion/common/schemas/source.yml
@@ -94,7 +94,7 @@ SourceSearch:
 
 SourceType:
   type: string
-  enum: ['bigcommerce', 'commercetools', 'json', 'csv', 'bigquery']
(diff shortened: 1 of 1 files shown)


specs/ingestion/common/schemas/source.yml
```diff
@@ -225,19 +225,59 @@ SourceBigQuery:
...
+      example: v2.1.0
+    imageType:
+      $ref: '#/DockerImageType'
+    outputFile:
+      type: string
+      description: The full name of the output file.
+      example: output.json
+    configuration:
+      type: object
+      description: The configuration of the spec.
```
Fluf22, line 256: Should we say that's a JSON schema?

Fluf22 (approved):

specs/ingestion/common/schemas/source.yml
millotp: is there for the API client ? where is it stored ?

PR #1599, opened 2023-06-07
revert: "fix(javascript): waitForApiKey helper (#1598)"

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

So actually the logic was correct, we compare the payload of the update with the payload of the get API key, if they are similar, then it has changed, otherwise we retry


millotp (approved):
♻️

Fluf22 (approved):
Sorry for misleading everyone yesterday 🤦🏻‍♂️

PR #1611, opened 2023-06-12
feat(dart): implement cts generation

aallam:
## 🧭 What and Why

Adds CTS generation:
- [x] Clients
- [x] Method/Requests

🎟 JIRA Ticket:

APIC-692

### Changes included:

- Add dart generation CTS
- Apply fixes to the issues found using CTS tests.


config/clients.config.json
```diff
@@ -63,13 +63,13 @@
   "dart": {
     "folder": "clients/algoliasearch-client-dart",
     "gitRepoId": "algoliasearch-client-dart",
-    "packageVersion": "0.1.0-alpha.2",
+    "packageVersion": "0.2.0-alpha.1",
```
shortcuts, line 66: is it expected that you bumped it manually?

generators/src/main/java/com/algolia/codegen/AlgoliaDartGenerator.java
```diff
@@ -38,18 +38,16 @@ public void processOpts() {
     String packageFolder;
     if (isAlgoliasearchClient) {
       libName = "algoliasearch";
+      clientName = "Algolia Search Lite";
```
shortcuts, line 41: this seems to have doubled the `Algolia` name in the generated banner

shortcuts: maybe the templates prints both?

shortcuts (commented):
this is HUGE

shortcuts (commented):
can you bump the cache key? it looks like it used the previously built version while the transporter changed :think

config/clients.config.json
aallam: yes, actually need to sync all versions manually for the last time, and drop the suffix since the package is `0.x.x` indicating that the API isn't stable yet, and move to `1.x.x` at GA.

generators/src/main/java/com/algolia/codegen/AlgoliaDartGenerator.java
aallam: True! nice catch!

Fluf22 (approved):
Looks good to me!

shortcuts (approved):
yup nothing to add on my side, GG!!

PR #1613, opened 2023-06-13
fix(specs): add processing run outcome

millotp:
## 🧭 What and Why

Add the `processing` outcome to run, returned when the compute is already taking place by someone else, or a previous request.


shortcuts (approved):
lovely

PR #1622, opened 2023-06-15
fix(specs): add blocking to reasonCode instead of outcome

damcou:
## 🧭 What and Why

🎟 JIRA Ticket: DI-1351

### Changes included:

- blocking value should belong to possible `reasonCode` and not `RunOutcome`


millotp (approved):
oops

PR #1625, opened 2023-06-16
fix(dart): changelog and versions generation

aallam:
## 🧭 What and Why

Dart package need special logic to handle changelog and versions generation.
The idea is to delegate this task to [`melos version`](https://melos.invertase.dev/commands/version).

🎟 JIRA Ticket:

### Changes included:

- Use `melos version` to generate packages and repository changelog and versions.
- Reset packages versions to match the umbrella package version (`algoliasearch`)


scripts/package.json
```diff
@@ -33,7 +33,7 @@
     "fs-extra": "10.1.0",
     "inquirer": "8.2.5",
     "jest": "29.5.0",
-    "js-yaml": "4.1.0",
+    "js-yaml": "^4.1.0",
```
millotp, line 36: we usually pin dependencies to avoid surprises

scripts/release/createReleasePR.ts
```diff
@@ -440,6 +442,7 @@ async function createReleasePR(): Promise<void> {
 
   console.log('Updating config files...');
   await updateAPIVersions(versions, changelog);
+  await updateDartPackages();
```
millotp, line 445: should this be inside a `if (lang === 'dart')` block ?

millotp (commented):
nice ! do you know where the changelog is generated ?

scripts/package.json
aallam: I've tried that but got rejected by the CI, I will try put it back

scripts/release/createReleasePR.ts
aallam: Not need in this case, everything is handled inside `updateDartPackages` :)

aallam:
> nice ! do you know where the changelog is generated ?

Every package will have it's own changelog (as required by pub.dev), and a global changelog is also generated at the repository level with list of changes at each release date :)

scripts/package.json
shortcuts: I guess you need to re-run a yarn install, otherwise the lock is outdated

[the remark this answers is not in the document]
shortcuts: oops I forgot about this one

PR #1644, opened 2023-06-20
chore(ci): fix melos version post scripts

aallam:
## 🧭 What and Why

The scripts is not able to update static version and config file after an update.

### Changes included:

- Update melos post script
- Update config file version script


Fluf22 (approved):
🤞🏻

PR #1645, opened 2023-06-20
chore(ci): fix melos post script compat

aallam:
## 🧭 What and Why

The command `sed` isn't behaving the same way in both macos and linux, causing the scripts to fail.

🎟 JIRA Ticket:

### Changes included:

Update the script to work for BSD/Linux platforms.

### DIFF
--- a/clients/algoliasearch-client-dart/melos.yaml
+++ b/clients/algoliasearch-client-dart/melos.yaml
@@ -42,8 +42,4 @@ scripts:
 command:
   version:
     hooks:
-      post: |
-        for dir in packages/*/ ; do \
-          version=$(grep "^version:" "${dir}pubspec.yaml" | sed 's/version: //') && \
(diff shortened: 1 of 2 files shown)


shortcuts:
alternatively you can do something like

```sh
  # MacOS/GNU sed's implementation differ
  if [ "$(uname)" == "Darwin" ]; then
    sed -i "" $expression $path
  else
    sed -i $expression $path
  fi
```

and store your logic in variables

aallam:
> ```shell
>   # MacOS/GNU sed's implementation differ
>   if [ "$(uname)" == "Darwin" ]; then
>     sed -i "" $expression $path
>   else
>     sed -i $expression $path
>   fi
> ```

@shortcuts I'm giving sed one last chance. If it doesn't work this time, I might just have to 'retire' it!

Fluf22 (approved):
Ah yes, I remember we already had this bug

PR #1649, opened 2023-06-21
fix(dart): update `dio` exceptions handling 

aallam:
## 🧭 What and Why

Starting from version `5.2.0`, the HTTP client Dio has started throwing errors in `dart analyze` due to a migration from `DioError` to `DioException`. Any errors thrown during `dart analyze` will cause the release process to fail.

### Changes included:

- The error handling logic has been updated to use `DioException`.
- The dio version has been updated.
- The Melos scripts have been updated to run `dart analyze` and will now fail if any errors are found.

PR #1650, opened 2023-06-21
chore(dart):  prepare the client for release 

aallam:
## 🧭 What and Why

- The update from #1649 was inadvertently reverted.
- The release process fails if no updates are found.
- The Melos version will check the history if no specific tag is specified.

### Changes Included:

- Reintroduce the changes from #1649.
- Fix the issues with the release process.


Fluf22 (approved):
New try 😎

PR #1651, opened 2023-06-21
feat(specs): updated_at removed from Run response

mehmetaligok:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-1362

We don't want to make updated_at visible for the users since it will be used internally and we don't want to confuse the users. 

### Changes included:

* updated_at removed from Run responses


aallam (approved):

shortcuts (approved):
lovely

shortcuts (approved):

PR #1647, opened 2023-06-20
docs(specs): review Insights API spec

kai687:
## 🧭 What and Why

Review and refine the OpenAPI specification for the Insights API.

🎟 JIRA Ticket: [DOC-1110]

### Changes included:

- [x] Correct parameter limits (maxItems for array types, maxLength for strings)
- [x] Updated API, operation, and parameter descriptions and summaries
- [x] Correct responses
- [x] Example requests and responses, particularly errors
- [x] Separate request bodies for allowed combinations of event attributes

## 🧪 Test

[DOC-1110]: https://algolia.atlassian.net/browse/DOC-1110?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1654, opened 2023-06-22
fix(cts): allow different ordering of `oneOf` schema

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-1379

### Changes included:

This PR makes the CTS code generation work for complex oneOf schemas like in https://github.com/algolia/api-clients-automation/pull/1647.

Today we only guess the matching oneOf by counting the number of matching elements, which is correct but enforce a specific order. The proposed solution is not more correct, but a bit looser as it doesn't enforce the element to be first in the oneOf matching list.

## 🧪 Test

This PR shouldn't impact the current CTS generation.

### DIFF
--- a/generators/src/main/java/com/algolia/codegen/cts/tests/ParametersWithDataType.java
+++ b/generators/src/main/java/com/algolia/codegen/cts/tests/ParametersWithDataType.java
@@ -558,7 +558,9 @@ public class ParametersWithDataType {
             }
           }
         }
-        if (commonCount > maxCount) {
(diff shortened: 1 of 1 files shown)


shortcuts:
shoot, not everywhere

shortcuts:
this can't work with the current way of generating the CTS, will try further with an other solution

shortcuts:
> @shortcuts let me know if this is too much trouble. We can just use a single request body and iterate at a later time.

it's fine we should be able to just remove the test for now, we know that the payload is corretly parsed anyway as we test other payload shape

PR #1664, opened 2023-06-26
refactor(dart): pin algolia packages versions

aallam:
## 🧭 What and Why

This PR pins the versions of the Algolia packages in the `pubspec.yaml` to ensure compatibility and stability as we await API stabilization. This proactive measure minimizes the risk of unexpected behavior or issues stemming from uncontrolled version updates. 

### Changes included:

Pinned versions of Algolia packages. On version update, `melos version` will automatically adjust these pinned versions.


millotp (approved):
good idea !

PR #1669, opened 2023-06-29
fix(dart): `collection` version to `1.17.1`

aallam:
## 🧭 Overview

This PR addresses the dependency issue with the latest version of Flutter, which relies on `collection` version `1.17.1`. Using a higher version prevents users from utilizing the library.

## 📝 Changes

The following modifications have been made:

- The `collection` dependency version has been reduced to `1.17.1` to align with the latest Flutter version.
- Resolved an issue with `pubspec.yaml` generation which involved improper string escaping.

### DIFF
--- a/templates/dart/pubspec.mustache
+++ b/templates/dart/pubspec.mustache
@@ -1,12 +1,12 @@
-name: {{pubName}}
-version: {{pubVersion}}
-description: {{pubDescription}}
-homepage: {{pubHomepage}}
+name: {{{pubName}}}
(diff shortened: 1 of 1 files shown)


millotp (approved):
nice ! does this mean people with up to date dependencies are not able to use the client as well ? or I guess the `^` allows other versions

aallam:
> nice ! does this mean people with up to date dependencies are not able to use the client as well ? or I guess the `^` allows other versions

Yes the caret should allow higher versions indeed :)

PR #1681, opened 2023-07-03
fix(generators): deprecated files gen removal

aallam:
## 🧭 What and Why

A problem was detected with a few of our API clients (Dart, Kotlin, and Go). During the generation process, the gen folders were not properly cleared, resulting in deprecated files not being removed as expected.

### Changes included:

Update API clients' configurations (including api/module paths) and the deletion script to address this issue. The modifications ensure proper cleanup of the gen folders to prevent deprecated files from persisting.


mehmetaligok (approved):
thanks for tackling this for all the clients 🙌

shortcuts (commented):
goooooood catch!!

PR #1682, opened 2023-07-03
docs(specs): Review OpenAPI Recommend

gazconroy:
This PR reviews the Recommend API.

#### getRecommendTaskStatus

- We have the models: trending-items, bought-together, related-products, trending-facets. What about _Looking similar_
 and  _Related content_?
- Adding enums to the recommend model string generates the error: “enum must be out of line automation-custom/out-of-line-enum”

#### getRecommendRule

In the response, it looks like `_metadata` just contains a last update date. Is this correct? What does this date refer to? Date of this `getRecommendRule` operation or something else?

#### saveRecommendRules

This note is taken from the [docs](https://www.algolia.com/doc/rest-api/recommend/#batch-recommend-rules) and raises a few questions:

Recommend rules are similar to [search rules](#tag/Rules/operation/saveRules), except that the conditions and consequences apply to a [_source item_](https://www.algolia.com/doc/guides/algolia-recommend/overview/#recommend-models) instead of a query. The main differences are the following:

 In _conditions_:

    - `pattern` and `anchoring` are unavailable.
    - Filters accepts numeric filters.
    
  In _consequences_:
   
    - params only covers filtering parameters.
    - automaticFacetFilters (in `params`) doesn't require a facet value placeholder. Instead, it tries to match the source item's attributes.

- _`pattern` and `anchoring` are unavailable._ Instead of re-using the Search Rules request object, should it just be copied and pattern and anchoring removed?
- _Filters accepts numeric filters._ Since this is conditions, there aren’t any obvious filters (let alone numeric ones). At least they’re not defined in the OAS for Search rules or Recommend rules.
- _`params` only covers filtering parameters._ It’s not clear from this what is and isn’t covered. Instead of re-using the Search Rules request object, should it just be copied and any inapplicable parameters removed? For example, If applying the request body from the [docs site example](https://www.algolia.com/doc/rest-api/recommend/#method-param-automaticfacetfilter), I get the message:  `{ "message": "promote is forbidden in Recommend rule consequence params", "status":400}`. 
- _`automaticFacetFilters` (in `params`) doesn't require a facet value placeholder…_ Probably better to write exactly what’s expected in a local version of the parameter than apply a caveat.

🎟 JIRA Ticket: [DOC-1108](https://algolia.atlassian.net/browse/DOC-1108)

### Changes included:

Update summary, description and examples

## 🧪 Test

[DOC-1108]: https://algolia.atlassian.net/browse/DOC-1108?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/specs/recommend/common/parameters.yml
+++ b/specs/recommend/common/parameters.yml
@@ -1,15 +1,20 @@
 Models:
   in: path
   name: model
+  required: true
   description: >
     [Recommend models](https://www.algolia.com/doc/guides/algolia-recommend/overview/#recommend-models).
   schema:
-    type: string
-    # enum: [related-products, bought-together, trending-facets, trending-items]
+    $ref: '#/recommendModels'
+
+recommendModels:
+  type: string
+  enum: [related-products, bought-together, trending-facets, trending-items]
 
 ObjectID:
   in: path
   name: objectID
+  required: true
   type: string
   description: Unique Recommend rule identifier.
   example: 'a-recommend-rule-id'
--- a/specs/search/common/schemas/SearchResponse.yml
+++ b/specs/search/common/schemas/SearchResponse.yml
@@ -117,7 +117,7 @@ baseSearchResponse:
     redirect:
       type: object
       description: >
-        [Redirect results to a URL.](https://www.algolia.com/doc/guides/managing-results/rules/merchandising-and-promoting/how-to/redirects/)
+        [Redirect results to a URL](https://www.algolia.com/doc/guides/managing-results/rules/merchandising-and-promoting/how-to/redirects/).
       properties:
         index:
           type: array
--- a/tests/CTS/methods/requests/recommend/deleteRecommendRule.json
+++ b/tests/CTS/methods/requests/recommend/deleteRecommendRule.json
@@ -1,0 +1,13 @@
+[
+  {
+    "parameters": {
+      "indexName": "indexName",
+      "model": "related-products",
+      "objectID": "objectID"
+    },
+    "request": {
+      "path": "/1/indexes/indexName/related-products/recommend/rules/objectID",
+      "method": "DELETE"
+    }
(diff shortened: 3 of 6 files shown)


specs/recommend/paths/getRecommendations.yml
```diff
@@ -4,21 +4,25 @@ post:
...
   x-use-read-transporter: true
   x-cacheable: true
-  summary: Get results.
-  description: Returns recommendations or trending results, for a specific model and `objectID`.
+  summary: Get recommendations or trending items.
+  description: >
+    Returns results from either recommendation or trending models:
+
+      - Recommendations are provided by the [Related Products and Related Content models](https://www.algolia.com/doc/guides/algolia-recommend/overview/#related-products-and-related-content)
+      - Trending models are [Trending Items and Trending Facet Values](https://www.algolia.com/doc/guides/algolia-recommend/overview/#related-products-and-related-content).
```
gazconroy, line 12:
```suggestion
      - Trending models are [Trending Items and Trending Facet Values](https://www.algolia.com/doc/guides/algolia-recommend/overview/#trending-items-and-trending-facet-values).
```

shortcuts:
checking the error

shortcuts (approved):
looks good, do we wait for the recommend team for some confirmation, or do we do that async?

francoischalifour:
> > We have the models: trending-items, bought-together, related-products, trending-facets. What about Looking similar
> > and Related content?
> 
> I guess cc @francoischalifour @anghelalexandra in case they have more context

Those are the only models Recommend supports for now ([see documentation](https://www.algolia.com/doc/ui-libraries/recommend/introduction/what-is-recommend/)).

PR #1688, opened 2023-07-04
fix(specs): add input to authentication list

millotp:
## 🧭 What and Why

The list authentications endpoint now return `input` too (although obfuscated)


shortcuts (approved):
great fix!

PR #1689, opened 2023-07-04
docs(specs): Review OpenAPI Analytics (search analytics)

gazconroy:
This is one of several PRs addressing the Analytics API.

This PR reviews just the search operations of the Analytics API.

### Questions and notes

In [the docs](https://www.algolia.com/doc/rest-api/analytics/#search-analytics), there are 15 API calls. In the OAS, just 12. Should the missing calls be added to the OAS?

#### getTopSearches

- In the response, `count` is always (much) smaller than `nbHits` (whatever the setting of `clickAnalytics`). This doesn’t seem intuitive. What’s the difference between `count` and `nbHits`?
- Response has a `clickPositions` object which isn’t defined in the OAS

#### getSearchesCountResponse

- Response has a `withFilterCount` field in the searches object which isn’t defined in the OAS

#### getNoResultsRate

- In the response, what is `noResultCount`? I’ve put in the placeholder “Number of occurrences” but that's not very helpful
- In the response, `noClickCount` is defined as “Number of click events.” This doesn’t seem to be correct in this context
getTopHits

#### getTopHits

- I’ve changed the description to “Returns the most popular clicked results in the latest 1,000 searches.” Is this accurate?
#### getTopFilterAttributes/getTopFilterForAttribute/getTopFiltersNoResults

- Does this apply to the “1,000 most used filters” (for all time?) or the “1,000 most recently used filters”. I’m assuming the latter but the former is the original wording

#### getTopFilterForAttribute

- In the response API calls I made, `operator` is returned as `:`. What does that indicate? What are the potential settings for `operator`?

#### getTopFiltersNoResults

- I’ve written “Returns top filters for filter-enabled searches that don’t return results.” Is that correct, though?
- If I don't specify search in the call, I’d either expect a large number of returned values (for all “no result” searches) or an error message saying search is necessary. Instead, I get `{"values": null}`

#### getTopCountries

- What does this do? Is this related to geo searching? To cluster location? To user IP address country correlation?
- The description says “Returns top countries. Limited to the 1,000 most frequent ones.” Most frequent country?
- Are there any prerequisites for this to work? When I make a call, I get `{"values": null}`

🎟 JIRA Ticket: [DOC-1112](https://algolia.atlassian.net/browse/DOC-1112)

### Changes included:

## 🧪 Test

[DOC-1112]: https://algolia.atlassian.net/browse/DOC-1112?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


specs/analytics/common/parameters.yml
```diff
@@ -37,73 +43,109 @@ direction:
...
     type: string
 
+Offset:
+  in: query
+  name: offset
+  description: Start position for retrieving results.
+  example: 1
+  schema:
+    type: integer
+    minimum: 0
```
shortcuts, line 76: https://github.com/algolia/api-clients-automation/blob/b0a25d59ed50c3283a983ad504b4a0ac1c32169a/specs/common/parameters.yml#L85

you can re-use the common one

shortcuts (approved):
awesome!

PR #1690, opened 2023-07-04
chore: add release using docker

aallam:
## 🧭 What and Why

This PR aims to streamline the process of creating release PRs for Docker images. 
We can now use the `yarn docker:release` command to create a release PR.

🎟 JIRA Ticket: APIC-699

### Changes included:

- Added `docker:release` command in `package.json`.
- Modified `Dockerfile` to install `git` during the Docker image build process.

_`GITHUB_TOKEN` environment variable is required to run the release command._

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "docker:build": "./scripts/docker/build.sh",
     "docker:clean": "docker stop api-clients-automation; docker rm -f api-clients-automation; docker image rm -f api-clients-automation",
     "docker:mount": "./scripts/docker/mount.sh",
-    "docker:release": "docker exec -e GITHUB_TOKEN -it api-clients-automation bash -lc \"yarn release\"",
+    "docker:release": "docker exec -e GITHUB_TOKEN=$GITHUB_TOKEN -it api-clients-automation bash -lc \"yarn release\"",
     "docker:setup": "yarn docker:clean && yarn docker:build && yarn docker:mount",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -16,6 +16,7 @@
     "docker:build": "./scripts/docker/build.sh",
     "docker:clean": "docker stop api-clients-automation; docker rm -f api-clients-automation; docker image rm -f api-clients-automation",
     "docker:mount": "./scripts/docker/mount.sh",
+    "docker:release": "docker exec -e GITHUB_TOKEN -it api-clients-automation bash -lc \"yarn release\"",
```
shortcuts, line 19: I can't remember but does the release script panics properly if the token is not provided?

shortcuts (approved):
that's nice!

package.json
millotp: I think so, there is a check at the top for that

millotp: should it be `-e GITHUB_TOKEN $GITHUB_TOKEN` ? I didn't know that syntax !

aallam: The script should fail yes! however, I found that git will look for the ssh config, so I've updated the mount script to use `~/.ssh` in read-only

aallam: Oh yes @millotp I think you are right!

millotp (approved):
very fancy

millotp (approved):

PR #1694, opened 2023-07-05
fix(specs): facet stats properties as double

aallam:
## 🧭 What and Why

Facets stats properties are defined as integers, but should be floating numbers.

### Changes included:

Change the type of the following facets stats properties (`max`, `min`, `avg` and `sum`)


shortcuts (commented):
wow I'm impressed Java passes, IIRC we had an issue with double, can you bump the cache key just to make sure we have a full CI run?

great find otherwise!

shortcuts (approved):
ok that's insane then 🤔

PR #1698, opened 2023-07-06
docs(specs): Review OpenAPI Search API (rules operations)

gazconroy:
## 🧭 What and Why

This is one of several PRs addressing the Search API.

This PR reviews just the rules operations of the Search API.

#### getRule

In the response, the description for query says:

“When providing a string, it replaces the entire query string. When providing an object, it describes incremental edits to be made to the query string (but you can't do both).”

However, since this is a response, the user can’t provide anything. What leads to the generation of a string in the response? What leads to the generation of an object in the response?

#### searchRules

If the body parameter `anchoring` is `endsWith` or `startsWith`, no responses are returned even if they should be. `contains` and `is` _do_ work.

#### saveRules

The upper and lower bounds of `validity` are required in what’s called Unix timestamp. Is this [Unix epoch time](https://en.wikipedia.org/wiki/Unix_time)? In seconds? Or milliseconds?

🎟 JIRA Ticket: [DOC-1107](https://algolia.atlassian.net/browse/DOC-1107)

### Changes included:

## 🧪 Test

[DOC-1107]: https://algolia.atlassian.net/browse/DOC-1107?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rules/clearRules.yml
shortcuts, line 1: Those new endpoints should be added to https://github.com/algolia/api-clients-automation/blob/main/specs/search/spec.yml#L149 also

rules/common/parameters.yml
shortcuts, line 1: Some of those parameters already exists in commons https://github.com/algolia/api-clients-automation/blob/main/specs/common/parameters.yml

specs/search/paths/rules/common/parameters.yml
```diff
@@ -4,4 +4,18 @@ ClearExistingRules:
...
+  description: Unique identifier of a rule object.
+  example: 'a-rule-id'
+  required: true
+  schema:
+    type: string
+
+query:
+  type: string
+  description: Rule object query.
+  default: ''
```
shortcuts, line 21: same for those, they seem duplicated

shortcuts (commented):
looks great, the CI will fail once the endpoints are added but I'll come back to add tests :)

rules/common/parameters.yml
shortcuts: yep the review comes from the same day, it make sense for objectID but the query is the same across any kind of objects I believe

non blocker anyway

shortcuts (commented):
do you need help adding the endpoints?

gazconroy:
@shortcuts  The endpoints I'm expecting are appearing in the preview. So, yes please, it would be useful to see what I'm missing

rules/clearRules.yml
shortcuts, line 1: I just realized but this rules folder is not at the right place, it should be in `specs/search`

shortcuts:
> @shortcuts The endpoints I'm expecting are appearing in the preview. So, yes please, it would be useful to see what I'm missing

- the rules folder is not at the right place (see https://github.com/algolia/api-clients-automation/pull/1698#discussion_r1266485546)
- since you added new files that contains new endpoints, they should be called from corresponding the `spec.yml` file (https://github.com/algolia/api-clients-automation/blob/main/specs/search/spec.yml#L149 in your case)

gazconroy:
I'd added an additional Rules folder at root level. Now removed

shortcuts (approved):
perfect!

PR #1699, opened 2023-07-06
docs(specs): Review OpenAPI Search API (synonym operations)

gazconroy:
## 🧭 What and Why

This is one of several PRs addressing the Search API.

This PR reviews just the synonym operations of the Search API.

#### saveSynonyms

You can add synonym objectIDs in the path parameter and the body parameter. It appears that the objectID in the body parameter takes precedence. Is that always the case?

🎟 JIRA Ticket: [DOC-1107](https://algolia.atlassian.net/browse/DOC-1107)

### Changes included:

## 🧪 Test

[DOC-1107]: https://algolia.atlassian.net/browse/DOC-1107?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/specs/search/paths/synonyms/common/parameters.yml
+++ b/specs/search/paths/synonyms/common/parameters.yml
@@ -3,12 +3,12 @@ ReplaceExistingSynonyms:
   name: replaceExistingSynonyms
   schema:
     type: boolean
-  description: Replace all synonyms of the index with the ones sent with this request.
+  description: If `true`, delete all synonyms in the index and then add the ones sent with this request.
 
 Query:
   in: query
   name: query
-  description: Search for specific synonyms matching this string.
+  description: Search for synonyms that match this string.
   schema:
     type: string
     default: ''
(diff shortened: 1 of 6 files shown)


specs/search/paths/synonyms/common/parameters.yml
```diff
@@ -3,29 +3,39 @@ ReplaceExistingSynonyms:
...
+  description: Synonym type.
+  example: onewaysynonym
   enum:
     - synonym
     - onewaysynonym
     - altcorrection1
     - altcorrection2
     - placeholder
+
+ObjectID:
```
shortcuts, line 34: we should use the common one

shortcuts (changes requested):
nice :)

specs/search/paths/synonyms/common/parameters.yml
shortcuts: https://github.com/algolia/api-clients-automation/blob/main/specs/common/parameters.yml#L11

millotp:
> It appears that the objectID in the body parameter takes precedence. Is that always the case ?

The body and path parameters are merged together for javascript, but maybe `openapi` merges then anyway because they have the same name ?

specs/search/paths/synonyms/common/parameters.yml
millotp: I think it's fine in this case, it's to improve documentation, and it doesn't create a new type (it's just translated to string), so there is no conflict.

shortcuts: I don't think there should be many definitions of an objectID

shortcuts: what changes is the context not the definition of the objectID 🤔 if you think it makes more sense then let's do it but I feel like it's more the responsibility of the documentation rather than the api spec definition

gazconroy: And there are also dictionary objects

shortcuts (approved):

PR #1731, opened 2023-07-11
chore: only setup dart when required

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-1420

### Changes included:

we setup dart for every clients but only need it for the release or the dart client itself


millotp (approved):
nice

PR #1732, opened 2023-07-11
chore: deps upgrade

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://github.com/algolia/api-clients-automation/issues/532

### Changes included:

deps upgrade


millotp:
we can't update redocly it seems like, it doesn't accept `nullable: true` but the alternative is not implemented in `openapi-generator` yet

millotp (approved):
lezz go

PR #1747, opened 2023-07-13
chore: refine renovate config

millotp:
## 🧭 What and Why

Using `renovate` locally I was able to make sure all `1926` versions we have in our monorepo are either tracked by renovate, or generated, for a total of of `373` tracked dependencies in `82` files !


renovate.json
```diff
@@ -134,28 +140,44 @@
...
   "ignorePaths": [
     "**/algoliasearch/**",
     "**/client-abtesting/**",
     "**/client-analytics/**",
     "**/client-insights/**",
+    "**/client_insights/**",
     "**/client-personalization/**",
     "**/client-query-suggestions/**",
     "**/client-search/**",
+    "**/client_search/**",
```
shortcuts, line 180: how come they are duplicated?

renovate.json
```diff
@@ -134,28 +140,44 @@
...
       "versioningTemplate": "{{#if versioning}}{{{versioning}}}{{else}}semver{{/if}}"
     }
   ],
+  "github-actions": {
+    "fileMatch": [
+      "\\.github/workflows/.*\\.ya?ml$"
+    ]
+  },
   "ignorePaths": [
     "**/algoliasearch/**",
```
shortcuts, line 172: does it allow negative patterns? maybe we can ignore everything in packages but the common ones

shortcuts (commented):
THIS IS HUGE

renovate.json
millotp: Unfortunately no it's only `glob`, not even regex, and the `includePaths` property doesn't work alongside it so we can't do like in `generation.config.mjs`

renovate.json
shortcuts: Ahhhhh

shortcuts (approved):
Gg!!

PR #1770, opened 2023-07-17
chore: optimize docker image

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

tried to make the dockerfile a bit more easier to navigate and reduce the size, it was 2.5gb before, now it's 1.5xgb


Dockerfile
```diff
@@ -1,64 +1,44 @@
...
-    curl \
-    zip \
-    unzip \
-    # python is used by nvm to install some packages
-    python3 \
-    git \
+# Global
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    curl zip unzip git \
+    php${PHP_VERSION} php${PHP_VERSION}-curl php${PHP_VERSION}-dom \
```
millotp, line 21: `PHP_VERSION` is empty at this point so it will just use the default version from apt, and the version we want doesn't exists that why we used the image.

millotp (commented):
Very nice improvement ! much more readable

Dockerfile
shortcuts: how come it's empty? build.sh feeds it so it should be good I believe

for the apt part IIRC it doesn't include the patch version, I'll double check

shortcuts: ahhhhh yes I was confused with envs 🥶 I'll check with the docker image how big it is, might even be easier to maintain actually

shortcuts: back to the original, much simpler to use the php image ahaha

aallam (dismissed):
Nice! GJ! 🚀

millotp (dismissed):
looks good !

millotp (approved):

PR #1771, opened 2023-07-17
chore: openapi metadata files

aallam:
## 🧭 What and Why

OpenAPI generators metadata files are pushed (and removed) from API clients repositories, although not useful for those repos.

### Changes included:

Ignore OpenAPI generators metadata files.

### DIFF
--- a/clients/algoliasearch-client-dart/.gitignore
+++ b/clients/algoliasearch-client-dart/.gitignore
@@ -28,3 +28,6 @@ pubspec.lock
 
 # Melos
 pubspec_overrides.yaml
+
+# OpenAPI
+**.openapi-generator*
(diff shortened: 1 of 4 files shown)


millotp (commented):
I think the file `.openapi-generator-ignore` is useful for generation, but we can replace it by editing the generator directly.

shortcuts (commented):
actually I know why JS ignore all of that, is because the folder structure is so different from what's proposed by the default generator that we've removed every supported files, and therefore don't need it.

---

maybe it's just a git issue? I know some . files are not always tracked but not sure why it added/removed them

Fluf22 (approved):
Let's go with this if there is no pushback

Fluf22:
_There was, indeed, pushback_

PR #1772, opened 2023-07-18
fix(scripts): esm release

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

See https://github.com/algolia/api-clients-automation/actions/runs/5584120552/jobs/10205172965, it seems like I forgot to convert the release script to esm.

Also use `--env-file` instead of inline `env` for the release, so that it matches what we recommend in the [release process](https://api-clients-automation.netlify.app/docs/contributing/release-process/)


Fluf22 (approved):
I feel like this will cause issues with our IDEs 😂

shortcuts:
> I feel like this will cause issues with our IDEs 😂

idk why they made esm usage so complicated 🤷🏻

PR #1774, opened 2023-07-18
fix(specs): search w/ hits & facets responses

aallam:
## 🧭 What and Why

Multi search endpoint support conducting both 'hits' and 'facets' search operations within the same call. Although this is already outlined in the specifications for requests, it is not yet included for responses.

### Changes included:

Update the endpoint to support both 'hits' and 'facets' for requests as well as responses.


specs/search/common/schemas/SearchForFacetValuesResponse.yml
shortcuts, line 1: this looks super cool as a workaround!

it's sad we can't une the discriminator field, is it because we need to specify the type?

shortcuts (commented):
awesome

specs/search/common/schemas/SearchForFacetValuesResponse.yml
aallam: Yes, indeed, the usual approach is to examine the value of the field, and from there, deduce its type. Unfortunately, we can't do that in this case. :/

shortcuts: what if the type was on its one and we pass it with a ref? would it be possible?

aallam: Can you give more detail? by pass in it with a ref, you mean in the specs ?

shortcuts: Yes but I guess here it's a matter of generation instead

specs/search/common/schemas/SearchForFacetValuesResponse.yml
```diff
@@ -0,0 +1,28 @@
+searchForFacetValuesResponse:
+  type: object
+  additionalProperties: false
+  required:
+    - facetHits
+  x-discriminator-fields:
+    - facetHits
```
shortcuts, line 7: considering our use of this field, can't we rely on the built-in one? 

```yml
discriminator:
  propertyName: facetHits
```

aallam: Regrettably, we can't rely on the built-in one in this context. The notion of a discriminator is to include type information within the object, such as type: SearchForFacetsResponse, but that's not present in our responses. In this case, we're checking for the existence of certain field(s), not their types. Furthermore, in anticipation of a day when proper type information will be included in the responses, it would be best to keep the semantics separate and reserve the discriminator for when the relevant fields become available.

shortcuts: :( oki as long as it works let's go with that and try to implement it in other clients too if possible!

the generator refactor might help with unifying all of that

shortcuts (approved):
awesome :D

PR #1775, opened 2023-07-18
fix(ci): setup dart deps for release

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

See https://github.com/algolia/api-clients-automation/actions/runs/5587125378/jobs/10212057849, we need to allow dart deps in the setup of the scheduled release to make it work

### DIFF
--- a/.github/workflows/scheduled-release.yml
+++ b/.github/workflows/scheduled-release.yml
@@ -24,6 +24,7 @@ jobs:
         uses: ./.github/actions/setup
         with:
           type: minimal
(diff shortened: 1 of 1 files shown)


millotp (approved):
interesting, maybe that's why `dart` was setup on every CI

shortcuts:
> interesting, maybe that's why `dart` was setup on every CI

yup, but this should be enough for the release and without impacting the rest

PR #1778, opened 2023-07-18
chore: prepare release 2023-07-18

shortcuts:
## Summary

This PR has been created using the `yarn release` script. Once merged, the clients will try to release their new version if their version has changed.

## Version Changes

- javascript: 5.0.0-alpha.73 -> **`prerelease` _(e.g. 5.0.0-alpha.74)_**
- java: 4.0.0-SNAPSHOT -> **`minor` _(e.g. 4.0.0-SNAPSHOT)_**
- php: 4.0.0-alpha.71 -> **`prerelease` _(e.g. 4.0.0-alpha.72)_**
- go: 4.0.0-alpha.20 -> **`prerelease` _(e.g. 4.0.0-alpha.21)_**
- kotlin: 3.0.0-SNAPSHOT -> **`minor` _(e.g. 3.0.0-SNAPSHOT)_**
- dart: 0.1.1+3 -> **`minor` _(e.g. 0.2.0)_**

### Skipped Commits

<p>It doesn't mean these commits are being excluded from the release. It means they're not taken into account when the release process figured out the next version number, and updated the changelog.</p>

<details>
  <summary>
    <i>Commits without language scope:</i>
  </summary>

  - docs: update release process (#1773)
- chore: optimize docker image (#1770)
- chore: refine renovate config (#1747)
- chore: deps upgrade (#1732)
- chore: only setup dart when required (#1731)
- chore: forward dart version (#1695)
- chore: add release using docker (#1690)
- docs: add migration guide for copyRules/settings/synonyms (#1686)
</details>

<details>
  <summary>
    <i>Commits with unknown language scope:</i>
  </summary>

  - fix(scripts): release script with ssh (#1777)
- fix(ci): setup dart deps for release (#1775)
- fix(scripts): esm release (#1772)
- chore(deps): dependencies 2023-07-17 (#1748)
- chore(renovate): don't use includePaths (#1745)
- chore(renovate): add regex for most dependencies (#1744)
- chore(deps): dependencies 2023-07-10 (#1705)
- fix(cts): generate lock and package.json for javascript (#1721)
- fix(scripts): unused imports in go tests (#1719)
</details>

PR #1779, opened 2023-07-18
fix(javascript): release process

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: - 

### Changes included:

fixes https://github.com/algolia/api-clients-automation/pull/1778

somehow the release process of the JavaScript client is broken (see https://github.com/algolia/api-clients-automation/pull/1778), it made me think that we could just leverage lerna for independant version bumping and remove any kind of JS-specific logic in the release bumping script.

to properly test this change, we can run `yarn docker:release` locally to see every packages getting bumped, then run the generation and cts generation to assert it works properly.

---

### detailed changes

- generators
  - [x] remove js specific stuff in the config files
  - [x] get package version from the `package.json`
  - [x] get package name and path to package via the `output` field of the `clients/openapitools.json` file
- scripts
  - [x] treat javascript packages as the other packages
  - [x] move `updateDartPackages` call to the `updateAPIVersions` method
  - [x] add `dart pub get` to prevent dart cache issue on release


config/clients.config.json
```diff
@@ -15,7 +15,7 @@
     "folder": "clients/algoliasearch-client-javascript",
     "npmNamespace": "@algolia",
     "gitRepoId": "algoliasearch-client-javascript",
-    "utilsPackageVersion": "5.0.0-alpha.73",
```
millotp, line 18: are you sure you want to rename that ? I think `utilsPackageVersion` was fine, especially since we use this variable name elsewhere still

millotp (commented):
looks good ! love the logic removal

config/clients.config.json
shortcuts: bah we were populating this version with the search client version, we know they will always be in sync so.. I thought it was easier

millotp: maybe rename it to `searchPackageVersion` ? just to differentiate from `ingestion` and the other

shortcuts: or wdyt if I go with `version` so it's broader and cover many cases?

for JS it only concerns the utils package, but for the other clients it concerns the package itself

shortcuts: https://github.com/algolia/api-clients-automation/pull/1779/commits/82c37d479a0ca9ea80c3a82bceffe60fe7f3fdf0 and https://github.com/algolia/api-clients-automation/pull/1779/commits/77f7e72f49364f817236139154c82cea6f02dc31

website/docs/contributing/release-process.md
```diff
@@ -36,53 +36,3 @@ After a full CI run, a release commit will be sent to every repository and sprea
 
 Each language repository should have their own release process, and should run only when the latest commit starts with `chore: release`. By doing so, we have a way to just update the repository, for example READMEs, without having to release.
 
-## Releasing manually
```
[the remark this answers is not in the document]
millotp: I think it could be nice to keep this as internal doc somewhere

clients/algoliasearch-client-dart/packages/client_core/lib/src/version.dart
```diff
@@ -1,2 +1,2 @@
 /// Current package version
-const packageVersion = '0.1.1+2';
+const version = '0.1.1+2';
```
millotp, line 2: this is not generated ? is this bumped by the release also ?

millotp (commented):
small comments

shortcuts:
ah Go is broken brb

shortcuts:
ok actually we've named it `packageVErsion` and not `version` because it already contains the API version 🤷🏻

shortcuts:
ok 0 changes now it's perfect, time to try a release and see if it works

PR #1780, opened 2023-07-19
fix(dart): custom calls path and deserialization

aallam:
## 🧭 What and Why

Custom calls  (.e.g. `post`) can encounter failure due to several factors, which include:
- Issues with path encoding
- Problems during response deserialization
- Absence of required request headers

### Changes included:

- Path encoding is now skipped
- Raw response is returned, rather than deserialized data
- The `Content-Type` header is now added by default if a request body is provided


shortcuts (approved):
make sense, great catch :)

PR #1781, opened 2023-07-19
fix(javascript): allow package.json to be pushed

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-1437

### Changes included:

now that we rely on lerna for version bumping, we should allow the package.json to be pushed, which allows us to let the generation read the updates when running `yarn docker:release`

alternatively, we could have this logic only for the release process I believe, but this would introduce much more logic change


Fluf22 (approved):
👌

PR #1782, opened 2023-07-19
fix(scripts): execa command for git commits

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

The spread logic is failing https://github.com/algolia/api-clients-automation/actions/runs/5598140574/jobs/10237573027, this is because the `execaCommand` doesn't seem to support multiline arguments, however the `execa` one does.

### DIFF
--- a/scripts/common.ts
+++ b/scripts/common.ts
@@ -2,7 +2,7 @@ import fsp from 'fs/promises';
 import path from 'path';
 
 import { Octokit } from '@octokit/rest';
-import { execaCommand } from 'execa';
+import { execaCommand, execa } from 'execa';
 import type { ExecaError } from 'execa';
 import { hashElement } from 'folder-hash';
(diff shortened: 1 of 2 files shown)


scripts/common.ts
```diff
@@ -154,7 +154,7 @@ export async function gitCommit({
     ? `${message}\n\n\n${coAuthors.join('\n')}`
     : message;
 
-  await execaCommand(`git commit -m "${messageWithCoAuthors}"`, { cwd });
+  await execa('git', ['commit', '-m', messageWithCoAuthors], { cwd });
```
millotp, line 157: can't you use the `run` util ?

shortcuts: it uses execaCommand too

millotp (dismissed):

millotp (approved):

PR #1784, opened 2023-07-19
fix(scripts): prevent branch exists error

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

release can fail if the branch already exists, we now force and reset the branch if it's the case.


Fluf22 (approved):
You find that's not useful to have this protection?

shortcuts:
> You find that's not useful to have this protection?

there's already some validation steps for the release to happen, and we already force delete the remote branch, so I believe that's a miss to not do it for local ones

millotp (approved):
nice

PR #1786, opened 2023-07-20
chore(dart): use melos to publish packages

aallam:
## 🧭 What and Why

Each package is published individually, which can pose a problem in cases of dependency between versions (since we currently use exact versions).

### Changes Included:

Use [`melos publish`](https://melos.invertase.dev/commands/publish) to automatically publish all packages.

### DIFF
--- a/clients/algoliasearch-client-dart/.github/workflows/publish.yml
+++ b/clients/algoliasearch-client-dart/.github/workflows/publish.yml
@@ -41,51 +41,25 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.ALGOLIA_BOT_TOKEN }}
       
     outputs:
-      PUBLISH_CORE: ${{ steps.version.outputs.core }}
-      PUBLISH_SEARCH: ${{ steps.version.outputs.search }}
-      PUBLISH_INSIGHTS: ${{ steps.version.outputs.insights }}
-      PUBLISH_ALGOLIASEARCH: ${{ steps.version.outputs.algoliasearch }}
+      PUBLISH: ${{ steps.version.outputs.publish }}
 
-  publish_algoliasearch:
+  publish:
     needs: version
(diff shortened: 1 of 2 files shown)


shortcuts (approved):
nice!

clients/algoliasearch-client-dart/.github/workflows/publish.yml
```diff
@@ -41,51 +41,25 @@ jobs:
...
       
     outputs:
-      PUBLISH_CORE: ${{ steps.version.outputs.core }}
-      PUBLISH_SEARCH: ${{ steps.version.outputs.search }}
-      PUBLISH_INSIGHTS: ${{ steps.version.outputs.insights }}
-      PUBLISH_ALGOLIASEARCH: ${{ steps.version.outputs.algoliasearch }}
+      PUBLISH: ${{ steps.version.outputs.publish }}
 
-  publish_algoliasearch:
+  publish:
```
millotp, line 46: would it make sense to merge all 3 jobs in the same one and just use more steps ? I feel like having a full job just to compute the version is a bit overkill, and because all jobs depends on each other it could work well in a single flow.
Anyway just rumbling it works fine like this !

millotp (approved):
nice !

clients/algoliasearch-client-dart/.github/workflows/publish.yml
shortcuts: _I think_ this is because the publish workflow uses a manual validation step

millotp: ahh then that make sense

aallam: I've structured it this way to prevent the repository from making a publish commit when there is nothing to publish. If such a scenario is impossible, then we can certainly eliminate this logic!

PR #1790, opened 2023-07-20
fix(playground): reflect spec changes

millotp:
## 🧭 What and Why

Add `kotlin` playground to the CLI with `yarn docker playground kotlin search`.

Also fix a bunch of issues in the playground, migration guide and example because the specs changed a while back.


playground/kotlin/src/main/kotlin/com/algolia/playground/Insights.kt
```diff
@@ -19,10 +19,10 @@ suspend fun main() {
     )
     val indexName = dotenv["SEARCH_INDEX"]
 
-    val events = InsightEvents(
+    val events = InsightsEvents(
         events = listOf(
-            InsightEvent(
-                eventType = EventType.Click,
+            EventsItems.ClickedObjectIDs(
+                eventType = ClickEvent.Click,
```
shortcuts, line 25: EventType felt better 🤔

playground/kotlin/bin/main/com/algolia/playground/AbTesting.kt
shortcuts, line 1: is the gitignore up to date?

shortcuts (commented):
lovely!

playground/kotlin/src/main/kotlin/com/algolia/playground/Insights.kt
millotp: here each enum has 1 option, so you can't make a mistake. In typescript it's just a string so it's nice

shortcuts: yup I was talking in term of naming, ClickEvent.Click is redundant IMO

playground/kotlin/bin/main/com/algolia/playground/AbTesting.kt
millotp: I don't know who generates those files (might be IntelliJ), but here running with the cli it's fine, I can add those files to gitignore yes

shortcuts (approved):

PR #1853, opened 2023-07-30
chore(deps): dependencies 2023-07-31

shortcuts:


aallam (dismissed):
All good! 
`sourceCompatibility` and `targetCompatibility` should be added in a separate PR to the Java API client module :) I will take care of it

aallam (approved):

PR #1854, opened 2023-07-31
refactor(java): normalize client repo structure

aallam:
## 🧭 What and Why

`algoliasearch-client-java-2` repo name and `algoliasearch-core` module name are a relics from the previous Java API client. The generated client don't need this anymore.

### Changes included:

- Rename the client directory name to match it git repo, from `algoliasearch-client-java-2` to `algoliasearch-client-java`
- Move from `algoliasearch-core` to `algoliasearch`:
  - Rename `algoliasearch-core` directory to `algoliasearch`
  - Update all config files to use `algoliasearch` directory.


shortcuts (commented):
Did you had to remove the files yourself? Normally if you changes everything that generates it, the bot should delete the others

aallam:
> Did you had to remove the files yourself? Normally if you changes everything that generates it, the bot should delete the others

Are you certain this will occur automatically? I was under the impression it wouldn't, given that the bot can't recognize the old paths once we alter them in the PR itself

shortcuts:
> > Did you had to remove the files yourself? Normally if you changes everything that generates it, the bot should delete the others
> 
> Are you certain this will occur automatically? I was under the impression it wouldn't, given that the bot can't recognize the old paths once we alter them in the PR itself

it is expected that these kind of changes are handled by the generator yes, IIRC we made it so the bot create a new empty folder and push the generated code to it, so that github sees the deleted files

if it doesn't work then it's a ticket we need to create

aallam:
> it is expected that these kind of changes are handled by the generator yes, IIRC we made it so the bot create a new empty folder and push the generated code to it, so that github sees the deleted files
> 
> if it doesn't work then it's a ticket we need to create

I believe it won't work, since even the pre-hook did't remove them (the bot can't guess which files to delete), but I will follow your suggestion for this

millotp:
looks like the CI didn't delete the old directory, I guess because the name changed in the scripts as well

millotp:
oops sorry I misclicked

aallam:
> looks like the CI didn't delete the old directory, I guess because the name changed in the scripts as well

Yes, that's what I mentioned earlier, the scripts deletes only the folder in the configs, which makes sense In my opinion!

millotp:
can you delete the old folder in your PR then pls ?

shortcuts:
I found a way to make the CI delete the code properly before generating but with JS it's a bit more complex since it supports independant generation... I guess the script will wait

millotp (approved):
looks great ! it's a bit concerning that 8000 lines are gone, must be from old files that never got deleted

aallam:
> looks great ! it's a bit concerning that 8000 lines are gone, must be from old files that never got deleted

it's mostly the generated code, it will be generated again in the correct path :D

PR #1855, opened 2023-07-31
fix(specs): allow JSON array for `aroundPrecision`

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

See [sandbox](https://codesandbox.io/s/morning-silence-tq67ps)

`aroundPrecision` can either be an integer or a JSON array.

Today [this is supported in the JS and C# clients](https://www.algolia.com/doc/api-reference/api-parameters/aroundPrecision/#set-non-linear-geo-search-precision), but not in the others, since the API allows it we should also reflect it in the specs.


specs/common/schemas/SearchParams.yml
```diff
@@ -232,6 +228,26 @@ aroundRadius:
...
+aroundPrecision:
+  description: Precision of a geographical search (in meters), to [group results that are more or less the same distance from a central point](https://www.algolia.com/doc/guides/managing-results/refine-results/geolocation/in-depth/geo-ranking-precision/).
+  oneOf:
+    - type: integer
+      default: 10
+    - $ref: '#/aroundPrecisionFromValue'
+  x-categories:
+    - Geo-Search
+
+aroundPrecisionFromValue:
```
shortcuts, line 240: @kai687 I did not found occurrences of documentation for this specific type, is there a place where I could find some info? If not, would you mind suggesting a description for the `from` and `value` parameters?

aallam:
Is one of the formats a legacy format ? if yes, maybe we don't have to support them all !

shortcuts:
> Is one of the formats a legacy format ? if yes, maybe we don't have to support them all !

I tried on the API and it's still there so I belive it's a different use case

PR #1856, opened 2023-07-31
fix(specs): allow partial input update for authentications

Fluf22:
## 🧭 What and Why

🎟 JIRA Ticket: [DI-1239](https://algolia.atlassian.net/browse/DI-1239)

### Changes included:

- auth input partial descriptions

[DI-1239]: https://algolia.atlassian.net/browse/DI-1239?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/specs/ingestion/common/schemas/authentication.yml
+++ b/specs/ingestion/common/schemas/authentication.yml
@@ -60,7 +60,7 @@ AuthenticationCreateResponse:
 
 AuthenticationUpdate:
   type: object
-  description: Payload to partialy update an Authentication.
+  description: Payload to partially update an Authentication.
   additionalProperties: false
   properties:
     type:
@@ -70,7 +70,7 @@ AuthenticationUpdate:
     platform:
       $ref: '#/Platform'
     input:
-      $ref: '#/AuthInput'
+      $ref: '#/AuthInputPartial'
 
 AuthenticationUpdateResponse:
   type: object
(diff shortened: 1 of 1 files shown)


specs/ingestion/common/schemas/authentication.yml
```diff
@@ -178,6 +166,33 @@ AuthAlgolia:
     apiKey:
       type: string
       description: Algolia API Key, with the correct rights to push to an index and change settings.
+
+AuthGoogleServiceAccount:
```
millotp, line 170: looks like this doesn't work, the `AuthInput` is using the `Partial` objects, I think it's because you cannot have anything next to a `$ref` prop and you have to use `allOf` instead, like:
```
AuthBasic:
  allOf:
    - $ref: '#/AuthBasicParial'
    - type: object
      required:
        - username
        - password
```

but I'm not sure if that works, and what it would do in other languages

Fluf22: Ah yes, you're right! Trying the `allOf` method to confirm

Fluf22: Works with JS locally! Let's see what it does on other clients

millotp (approved):
> make two separate object desc for each input type

that's what I feared, we have a similar pattern in other places and I'd love to find a better way of doing it, maybe with some custom logic at some point, but this works best for every language

shortcuts (approved):
gg!

Fluf22:
@millotp the best would be to just have `Partial<MyStruct>` in ts 😁 but how could this be replicated in other languages? 🤔🤷🏻‍♂️

PR #1859, opened 2023-08-01
fix(dart): correctly rethrow 4xx errors

aallam:
## 🧭 What and Why

The current error-throwing logic doesn't stop the retry operations, which it should.

### Changes included:

Correctly detect `4xx` errors to immediately stop retrying calls.

### DIFF
--- a/clients/algoliasearch-client-dart/packages/client_core/lib/src/transport/retry_strategy.dart
+++ b/clients/algoliasearch-client-dart/packages/client_core/lib/src/transport/retry_strategy.dart
@@ -61,7 +61,7 @@ final class RetryStrategy {
         host.failed();
         errors.add(e);
       } on AlgoliaApiException catch (e) {
-        if (e.statusCode / 100 == 4) rethrow;
+        if (e.statusCode ~/ 100 == 4) rethrow;
         host.failed();
         errors.add(e);
       }


clients/algoliasearch-client-dart/packages/client_core/lib/src/transport/retry_strategy.dart
```diff
@@ -61,7 +61,7 @@ final class RetryStrategy {
         host.failed();
         errors.add(e);
       } on AlgoliaApiException catch (e) {
-        if (e.statusCode / 100 == 4) rethrow;
+        if (e.statusCode ~/ 100 == 4) rethrow;
```
millotp, line 64: what in hell is this operator ahah

millotp (dismissed):
can't melos bump the version ? I think it works for non generated file too no ?

clients/algoliasearch-client-dart/packages/client_core/lib/src/transport/retry_strategy.dart
shortcuts: is that a divide but approximative that rounds? XD

aallam: Well, it's JS effect 😅
In dart, when you divide an `int` by and `int`, (e.g. `403 / 100`) guess what you get ? the answer: `4.03` 🤦‍♂️
The operator `~/` is to get an `int` result 🥲

shortcuts: 🤯 oh oki, I'd go with `e.StatusCode >= 400 && e.StatusCode < 500` in general but if it works like this then it's nice

shortcuts (approved):
noice

PR #1862, opened 2023-08-03
chore(dart): remove client packages

aallam:
## 🧭 What and Why

- Dart is a language used mainly (almost exclusively) in front end applications context.
- A few APIs are meant to be used from a secure environment.

### Changes included:

Remove `abtesting`, `analytics`, `monitoring`, `personalization`,`ingestion` and `query_suggestion` until a need arises.


millotp:
maybe you can leave `ingestion` ? it can be used on front or back, like the dashboard

config/openapitools.json
```diff
@@ -181,42 +181,6 @@
...
           "packageVersion": "0.1.2"
         }
-      },
-      "dart-personalization": {
-        "output": "#{cwd}/clients/algoliasearch-client-dart/packages/client_personalization",
-        "additionalProperties": {
-          "packageVersion": "0.1.2"
-        }
-      },
-      "dart-analytics": {
```
shortcuts, line 191: analytics also is front-end related stuff, no?

aallam: Although it's possible to use these APIs in the frontend, they are not typically utilized to build search experiences in frontend apps. However, we can always add them back if a need arises for some special case we are unaware of.

shortcuts (approved):
gg lets gooo

PR #1864, opened 2023-08-03
fix(specs): optional/required `query` and `params`

aallam:
## 🧭 What and Why

`query` and `params` are required for _Search_ (and _Browse_), but not for _Recommend_

### Changes included:

Set `query` and `params` as required for _Search_, and as optional for _Recommend_.

### DIFF
--- a/specs/search/common/schemas/SearchResponse.yml
+++ b/specs/search/common/schemas/SearchResponse.yml
@@ -32,7 +32,6 @@ baseSearchResponse:
     - hitsPerPage
     - processingTimeMS
     - exhaustiveNbHits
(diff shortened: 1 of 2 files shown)


specs/search/common/schemas/SearchResponse.yml
```diff
@@ -24,7 +32,6 @@ baseSearchResponse:
     - hitsPerPage
     - processingTimeMS
     - exhaustiveNbHits
-    - query
     - params
```
shortcuts, line 35:
```suggestion
```

this one should also be removed, right?

PR #1882, opened 2023-08-07
refactor(dart): unpin package dependency versions

aallam:
## 🧭 What and Why

During the alpha stage, package versions were strictly pinned. However, as we move to the beta stage, we will need to unpin these versions.

### Changes included:

- Unpin package dependency versions

For further information on version constraints, please refer to the [Dart documentation](https://dart.dev/tools/pub/versioning#version-constraints-the-dart-approach).


templates/dart/pubspec.mustache
shortcuts, line 1: do you expect releases to be desynchronized between those clients? otherwise I believe they will always be at the same stage

aallam: By the way, Dart handles package versions somewhat differently compared to other package managers such as gradle or npm.

shortcuts: yup I got the idea but since they are always bumped at the same time it made sense to me to keep this restriction

templates/dart/pubspec.mustache
```diff
@@ -16,11 +16,10 @@ environment:
...
-  algolia_client_core: 0.1.1+4
+  algolia_client_core: ^0.1.1+3
   {{#isAlgoliasearchClient}}
-  algolia_client_search: {{{searchVersion}}}
-  algolia_client_insights: {{{insightsVersion}}}
+  algolia_client_search: ^{{{searchVersion}}}
+  algolia_client_insights: ^{{{insightsVersion}}}
   {{/isAlgoliasearchClient}}
   json_annotation: ^4.8.1
   collection: ^1.17.1
```
shortcuts, line 25: btw should we pin for those? since we don't control it at all

aallam: No, we shouldn't. Pinning versions could make it difficult for users to use the library, or even make it unusable if they depend on another library that uses a different version. 
However, my answer would differ in the case of JVM libraries (for example, Java, Kotlin, etc.) as the dependency management there operates differently.

shortcuts (approved):
:okidoki:

templates/dart/pubspec.mustache
shortcuts: feels like they decided to go with a different logic than other languages... that's weird 🤔

PR #1885, opened 2023-08-08
chore(dart): add recommend version test

aallam:
## 🧭 What and Why

Each package has a version test that checks if the version in `pubspec.yaml` is the same as the version inside `version.dart`.

### Changes included:

Added version test for the package `algolia_client_recommend`.


Fluf22 (approved):
🆗

PR #1886, opened 2023-08-08
fix(dart): user agent for web

aallam:
## 🧭 What and Why

The platform information are already included in the `User-Agent` header

### Changes included:

- Simplify the web platform header


shortcuts (approved):
nice!

PR #1891, opened 2023-08-09
chore(java): use new gradle syntax

millotp:
## 🧭 What and Why

Gradle 9 will deprecate the `sourceVersion` so we need to use the new syntax.
Also disable the `generate client` cache on main to always build and run tests.

### DIFF
--- a/generators/build.gradle
+++ b/generators/build.gradle
@@ -7,8 +7,8 @@ version = '1.0.0'
 description = 'algolia-java-openapi-generator'
 
 java {
-  sourceCompatibility = JavaVersion.VERSION_18
-  targetCompatibility = JavaVersion.VERSION_18
+  sourceCompatibility = JavaVersion.VERSION_1_8
+  targetCompatibility = JavaVersion.VERSION_1_8
 }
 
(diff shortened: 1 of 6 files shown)


Fluf22 (approved):
I wonder how many devs had to make this kind of change because gradle decided to introduce a new syntax

shortcuts (approved):
noiceee

generators/build.gradle
```diff
@@ -5,7 +5,11 @@ plugins {
 group = 'org.openapitools'
 version = '1.0.0'
 description = 'algolia-java-openapi-generator'
-java.sourceCompatibility = JavaVersion.VERSION_1_8
+
+java {
+  sourceCompatibility = JavaVersion.VERSION_1_8
+  targetCompatibility = JavaVersion.VERSION_1_8
+}
```
aallam, line 12: I believe this is not necessary for the generators, actually, it matter only for the client's gradle config!

millotp: oh I didn't know, I'll remove it in the next PR :)

PR #1892, opened 2023-08-09
feat(cs): code snippets generation

aallam:
## 🧭 What and Why

Code snippets are generated based on the CTS. 
A new _(bundled)_ API specification is created, incorporating `x-meta.examples` into each endpoint.


scripts/cli/index.ts
```diff
@@ -155,6 +156,26 @@ ctsCommand
     await runCts(language === ALL ? LANGUAGES : [language]);
   });
 
+const csCommand = program.command('cs');
```
millotp, line 159: `snippets` is easier to understand I think

millotp (commented):
this is so nice !
wdyt about renaming every `cs` to `snippet` ? `cs` already mean a lot of things, and it's very close to `cts`

PR #1916, opened 2023-08-17
fix(scripts): aims at improving readability and reduce complexity

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-1491

### Changes included:

While coming back to the scripting side for a bit, I realized it was really complicated in some places, so I'll try to remove unused logic here and there, loose the cache key computing a bit more etc.

Also moves line length from 80 chars to 100, it was way too bulky with 80 IMO


scripts/ci/githubActions/setRunVariables.ts
```diff
@@ -50,65 +19,50 @@ export const COMMON_DEPENDENCIES = {
...
-  ],
-};
+export const DEPENDENCIES = LANGUAGES.reduce(
+  (finalDependencies, lang) => {
+    const key = `${lang.toUpperCase()}_CLIENT_CHANGED`;
+    const langFolder = getLanguageFolder(lang);
+    let langGenerator = capitalize(lang);
+
+    if (lang === 'javascript') {
+      langGenerator = 'JavaScript';
```
millotp, line 29: lol why did we do this to ourselves.
maybe renaming the generator to `AlgoliaJavascriptGeneror` makes more sense ?

scripts/ci/githubActions/setRunVariables.ts
```diff
@@ -50,65 +19,50 @@ export const COMMON_DEPENDENCIES = {
...
+    }
+
+    return {
+      ...finalDependencies,
+      [key]: [
+        // Files that are common to every clients, this is used to determine if we should generate the matrix for this job.
+        'config/generation.config.mjs',
+        'config/openapitools.json',
+        'config/clients.config.json',
+        'generators/src/main/java/com/algolia/codegen/Utils.java',
```
millotp, line 39: maybe `generators/src/main/java/com/algolia/codegen/GenericPropagator.java` also

scripts/ci/githubActions/createMatrix.ts
```diff
@@ -1,81 +1,65 @@
...
-      [lang]: {
-        path: getLanguageFolder(lang),
-        toRun: [],
-        cacheToCompute: [],
-      },
-    }),
-    {} as Record<Language, ToRunMatrix>
-  );
+async function createClientMatrix(baseBranch: string): Promise<void> {
+  const matrix: Record<string, ToRunMatrix> = {};
```
millotp, line 20:
```suggestion
  const matrix: Record<Language, ToRunMatrix> = {};
```
this is too precise ?

scripts/ci/githubActions/setRunVariables.ts
```diff
@@ -50,65 +19,50 @@ export const COMMON_DEPENDENCIES = {
...
-  ],
-  DART_CLIENT_CHANGED: [
-    ...CLIENTS_COMMON_FILES,
-    DART_CLIENT_FOLDER,
-    'config/.dart-version',
-    'templates/dart',
-    'generators/src/main/java/com/algolia/codegen/AlgoliaDartGenerator.java',
-  ],
-};
+export const DEPENDENCIES = LANGUAGES.reduce(
```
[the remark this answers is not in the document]
millotp: I think this is used in only one place, maybe it would be easier is this was just `getFilesDependencies(lang: Language): string[]` ?

scripts/ci/githubActions/createMatrix.ts
```diff
@@ -1,81 +1,65 @@
...
-    });
-    const clientChanges = await getNbGitDiff({
-      branch: baseBranch,
-      path: output,
-    });
-    const baseChanged = await isBaseChanged(baseBranch, {
-      ...COMMON_DEPENDENCIES,
+    const dependenciesChanged = await isBaseChanged(baseBranch, {
       ...languageDependencies,
+      GITHUB_ACTIONS_CHANGED: ['.github/actions', '.github/workflows', '.github/.cache_version'],
```
millotp, line 43: this breaks the `setRunVariables` function, this obj should be in the other file, here we don't care about the specific changes so you can just add the list of files

millotp (commented):
good idea !

scripts/ci/githubActions/setRunVariables.ts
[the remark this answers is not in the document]
millotp: ah no it's not the only place, there is `setRunVariables` too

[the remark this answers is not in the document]
shortcuts: yup I guess there's still room for improvements with this but so far it seems :okidoki:

scripts/ci/githubActions/setRunVariables.ts
shortcuts: ahah I did that at first https://github.com/algolia/api-clients-automation/pull/1916/commits/60a178cea7166b6b562cf6d3647d462fe9991559 but then idk... I'm fine with renaming for simplicity

scripts/ci/githubActions/setRunVariables.ts
shortcuts: make sense

scripts/ci/githubActions/createMatrix.ts
shortcuts: yup because the enum enforces the initialization of every keys even if it's empty :( so since we have 2 loops anyway it's good to keep a string

scripts/ci/githubActions/createMatrix.ts
shortcuts: ah true, added back here: https://github.com/algolia/api-clients-automation/pull/1916/commits/312c00598f501ae13ba3bebb89cf3c6a7c7a961a

scripts/.eslintrc.cjs
```diff
@@ -24,11 +22,24 @@ module.exports = {
...
-    '@typescript-eslint/sort-type-union-intersection-members': 0
+    'max-len': [0, { code: 100, ignoreComments: true }],
+    'prettier/prettier': [
+      'error',
+      {
+        singleQuote: true,
+        printWidth: 100,
+        trailingComma: 'es5',
+      },
+    ],
```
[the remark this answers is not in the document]
shortcuts: I took a look at the generate code and it's less bulky so I guess it's fine like that

millotp (approved):
GG ! I hope the behavior is the same

PR #1917, opened 2023-08-17
fix(scripts): remove deadcode

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-1491

### Changes included:

leverage https://github.com/nadeesha/ts-prune to find deadcode and remove it, also added `lint:deadcode` for future usage.


scripts/package.json
```diff
@@ -46,6 +47,7 @@
     "spinnies": "0.5.1",
     "ts-jest": "29.1.1",
     "ts-node": "10.9.1",
+    "ts-prune": "0.10.3",
```
millotp, line 50: `ts-prune` is still deprecated I don't thinks it's a good idea to add this as a dependency

shortcuts: yup I'm in the process of moving to an other tool but got other things to do in the meantime so... I'll leave this as a draft, unless you want to take it over?

shortcuts: I've pushed the progress to the branch in case you are interested

millotp: oh sorry I didn't check the date of the PR, no it's okay it's not urgent

millotp (commented):
this is so satisfying GG !

millotp (approved):
nvm it's already in the scripts ahah

PR #1934, opened 2023-08-22
feat(scripts): date monorepo released tags

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

somehow melos can't force pull tags which makes the script fail locally and requires us to update the tags manually

```
Unhandled exception:
ProcessException: Melos: Failed executing a git command:  From github.com:algolia/api-clients-automation
 ! [rejected]            released   -> released  (would clobber existing tag)
```

while we could automate this, it would be better if we had a tag for each release instead of overwriting it every time.

### DIFF
--- a/scripts/release/common.ts
+++ b/scripts/release/common.ts
@@ -3,8 +3,8 @@ import fsp from 'fs/promises';
 import config from '../../config/release.config.json' assert { type: 'json' };
 import { run } from '../common.js';
 
-export async function getLastReleasedTag(): Promise<string> {
-  return await run(`git describe --abbrev=0 --tags --match "${config.releasedTag}*"`);
+export function getLastReleasedTag(): Promise<string> {
+  return run(`git describe --abbrev=0 --tags --match "${config.releasedTag}*"`);
 }
 
 export function getNewReleasedTag(): string {
--- a/scripts/release/common.ts
+++ b/scripts/release/common.ts
@@ -1,8 +1,17 @@
 import fsp from 'fs/promises';
 
 import config from '../../config/release.config.json' assert { type: 'json' };
+import { run } from '../common.js';
 
-export const RELEASED_TAG = config.releasedTag;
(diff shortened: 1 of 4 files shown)


scripts/release/common.ts
```diff
@@ -1,8 +1,17 @@
 import fsp from 'fs/promises';
 
 import config from '../../config/release.config.json' assert { type: 'json' };
+import { run } from '../common.js';
 
-export const RELEASED_TAG = config.releasedTag;
+export async function getLastReleasedTag(): Promise<string> {
```
millotp, line 6: you can remove async and await from this function because you just return the promise

scripts/release/createReleasePR.ts
```diff
@@ -269,7 +269,7 @@ async function getCommits(): Promise<{
 }> {
   // Reading commits since last release
   const latestCommits = (
-    await run(`git log --pretty=format:"%h|%ae|%s" ${RELEASED_TAG}..${MAIN_BRANCH}`)
+    await run(`git log --pretty=format:"%h|%ae|%s" ${await getLastReleasedTag()}..${MAIN_BRANCH}`)
```
millotp, line 272: does it mean it can skip commit if this part fails once ?

scripts/release/common.ts
```diff
@@ -1,8 +1,17 @@
...
+  return await run(`git describe --abbrev=0 --tags --match "${config.releasedTag}*"`);
```
millotp, line 7: does this return all or only the last tag ?

millotp (approved):
good idea hopefully this works better ! nothing blocking

scripts/release/common.ts
shortcuts: only the last one, we can also add some sorting but it should be good by default

millotp (approved):
🦔

PR #1935, opened 2023-08-22
chore(java): rename artifact to `algoliasearch`

aallam:
## 🧭 What and Why

The current java client artifacts are published as `algoliasearch`, `algoliasearch-core`, `algoliasearch-apache`..etc.

### Changes included:

- Rename the artifact `algoliasearch` to match the current naming.


shortcuts (approved):
great catch!

PR #1964, opened 2023-08-31
fix(cts): improve oneOf matching with enum

millotp:
## 🧭 What and Why

🎟 JIRA Ticket: [DI-1384](https://algolia.atlassian.net/browse/DI-1384)

Fix oneOf matching for `EventItem`, by checking that the enum string is allowed.
It uses the tests cases added and removed in #1647.

Also clean up unused lambdas.

[DI-1384]: https://algolia.atlassian.net/browse/DI-1384?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


generators/src/main/java/com/algolia/codegen/cts/AlgoliaCTSGenerator.java
```diff
@@ -85,9 +86,6 @@ protected Builder<String, Lambda> addMustacheLambdas() {
     lambdas.put("escapeQuotes", new EscapeQuotesLambda());
     lambdas.put("escapeSlash", new EscapeSlashLambda());
     lambdas.put("replaceBacktick", new ReplaceBacktickLambda());
-    lambdas.put("escapeReserved", new EscapeDartLambda());
-    lambdas.put("escapeDollarSign", new EscapeDollarSignLambda());
-    lambdas.put("lowerCamelCase", new LowerCamelCaseLambda());
```
millotp, line 90: do you know the use case for those lambdas @aallam ?

shortcuts (commented):
INSANE

there's so many cases where this will allow us to write more tests

millotp:
yes this is like a discriminator, but automatic 😀

generators/src/main/java/com/algolia/codegen/cts/AlgoliaCTSGenerator.java
aallam: I remember some of them where used for dart, but I can't find the usage anymore! maybe got refactored away!

shortcuts (approved):
GGG!!

(just realized PHP tests are not formatted, that's weird)

millotp:
yeah I'm not sure why I'll try to make it work in another PR

PR #1986, opened 2023-09-06
fix(dart): remove description from enum

millotp:
## 🧭 What and Why

There is no specific description for enum variant, so the description was copied multiple time in each enum, was that the expected behavior @aallam ?


aallam:
> ## 🧭 What and Why
> There is no specific description for enum variant, so the description was copied multiple time in each enum, was that the expected behavior @aallam ?

No it shouldn't be there indeed! nice catch!

aallam (approved):

PR #1990, opened 2023-09-08
refactor(java): improve `oneOf` types DX

aallam:
## 🧭 What and Why

Currently, all `oneOf` types are built using a wrapper, even if no wrapper is needed. Also, generics are added for `oneOf` types too, although it doesn't make sense, especially if the underlying types are different (e.g., different records from each index).

### Changes included:

- Refactor the `oneOf` types to become interfaces, with concrete implementation when the `oneOf` type is a class or enum.
- Use wrappers in the case of primitives and collections.
- During deserialization, always start with types that have discriminator fields.
- Move from jetbrains `@NotNull`to javax `@Nonnull`


generators/src/main/java/com/algolia/codegen/GenericPropagator.java
```diff
@@ -13,10 +14,10 @@ public class GenericPropagator {
   private GenericPropagator() {}
 
   private static void setVendorExtension(IJsonSchemaValidationProperties property, String key, Object value) {
-    if (property instanceof CodegenModel) {
-      ((CodegenModel) property).vendorExtensions.put(key, value);
-    } else if (property instanceof CodegenProperty) {
-      ((CodegenProperty) property).vendorExtensions.put(key, value);
+    if (property instanceof CodegenModel model) {
+      model.vendorExtensions.put(key, value);
```
millotp, line 18: insane syntax !

playground/java/src/main/java/com/algolia/playground/Insights.java
```diff
@@ -26,12 +26,12 @@ public static void main(String[] args) throws Exception {
 
         var client = new InsightsClient(appId, apiKey, options);
         var params = new InsightsEvents();
-        var event = EventsItems.of(new ClickedObjectIDs()
+        var event = new ClickedObjectIDs()
```
millotp, line 29: this is so cool

generators/src/main/java/com/algolia/codegen/utils/OneOfUtils.java
```diff
@@ -0,0 +1,119 @@
...
+
+  public static void updateModelsOneOf(Map<String, ModelsMap> models, String modelPackage) {
+    for (ModelsMap modelContainer : models.values()) {
+      // modelContainers always have 1 and only 1 model in our specs
+      var model = modelContainer.getModels().get(0).getModel();
+      if (model.oneOf.isEmpty()) continue;
+      markOneOfChildren(models, model);
+      generateSealedChildren(models, modelPackage, model);
+      model.vendorExtensions.put("x-is-one-of", true);
+      model.vendorExtensions.put("x-one-of-explicit-name", Utils.shouldUseExplicitOneOfName(model.oneOf));
```
millotp, line 23: i think you can move the function `shouldUseExplicitOneOfName` to this file

templates/java/api_helpers.mustache
```diff
@@ -348,60 +322,9 @@ public Iterable<Rule> browseRules(String indexName, SearchRulesParams params) {
...
-}
-
-/**
-  * (asynchronously) Send multiple search queries to one or more indices.
-  *
-  * @param searchMethodParams Query requests and strategies. Results will be received in the same
-  *     order as the queries. (required)
-  * @throws AlgoliaRuntimeException If it fails to process the API call
-  */
-public CompletableFuture<SearchResponses<Object>> searchAsync(SearchMethodParams searchMethodParams, RequestOptions requestOptions)
```
millotp, line 365: this is not needed anymore ?

millotp (commented):
small review, I'm waiting for the code to generate, but it looks super cool !

templates/java/api_helpers.mustache
aallam: No need anymore, this same method is generated automatically now

generators/src/main/java/com/algolia/codegen/utils/OneOfUtils.java
aallam: Seems to be used also inside `Utils`

aallam:
> ticket

There is, in fact, `SearchResponse<T>` which remains generic, and there's `SearchResult`, which is the response used exclusively for multi-searches. In that context, generics don't make sense :D

millotp (approved):
oh okay then it's perfect !

PR #2003, opened 2023-09-12
fix(clients): user token character limit

kai687:
## 🧭 What and Why

The character limit for a user token in the Insights API is 129 characters.
I don't know when this was changed, but I tested to verify.

🎟 JIRA Ticket: N/A

### Changes included:

- Fix the `pattern` of the `userToken` parameter in the Insights API to allow up to 129 characters


millotp (approved):
what a weird limit, thanks !

PR #2005, opened 2023-09-12
fix(specs): introduced SourceUpdateDocker

Fluf22:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

To improve clarity and code readability, the 'SourceDocker' was renamed to 'SourceUpdateDocker'. The required property 'configuration' was moved to SourceUpdateDocker, while the 'imageType' property now exists in both the SourceUpdateDocker and SourceDocker. In addition, SourceDocker now includes a reference to SourceUpdateDocker. These changes ensure a more logical and understandable schema structure.

### DIFF
--- a/specs/ingestion/common/schemas/source.yml
+++ b/specs/ingestion/common/schemas/source.yml
@@ -266,6 +266,8 @@ SourceDocker:
   type: object
   additionalProperties: false
   properties:
+    imageType:
+      $ref: '#/DockerImageType'
     registry:
       $ref: '#/DockerRegistry'
     image:
@@ -276,8 +278,6 @@ SourceDocker:
       type: string
       description: The version of the image, defaults to `latest`.
       example: v2.1.0
-    imageType:
-      $ref: '#/DockerImageType'
     configuration:
       type: object
       description: The configuration of the spec.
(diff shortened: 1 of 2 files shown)


Fluf22:
Let's see if the `allOf` works as expected 😇

specs/ingestion/common/schemas/source.yml
```diff
@@ -276,11 +276,19 @@ SourceDocker:
...
+SourceDocker:
+  allOf:
+    - $ref: '#/SourceUpdateDocker'
+    - type: object
+      properties:
+        imageType:
+          $ref: '#/DockerImageType'
   required:
     - registry
     - image
```
shortcuts, line 294: This won't have effect in the allOf, it either needs to be in the update or redefined :/

specs/ingestion/common/schemas/source.yml
```diff
@@ -276,11 +276,19 @@ SourceDocker:
...
+  required:
+    - configuration
+
+SourceDocker:
+  allOf:
+    - $ref: '#/SourceUpdateDocker'
+    - type: object
+      properties:
+        imageType:
+          $ref: '#/DockerImageType'
```
shortcuts, line 291: Maybe as a standalone object with a define name? so that the generator don't append `AllOf`

shortcuts (commented):
noicee

Fluf22:
@shortcuts in the end, I made two different objects so I could control the required list of props.

shortcuts (approved):
make sense, might be worth to share properties but up to you

PR #2009, opened 2023-09-14
feat(scripts): sync dart client versions

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

Dart release process is broken, and since we now sync versions between all packages, we can keep the same package bumping logic as the other clients.

also close https://github.com/algolia/api-clients-automation/pull/1890


aallam (approved):
GJ! Looks much better!

PR #2010, opened 2023-09-14
fix(scripts): disable nx cache

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

nx cache seems broken, lets disable it for a moment until we find the issue

see all attempts https://github.com/algolia/api-clients-automation/actions/runs/6183577894/job/16788300465


Fluf22 (approved):
Let's go with that and check if it degrades build perf

PR #2012, opened 2023-09-14
fix(scripts): generate some core files

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

some `core` dart files were not generated on release, meaning their version is not bumped.

this PR adds them to the list so we don't have to do it manually anymore.

### DIFF
--- a/generators/src/main/java/com/algolia/codegen/AlgoliaDartGenerator.java
+++ b/generators/src/main/java/com/algolia/codegen/AlgoliaDartGenerator.java
@@ -62,13 +62,18 @@ public class AlgoliaDartGenerator extends DartDioClientCodegen {
     setPubLibrary(libName);
     setPubRepository("https://github.com/algolia/algoliasearch-client-dart/tree/main/packages/" + packageFolder);
 
+    // core package
+    additionalProperties.put("coreVersion", version);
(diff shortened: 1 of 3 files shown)


generators/src/main/java/com/algolia/codegen/AlgoliaDartGenerator.java
```diff
@@ -62,13 +62,18 @@ public void processOpts() {
...
+    supportingFiles.add(new SupportingFile("version.mustache", "../client_core/" + srcFolder, "version.dart"));
+
     // configs
     additionalProperties.put(CodegenConstants.SERIALIZATION_LIBRARY, SERIALIZATION_LIBRARY_JSON_SERIALIZABLE);
 
     super.processOpts();
 
     Arrays.asList("source", "get", "hide", "operator").forEach(reservedWords::remove); // reserved words from dart-keywords.txt
-
+    //
```
Fluf22, line 76: Is this expected?

Fluf22: :gottagofast:

Fluf22 (approved):

PR #2018, opened 2023-09-16
fix(dart): reset only hosts w/ call type

aallam:
### Changes included:

- Reset and use hosts for the call type only.


millotp (approved):
nice

PR #2029, opened 2023-09-18
feat(javascript): add batchSize to the observability event object

Fluf22:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-1521

### Changes included:

- add batchSize to the observability event object

### DIFF
--- a/specs/ingestion/common/schemas/event.yml
+++ b/specs/ingestion/common/schemas/event.yml
@@ -15,6 +15,12 @@ Event:
       $ref: '#/EventStatus'
     type:
       $ref: '#/EventType'
+    batchSize:
+      type: number
+      description: The extracted record batch size.
+      example: 10
+      minimum: 0
+      multipleOf: 1
     data:
       type: object
       additionalProperties: true
@@ -25,6 +31,7 @@ Event:
     - runID
     - status
     - type
+    - batchSize
     - publishedAt
 
 EventStatus:
--- a/specs/ingestion/common/schemas/event.yml
+++ b/specs/ingestion/common/schemas/event.yml
@@ -16,7 +16,7 @@ Event:
     type:
       $ref: '#/EventType'
     batchSize:
-      type: number
+      type: integer
       description: The extracted record batch size.
       example: 10
       minimum: 0


specs/ingestion/common/schemas/event.yml
```diff
@@ -25,6 +31,7 @@ Event:
     - runID
     - status
     - type
+    - batchSize
```
shortcuts, line 34:
```suggestion
```

it's not required we will default it

shortcuts (commented):
thanks!

specs/ingestion/common/schemas/event.yml
millotp: yes it's required that the API returns it

shortcuts: ah true it's not the insert here gg

Fluf22: Is it just me or is there an indentation issue? 🤔

shortcuts: seems good on my side!

Fluf22: image

millotp: tabs vs spaces on iphone is not the same size as on mac ?

Fluf22: Ah maybe! As long as it's good in reality 👌🏻

shortcuts (dismissed):
gg!

specs/ingestion/common/schemas/event.yml
```diff
@@ -15,6 +15,12 @@ Event:
       $ref: '#/EventStatus'
     type:
       $ref: '#/EventType'
+    batchSize:
```
shortcuts, line 18: ah we maybe need to say it's an integer, dart generates a double

specs/ingestion/common/schemas/event.yml
```diff
@@ -15,6 +15,12 @@ Event:
...
+      type: number
```
shortcuts, line 19:
```suggestion
      type: integer
```

should do it

specs/ingestion/common/schemas/event.yml
Fluf22: Ah, I'm always lost between number types.
That's why I put some validation, but it's better to use `integer` you're right.

shortcuts (approved):
lovely! gg

PR #2031, opened 2023-09-19
chore(dart): update ci packages directories

aallam:
### Changes included:

- CI: fix dart packages directories


shortcuts (approved):
oops good catch

PR #2032, opened 2023-09-19
refactor(kotlin): prepare beta release

aallam:
## 🧭 What and Why

Prepare Kotlin API client for beta phase.

### Changes included:

- Update changelog/version
- Stabilize the API
    - Simplify "oneOf" interfaces
    - Add Gzip support for JVM
    - Use `Duration` for all timeout params
- Misc: add missing gradle files


clients/algoliasearch-client-kotlin/client/src/commonMain/kotlin/com/algolia/client/configuration/internal/AlgoliaAgent.kt
```diff
@@ -36,21 +35,3 @@ internal class AlgoliaAgent(clientVersion: String) {
     }
   }
 }
-
-/**
- * A plugin that adds Algolia agent to all requests.
- *
- * @property agent `X-Algolia-Agent` header value.
- */
-internal fun algoliaAgent(agent: AlgoliaAgent) =
```
millotp, line 45: same here

generators/gradle/wrapper/gradle-wrapper.jar
millotp, line 1: maybe you can add those files to the root gitignore ? they shouldn't be pushed

templates/kotlin/tests/client/suite.mustache
```diff
@@ -42,7 +42,7 @@ class {{clientPrefix}}Test {
                 {{#match}}
                 {{#testUserAgent}}
                     val regexp = "{{#lambda.escapeSlash}}{{{match}}}{{/lambda.escapeSlash}}".toRegex()
-                    val header = it.url.parameters["X-Algolia-Agent"]?.decodeURLPart().orEmpty()
+                    val header = it.headers["User-Agent"].orEmpty()
```
millotp, line 45: all clients should use `X-Algolia-Agent`, can you edit the generation pls ?

aallam: I think "X-Algolia-Agent" is required only client that can be used for frontend (js & dartjs for example) since web browsers don't allow overriding `user-agent` headers

generators/gradle/wrapper/gradle-wrapper.jar
aallam: those are actually missing, and should be included in every gradle project, since all gradle projects don't share the same config

clients/algoliasearch-client-kotlin/client/src/commonMain/kotlin/com/algolia/client/extensions/SearchClient.kt
```diff
@@ -26,7 +27,7 @@ import com.algolia.client.transport.RequestOptions
 public suspend fun SearchClient.waitTask(
   indexName: String,
   taskID: Long,
-  timeout: Long? = null,
+  timeout: Duration? = null,
```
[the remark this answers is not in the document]
aallam: weird! those are not generated! it's maybe the cache!

templates/kotlin/tests/client/suite.mustache
millotp: oh right, I thought for consistency that we used `X-Algolia-Agent` for all langages but maybe `User-Agent` is fine too

shortcuts: don't we set both on backend clients? IIRC it's also for tracking purposes but I could be wrong

aallam: Not only is one enough, but tracking also occurs regardless of the method used. Most non-generated API clients use `User-Agent`, and the analytics are still collected.

shortcuts: nice then!

millotp (approved):
looks great !

PR #2033, opened 2023-09-19
fix(specs): `userData` type as 'any'

aallam:
## 🧭 What and Why

`userData` can be any valid json (i.g. array, map, string...)

### Changes included:

- Set `userData` type as "any"
- Update CTS generation*.

_*Note: For simplicity, CTS generator assumes that "any" types are maps_.


millotp:
🤔 java didn't generate any changes, the cache is really not working, can you bump the cache key to try pls ?

aallam:
> 🤔 java didn't generate any changes, the cache is really not working, can you bump the cache key to try pls ?

@millotp there is nothing to generate, the CTS was broken, but the final output is the same for java :D

millotp (approved):

PR #2053, opened 2023-09-28
feat(specs): add `attributesToExclude` for Ingestion

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-1501

### Changes included:

This PR adds the new `attributesToExclude` property to the Ingestion specs.

### DIFF
--- a/specs/ingestion/common/schemas/destination.yml
+++ b/specs/ingestion/common/schemas/destination.yml
@@ -124,6 +124,8 @@ DestinationIndexName:
       description: The index name to store data in.
     recordType:
       $ref: '#/RecordType'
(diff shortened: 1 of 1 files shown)


millotp (approved):
nice

shortcuts:
we need to find a way to keep newlines in the comments display :( cc @aallam if you have any idea

aallam:
> we need to find a way to keep newlines in the code docs :( cc @aallam if you have any idea

@shortcuts for multi lines entries, I think using `|` (instead of `>`) keeps the newlines

PR #2078, opened 2023-10-02
feat(specs): Remove attributeForDistinct from the indexSettingsAsSearc…

clemfromspace:
## 🧭 What and Why

`attributeForDistinct` is a settings-only parameter and shouldn't be present in the `search` / `browse` params.


Fluf22 (approved):
Looks good to me!

PR #2106, opened 2023-10-11
fix(javascript): release process

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket:

### Changes included:

there was a change in this file which made the release script unable to push package.json, this means versions became unsync'd at some point.

this should do the trick to allow release process changes, but prevent it in normal


Fluf22 (approved):
This is becoming a work of art. A masterpiece.

PR #2113, opened 2023-10-12
fix(specs): remove duplicate objectID

millotp:
## 🧭 What and Why

this removes the warning discovered by @shortcuts, two parameters were named `objectID` and redocly didn't like that


shortcuts (approved):
nice!! so weird it didn't threw an error

millotp:
> so weird it didn't threw an error

it's because it's just a path parameter, it doesn't generate a model like other objects, so the name is transparent

PR #2114, opened 2023-10-13
fix(playground): update go recommend test

Fluf22:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

- Because it's a pain in the 🍑 when trying to run the go playground.


shortcuts (approved):
lovely ty!

PR #2115, opened 2023-10-13
refactor(generators): update openapi generator to v7

aallam:
### Changes included:

Update openapi generator to [v7](https://github.com/OpenAPITools/openapi-generator/releases/tag/v7.0.1).


generators/src/main/java/com/algolia/codegen/AlgoliaDartGenerator.java
```diff
@@ -167,24 +167,14 @@ private void removeModels(String libName, Map<String, ModelsMap> modelsMap) {
       ModelsMap modelContainer = entry.getValue();
       ModelMap modelMap = modelContainer.getModels().get(0);
       CodegenModel model = modelMap.getModel();
-      if (!model.oneOf.isEmpty()) {
-        String classname = modelMap.getModel().classname;
-        oneOfs.put(classname, new OneOfMetadata(asImport(libName, classname), model.oneOf));
-        iterator.remove();
-        continue;
+      if (model.name.startsWith("base")) {
```
shortcuts, line 170: is this to prevent generating unused models?

shortcuts (commented):
I wasn't expecting such few changes, lovely!!

(waiting to check gen before approval)

shortcuts:
looking at https://github.com/algolia/api-clients-automation/commit/3a15ddb2cb69936c56bbb53cda2bd80e64d5d7f2#diff-36c28ddaac9898fc90774bc7798a40fe7612c859019174fb72f8e913ba3d7e22 it seems like we've lost some props definition in allofs?

generators/src/main/java/com/algolia/codegen/AlgoliaDartGenerator.java
aallam: Oh this line could be removed actually!

millotp:
> we've lost some props definition in allofs

yes the `QuerySuggestionsConfigurationResponse` is not strictly typed like before, it replaces the 2nd allOf with `Record<string, any>`, this was the same issue when I tried to upgrade

shortcuts:
> > we've lost some props definition in allofs
> 
> yes the `QuerySuggestionsConfigurationResponse` is not strictly typed like before, it replaces the 2nd allOf with `Record<string, any>`, this was the same issue when I tried to upgrade

so the generator assumes that if every props are optional, any props can be passed? that looks like something configurable 🤔

aallam:
> Maybe something changed in the templates in v7, but it's very hard to merge the changes

I think this can be fixed at the specification level by moving the object declaration away from the "allOf" section

shortcuts (commented):
looks wonderful gg!

specs/bundled/README.md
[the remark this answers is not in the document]
aallam: > looks great ! A type is renamed so it's a breaking change for dart, let's go dart@v2 !

We don't publish query suggestions client for dart :D (for now)

> I think this file can stay

Sure, let's will keep it :)

millotp (approved):

PR #2117, opened 2023-10-13
Feat/go wait task

Fluf22:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

- add `WaitForTask` and `WaitForApiKey` helpers
- add WaitErrs
- add docs

## 🧪 Test
Tested in the playground.


templates/go/api.mustache
```diff
@@ -336,3 +345,266 @@ func (c *APIClient) {{nickname}}WithContext(ctx context.Context, {{#hasParams}}r
...
+/*
+WaitForTaskWithContext waits for a task to be published.
+It returns the task response if the operation was successful.
+It returns an error if the operation failed.
+
+The maxRetries parameter is the maximum number of retries.
+The initialDelay parameter is the initial delay between each retry.
+The maxDelay parameter is the maximum delay between each retry.
+
+  @param indexName string - Index name.
```
Fluf22, line 397:
```suggestion
  @param ctx context.Context - The context that will be drilled down to the actual request.
  @param indexName string - Index name.
```

templates/go/api.mustache
```diff
@@ -336,3 +345,266 @@ func (c *APIClient) {{nickname}}WithContext(ctx context.Context, {{#hasParams}}r
...
+  - "update": wait for the API key to be updated
+
+The maxRetries parameter is the maximum number of retries.
+The initialDelay parameter is the initial delay between each retry.
+The maxDelay parameter is the maximum delay between each retry.
+
+If the operation is "update", the apiKey parameter must be set.
+If the operation is "delete" or "add", the apiKey parameter is not used.
+
+  @param key string - API key.
```
Fluf22, line 494:
```suggestion
  @param ctx context.Context - The context that will be drilled down to the actual request.
  @param key string - API key.
```

Fluf22 (commented):
I forgot to document the context props

playground/go/search.go
```diff
@@ -11,13 +11,93 @@ func testSearch(appID, apiKey string) int {
...
+		nil,
+		nil,
+		nil,
+	)
+	if err != nil {
+		panic(err)
+	}
+
+	printResponse(taskResponse)
+
```
aallam, line 82: is this duplicated code ?

templates/go/api.mustache
```diff
@@ -336,3 +345,268 @@ func (c *APIClient) {{nickname}}WithContext(ctx context.Context, {{#hasParams}}r
...
+
+        if apiKey.GetValidity() != response.GetValidity() {
+          return false
+        }
+
+        if apiKey.GetValidity() != response.GetValidity() {
+          return false
+        }
+
+        slices.Sort(apiKey.Acl)
```
aallam, line 558: I think `slices.Sort` and `slices.Equal` were added version 1.21 :/

aallam (changes requested):
GJ 🙌

templates/go/api.mustache
Fluf22: You're right 😭 
I'm gonna replace the `slices.Sort` by `sort.Slice(response.Acl, func(i, j int) bool { return response[i] < response[j] })`, for example, and I still have to think about a way to check the equality in an elegant and efficient way.

shortcuts: maybe it's also fine to bump the client to 1.21, we don't yet know when the go client will be released as stable

aallam: I wouldn't mind, but I think [we can't](https://algolia.atlassian.net/wiki/spaces/DA/pages/4506288343/RFC-25%3A+Generated+Golang+API+Client?focusedCommentId=4509139021) :/

shortcuts: we should re-export those numbers before deciding, we evaluated this in march and .20 was released in feb

Fluf22: For the record, here is the link to the chart where we can get the list of Go version by usage: https://app.hex.tech/ec8d7adc-d435-47ec-841b-ff2cff6432a3/hex/2cf5bff6-7ce1-4e10-b158-b898fffb5fa0/draft/logic (thanks @aallam for that)

According to it, we decided to pin the Go version to 1.21!

Fluf22: For those who don't have access to the chart, here is a cleaned up image: 
image

aallam (approved):
Let's GO! 🚀

playground/go/search.go
```diff
@@ -11,13 +11,93 @@ func testSearch(appID, apiKey string) int {
...
 
-	printResponse(searchResponse)
+	apiKeyStruct := search.NewApiKey([]search.Acl{"search"})
+
+	addApiKeyResponse, err := searchClient.AddApiKey(searchClient.NewApiAddApiKeyRequest(apiKeyStruct))
+	if err != nil {
+		panic(err)
+	}
+
+	taskResponse, err := searchClient.WaitForApiKey(
```
shortcuts, line 27: we definitely need golangci lint ahah

shortcuts (approved):
GG!!

PR #2145, opened 2023-10-17
fix(go): fix the spec and update the base model template

Fluf22:
## 🧭 What and Why

🎟 JIRA Ticket: [DI-1604]

### Changes included:

- fix the SearchResponse spec
- update the base model template to use pointer wherever applicable to leverage validation for required properties

## 🧪 Test
Updated the playground

[DI-1604]: https://algolia.atlassian.net/browse/DI-1604?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


templates/go/model_simple.mustache
```diff
@@ -12,12 +12,9 @@ type {{classname}} struct {
...
 {{#vars}}
 {{#description}}
-	// {{{.}}}
+	// {{#deprecated}}Deprecated. {{/deprecated}}{{{.}}}
 {{/description}}
-{{#deprecated}}
-	// Deprecated
-{{/deprecated}}
-	{{name}} {{^required}}{{^isNullable}}{{^isArray}}{{^isFreeFormObject}}*{{/isFreeFormObject}}{{/isArray}}{{/isNullable}}{{/required}}{{{dataType}}} `json:"{{baseName}}{{^required}},omitempty{{/required}}"{{#required}} validate:"required"{{/required}}{{#withXml}} xml:"{{baseName}}{{#isXmlAttribute}},attr{{/isXmlAttribute}}"{{/withXml}}{{#vendorExtensions.x-go-custom-tag}} {{{.}}}{{/vendorExtensions.x-go-custom-tag}}`
+	{{name}} {{^isNullable}}{{^isArray}}{{^isFreeFormObject}}*{{/isFreeFormObject}}{{/isArray}}{{/isNullable}}{{{dataType}}} `json:"{{baseName}}{{^required}},omitempty{{/required}}"{{#required}} validate:"required"{{/required}}{{#vendorExtensions.x-go-custom-tag}} {{{.}}}{{/vendorExtensions.x-go-custom-tag}}`
```
millotp, line 17: I think having pointer everywhere makes the client more complicated to use, if this is just to make the validator happy isn't it better to remove `validate:"required"` ? it's also future proof in case the api doesn't send back one property

millotp (commented):
thanks for fixing the spec !

templates/go/model_simple.mustache
millotp: maybe just remove it for types where the zero value is a valid one, like `number` or `bool` (maybe string)

Fluf22: I'm not sure it's more complicated, because it also generates getters and setters by default for the properties.

I don't want to mixup the validation and the API having more control over the data schema than the client. They are two different issues that I want to treat separately. Btw, `additionalProperties` seems to be a more appropriate solution for this problem.

Although, it would not be smart to put everything as a pointer and revert it just after...

Would removing `validate:"required"` be a rock-solid solution? I'm not keen to remove features that are useful just for avoiding a bug. 😄 Using pointers everywhere seems a little bit overwhelming for all the stars it adds up, but if it allows more flexibility for validating, I think we should consider keeping this as a solution.

Fluf22: @aallam @shortcuts since the conversation started again here, feel free to add your thoughts 😁

millotp: looking at other clients, and the base template in [openapi-generator](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/go/go.mod.mustache), it was never intended to have validation, it was added in this PR https://github.com/algolia/api-clients-automation/pull/1509/files#diff-21fc92b0f6770eea608468cad67e92c0e493fa6a0025d0819b5b3c244598ea05R316 just for the CTS so I think it's safe to rollback and remove the validator all together

Fluf22: The more I read what you both wrote, the more you are convincing me we shouldn't validate the schemas...

Although, I'm not sure it resolves the question of having pointers everywhere 😁 
I also now think you are right about not needing pointers everywhere: if the API require properties, it should send some value, so we shouldn't have zero-defaulted ones...

I think it safe to go with your solution: not using pointers and removing validation.
If there aren't any other objections, I will go with that.

Fluf22: I had to skip a test for Go, for now.

Removing the validators dug up another issue with tests, where we leverage marshalling JSON to setup mock objects quickly.

For instance, "SearchForHit" expects only an "indexName," but "SearchForFacet" expects an "indexName," a "facet," and a "type." However, since it initializes all three properties by default, it validates with "SearchForFacet," even though the JSON only contains an "indexName" since it's intended to be marshaled as "SearchForHit."

generators/src/main/java/com/algolia/codegen/utils/OneOfUtils.java
```diff
@@ -113,6 +113,20 @@ private static boolean isNumberType(String typeName) {
...
+      if (hasDiscriminatorA && hasDiscriminatorB) {
+        List<?> discriminatorsA = (List<?>) mapA.get("discriminators");
+        List<?> discriminatorsB = (List<?>) mapB.get("discriminators");
+
+        // Compare the lengths of the lists
+        int lengthComparison = Integer.compare(discriminatorsB.size(), discriminatorsA.size());
+
+        if (lengthComparison != 0) {
+          return lengthComparison;
+        }
```
aallam, line 126: _nitpick_: I think having simply this would make it work too:
```suggestion
        return discriminatorsB.size() - discriminatorsA.size();
```

generators/src/main/java/com/algolia/codegen/AlgoliaGoGenerator.java
```diff
@@ -20,6 +23,7 @@ public String getName() {
   public void processOpts() {
     String client = (String) additionalProperties.get("client");
     additionalProperties.put("enumClassPrefix", true);
+    //     additionalProperties.put("useOneOfDiscriminatorLookup", true);
```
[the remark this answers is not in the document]
Fluf22: Yes ofc, thanks!!

generators/src/main/java/com/algolia/codegen/utils/OneOfUtils.java
Fluf22: Hmm, yes, I think not being used to the language made me a bit confused, here 😆

Fluf22:
Great question about the `int` choice!
We should just ask confirmation form @mehmetaligok , maybe there was a special reason for it.

Also, I think this can be tackled in a new PR ☺️

templates/go/model_simple.mustache
Fluf22: Tests have been enabled again, everything should be back to normal!

aallam:
> Great question about the `int` choice! We should just ask confirmation form @mehmetaligok , maybe there was a special reason for it.

I think it is just the [default type](https://github.com/OpenAPITools/openapi-generator/blob/bf9fae641e178391d0539b7c9232cef61b2ece6a/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/AbstractGoCodegen.java#L110) in the generator, that's all.

> Also, I think this can be tackled in a new PR

Yes, for sure!

aallam (approved):

PR #2147, opened 2023-10-18
fix(specs): Change trigger input for task update

damcou:
## 🧭 What and Why

🎟 JIRA Ticket: DI-1601

### Changes included:

- Change trigger input for task update to allow changing cron


shortcuts (approved):
Looks good ty!

PR #2152, opened 2023-10-19
chore(dart): promote client to stable

aallam:
## 🧭 What and Why

- General Availability: Promote Dart client to v1.0.0 (Stable) 🎉

### DIFF
--- a/config/clients.config.json
+++ b/config/clients.config.json
@@ -63,7 +63,7 @@
   "dart": {
     "folder": "clients/algoliasearch-client-dart",
     "gitRepoId": "algoliasearch-client-dart",
-    "packageVersion": "0.6.0",
+    "packageVersion": "1.0.0",
     "modelFolder": "lib/src/model",
     "apiFolder": "lib/src/api",
     "customGenerator": "algolia-dart",


shortcuts (commented):
lets gooooo

we should drop https://github.com/algolia/api-clients-automation/blob/94306a84228eea151edc50e1fdb12eb3f25cd3e6/clients/algoliasearch-client-dart/packages/algoliasearch/lib/src/extension.dart#L30 since it's a major, wdyt?

aallam:
> lets gooooo
> 
> we should drop
> 
> https://github.com/algolia/api-clients-automation/blob/94306a84228eea151edc50e1fdb12eb3f25cd3e6/clients/algoliasearch-client-dart/packages/algoliasearch/lib/src/extension.dart#L30
> 
> since it's a major, wdyt?

Yes true! let's do that!

shortcuts:
> > btw you can remove the manual upgrade, the release script detects it if you add `BREAKING CHANGE` to the merge commit
> 
> Do you mean it will update to 1.0.0 automatically ?

yup it should create the changelog etc. with a major bump

PR #2184, opened 2023-10-26
chore(scripts): upgrade node and yarn

millotp:
## 🧭 What and Why

Upgrade to node 20 and yarn 4.
A lot had to change for this PR, mainly the tests because `ts-node` and `ts-jest` are not maintained and doesn't work with node 20, so I got rid of them.
- When possible we use `tsc && node script.js` or `tsc && jest`
- When the typescript is too hard for jest (with import assertions for example), we use `babel` to transpile the ts before feeding it to `jest`
- The scripts now use `vitest` because I couldn't get `jest` working with all the mocks

`fetch` is not behaving the same locally and on the CI, so I got rid of the call to `google.com:81` and use the local server instead, which is more reliable.

`slice` is deprecated so we use `subarray` now.

Updated the js playground.

Note: `tsc` is not very fast, it would be nice to use `bun` or `swc-node` to be blazing fast⚡


millotp:
no ahah, I don't know why they are not upgrading, because github actions are now shipped with node20 by default

millotp:
actually setup-node is working

clients/algoliasearch-client-javascript/packages/algoliasearch/babel.config.cjs
shortcuts, line 1: do we still need this?

clients/algoliasearch-client-javascript/packages/algoliasearch/jest.config.cjs
shortcuts, line 1: same we don't need those anymore, right?

shortcuts (commented):
loookksss super clean!

clients/algoliasearch-client-javascript/packages/algoliasearch/babel.config.cjs
millotp: I couldn't get it to work with tsc for all js packages, because they use weird builds, so I had to use jest with babel :( it add a dev dependency, so maybe it's fine

shortcuts (approved):
looks AWESOME

gg for tackling this

PR #2185, opened 2023-10-26
fix(javascript): algoliaseach bundling

millotp:
## 🧭 What and Why

While trying to upgrade node I noticed that the playground and algoliasearch was broken, thanks @shortcuts for the help


shortcuts (approved):
gggggg much cleaner!

PR #2229, opened 2023-11-09
feat(specs): add `startDate` and `endDate` query parameter to list events

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

events now allow the `startDate` and `endDate` query parameters to be passed.


Fluf22 (approved):
YES THANK YOU

PR #2255, opened 2023-11-13
chore(specs): forward tags in bundled specs

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

tags are not included in the bundled specs, which makes it unusable for the dx team to leverage it on the new docs


millotp (approved):
nice

PR #2285, opened 2023-11-23
fix(python): template cleanup and playgrounds

shortcuts:
## 🧭 What and Why

### Changes included:

follow up of https://github.com/algolia/api-clients-automation/pull/2283

This PR removes some of the unused stuff of the templates (librairies, imports, some debugging methods etc.)

Also adds the playground, which allowed me to properly setup the imports.

The next PR will focus on the requester using aiohttp for asynchronous calls.


config/.python-version
```diff
@@ -1 +1 @@
-3.12.0
+3.11.6
```
shortcuts, line 1: pep517 builds are broken with aiohttp on 3.12 so I had to downgrade for now

clients/algoliasearch-client-python/pyproject.toml
```diff
@@ -15,11 +15,13 @@ classifiers = [
 
 [tool.poetry.dependencies]
 python = "^3.8.1"
-requests = "2.31.0"
+urllib3 = ">= 1.25.3"
+aiohttp = ">= 3.9.0"
+python-dateutil = ">= 2.8.2"
+pydantic = ">= 2"
+typing-extensions = ">= 4.7.1"
```
shortcuts, line 22: those are typing dependencies, for now I've kept them since they are included in the template but if it becomes cumbersome to have them required for the user I'll loose the types a bit

clients/algoliasearch-client-python/pyproject.toml
```diff
@@ -30,6 +32,13 @@ isort = "5.12.0"
 max-line-length = 88
 ignore = ['E501']
 
+[tool.isort]
+line_length=88
```
millotp, line 36: why 88 ahah
i think we use 100 or 120 in other languages

clients/algoliasearch-client-python/requirements.txt
```diff
@@ -0,0 +1,5 @@
+argcomplete==3.1.6
```
millotp, line 1: what's the diff between `requirements.txt` and `pyproject.yml` ?

clients/algoliasearch-client-python/setup.cfg
```diff
@@ -0,0 +1,44 @@
...
+    Topic :: Internet :: WWW/HTTP
+    Topic :: Internet :: WWW/HTTP :: Indexing/Search
+    Development Status :: 5 - Production/Stable
+
+[options]
+zip_safe = True
+include_package_data = True
+packages = find:
+install_requires =
+    urllib3>=1.25.3,<2.0
```
millotp, line 39: can't we have 1 file listing all deps ahah

clients/algoliasearch-client-python/pyproject.toml
shortcuts: ahah it's the python norm, it was 79 before and now 88 is more widely used

clients/algoliasearch-client-python/requirements.txt
shortcuts: requirements is the deps installed by pip (for dev mostly), pyproject is more like the package.json when you ship the package, it allows you to define what build system, the dependencies, some scripts etc. but when users install your package, if they use poetry it will respect what you've defined in it

clients/algoliasearch-client-python/setup.cfg
shortcuts: I think we can go with poetry only but I wanted to cover the most common tools just for the sake of the users journey

PR #2286, opened 2023-11-24
fix(python): template cleanup and playgrounds

shortcuts:
## 🧭 What and Why

### Changes included:

follow up of https://github.com/algolia/api-clients-automation/pull/2283, initially done in https://github.com/algolia/api-clients-automation/pull/2285

This PR removes some of the unused stuff of the templates (librairies, imports, some debugging methods etc.)

Also adds the playground, which allowed me to properly setup the imports.

The next PR will focus on the requester using aiohttp for asynchronous calls.


clients/algoliasearch-client-python/setup.py
```diff
@@ -0,0 +1,4 @@
+import sys
+
+if sys.version_info < (3, 8):
```
millotp, line 3: this comparaison works ? python is too powerful

scripts/playground.ts
```diff
@@ -41,6 +41,9 @@ export async function playground({
       // eslint-disable-next-line no-console
       console.log(await run(`cd playground/go && /usr/local/go/bin/go run . --client ${client}`));
       break;
+    case 'python':
+      await run(`cd playground/python && poetry install && poetry run ${client}`);
```
millotp, line 45: maybe we should use `cwd` instead of `cd`, for the others too

templates/python/partial_header.mustache
```diff
@@ -1,3 +1,5 @@
+# coding: utf-8
+
 """
     {{{generationBanner}}}
```
millotp, line 4: I think this should not be indented

clients/algoliasearch-client-python/pyproject.toml
millotp, line 1: this file should be deleted since it's generated

millotp (commented):
beautiful cleanup ! I'll look at the generated code in the next review

clients/algoliasearch-client-python/setup.py
shortcuts: Yup that's a cute syntax

scripts/playground.ts
shortcuts: True!

clients/algoliasearch-client-python/pyproject.toml
shortcuts: good point

millotp (approved):
that's very nice

PR #2287, opened 2023-11-24
chore: next branch generation tutorial

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

I rewrote the doc to properly create the next branch, also moved the script to the repository to avoid any breaking changes on the gist side, plus added best effort cleanup script.


scripts/next-branch-generation/cleanup-repository
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+
+# Move to the root
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null && pwd)"
+cd ${DIR}/../../
+
+rm -rf .codesandbox \
```
millotp, line 7: maybe just remove everything but `clients` ?

shortcuts: I wanted to avoid deleting stuff generated by the git mv stuff since it moves it to the root level, so this way it only deletes was we know is on this repo

millotp (approved):

PR #2306, opened 2023-11-28
feat(python): add Algolia transporter, response deserializer and config setup

shortcuts:
## 🧭 What and Why

### Changes included:

This PR adds the `http` pacakage (Algolia transporter and retry strategy) to the Python client.

There is also a bit of cleanup on the template side in order to remove the redundant definition of the serialization/deserialization methods, there is still iteration to be made but at least we can better see how to use the client

I kinda liked the fact that the generator provided a `$method_with_http_info` function which allows the user to deserializes the data themselves/play with the API response, lmk if you want it to be removed


templates/python/pyproject.mustache
shortcuts, line 1: I will do a dependency cleanup once I've reached the DX improvements steps

playground/python/app/search.py
```diff
@@ -1,10 +1,23 @@
...
 from algoliasearch.search import __version__
 from algoliasearch.search.client import SearchClient
 
 
-def main():
+async def main():
     print("SearchClient version", __version__)
 
-    client = SearchClient()
+    client = SearchClient.create("QPBQ67WNIG", "b590ae1153bf574215ca1605c19eb1fe")
```
shortcuts, line 10: this is a public search key so for now it's fine to leave it, I'll remove once the client is more prodable

clients/algoliasearch-client-python/algoliasearch/http/transporter.py
shortcuts, line 1: except the serialization/deserialization steps folder is pretty much the same as the current client. I'll iterate it for oneOf handling since it doesn't fully work yet

clients/algoliasearch-client-python/algoliasearch/http/hosts.py
```diff
@@ -0,0 +1,45 @@
...
+    def __init__(
+        self, url: str, priority: Optional[int] = 0, accept: Optional[int] = None
+    ) -> None:
+        self.url = url
+        self.priority = cast(int, priority)
+        self.accept = (CallType.WRITE | CallType.READ) if accept is None else accept
+
+        self.last_use = 0.0
+        self.retry_count = 0
+        self.up = True
```
millotp, line 17: maybe call `self.reset()` here

clients/algoliasearch-client-python/algoliasearch/http/request_options.py
```diff
@@ -0,0 +1,81 @@
...
+class AllowedRequestsOptionsParams(object):
+    HEADERS = (
+        "Content-type",
+        "User-Agent",
+        "X-Algolia-User-ID",
+        "X-Algolia-UserToken",
+        "X-Forwarded-For",
+    )
+
+    QUERY_PARAMETERS = (
```
millotp, line 69: we have more qp than that, I don't think we need an allowlist

clients/algoliasearch-client-python/algoliasearch/http/verb.py
```diff
@@ -0,0 +1,5 @@
+class Verb(object):
```
millotp, line 1: we also have `PATCH` for ingestion

templates/python/__init__model.mustache
```diff
@@ -1,7 +1 @@
-{{>partial_header}}
-
-{{#models}}
-{{#model}}
-from algoliasearch.{{modelPackage}}.{{classFilename}} import {{classname}}
-{{/model}}
-{{/models}}
\ No newline at end of file
+{{>partial_header}}
```
millotp, line 1: do you really need that file ? maybe replace where it's called by `>partial_header` no ?

millotp (commented):
nice

clients/algoliasearch-client-python/algoliasearch/http/request_options.py
shortcuts: true, I just imported the actual implementation but I can do a cleanup on this part

templates/python/config.mustache
```diff
@@ -0,0 +1,72 @@
...
+from typing import Optional
+
+from algoliasearch.http.exceptions import AlgoliaException
+from algoliasearch.http.hosts import CallType, Host, HostsCollection
+from algoliasearch.http.user_agent import UserAgent
+
+
+class Config:
+    def __init__(self, app_id: Optional[str] = None, api_key: Optional[str] = None, {{#hasRegionalHost}}region: Optional[str] = None{{/hasRegionalHost}}) -> None:
+        app_id = environ.get("ALGOLIA_APP_ID") if app_id is None else app_id
```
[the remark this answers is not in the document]
shortcuts: it's implemented in the current client so I assume it has been requested at some point, I kind of don't want to break this :/

templates/python/__init__model.mustache
shortcuts: the mustache? yes I can have the same one for each package, but I can't remove the __init__

clients/algoliasearch-client-python/algoliasearch/http/request_options.py
shortcuts: I've removed the check like in other clients

millotp (approved):
let's go !

PR #2309, opened 2023-11-30
fix(python): deserializer and unused helpers

shortcuts:
## 🧭 What and Why

### Changes included:

- fixes response deserializing
- use latest pydantic methods to output JSON data
- remove defaults API values from the API client
- remove unused serialized/deserializer helpers
- fixes region fallback definition
- improve imports to prevent user finding the correct files

next step will be about CTS coverage

### DIFF
--- a/generators/src/main/java/com/algolia/codegen/AlgoliaPythonGenerator.java
+++ b/generators/src/main/java/com/algolia/codegen/AlgoliaPythonGenerator.java
@@ -6,7 +6,6 @@ import org.openapitools.codegen.*;
 import org.openapitools.codegen.CodegenConstants;
 import org.openapitools.codegen.languages.PythonClientCodegen;
 import org.openapitools.codegen.model.ModelMap;
-import org.openapitools.codegen.model.ModelsMap;
 import org.openapitools.codegen.model.OperationsMap;
 
 public class AlgoliaPythonGenerator extends PythonClientCodegen {
@@ -117,27 +116,5 @@ public class AlgoliaPythonGenerator extends PythonClientCodegen {
     operations.setImports(imports);
 
     return operations;
-  }
-
-  // we iterate over every models in order to remove the default values because we want to keep them
-  // on the API side
-  @Override
-  public Map<String, ModelsMap> postProcessAllModels(Map<String, ModelsMap> objs) {
-    Map<String, ModelsMap> models = super.postProcessAllModels(objs);
-    for (ModelsMap modelContainer : models.values()) {
-      // modelContainers always have 1 and only 1 model in our specs
-      var model = modelContainer.getModels().get(0).getModel();
-      if (model.vars.isEmpty()) continue;
-
-      for (CodegenProperty property : model.vars) {
-        if (property.vendorExtensions.isEmpty()) continue;
-
-        String typing = (String) property.vendorExtensions.get("x-py-typing");
-        if (typing == null) continue;
-
-        property.vendorExtensions.put("x-py-typing", typing.replaceAll("default=.*?, ", ""));
-      }
-    }
-    return models;
   }
 }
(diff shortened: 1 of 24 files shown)


templates/python/config.mustache
```diff
@@ -47,16 +47,12 @@ class Config:
 
         {{#hasRegionalHost}}
         _regions = [{{#allowedRegions}}'{{.}}'{{^-last}},{{/-last}}{{/allowedRegions}}]
-        _region = {{#fallbackToAliasHost}}"{{{hostWithFallback}}}" if region is None else {{/fallbackToAliasHost}} region
+        _url = {{#fallbackToAliasHost}}"{{{hostWithFallback}}}" if region is None else {{/fallbackToAliasHost}} "{{{regionalHost}}}".replace("{region}", region)
 
-        if {{^fallbackToAliasHost}}_region == "" or {{/fallbackToAliasHost}}(_region and _region not in _regions):
+        if {{^fallbackToAliasHost}}not region or {{/fallbackToAliasHost}}(region is not None and region not in _regions):
             raise AlgoliaException(f"`region` {{^fallbackToAliasHost}}is required and {{/fallbackToAliasHost}}must be one of the following: {', '.join(_regions)}")
```
millotp, line 53: I think this check should be before creating the `_url`, no need to create it if it's invalid

millotp (commented):
nice

templates/python/config.mustache
shortcuts: tru

generators/src/main/java/com/algolia/codegen/AlgoliaPythonGenerator.java
```diff
@@ -73,9 +73,10 @@ public void processOpts() {
     // repository
     supportingFiles.add(new SupportingFile("pyproject.mustache", "../", "pyproject.toml"));
     supportingFiles.add(new SupportingFile("gitignore.mustache", "../", ".gitignore"));
-    supportingFiles.add(new SupportingFile("__init__.mustache", "", "__init__.py"));
-    supportingFiles.add(new SupportingFile("__init__.mustache", packageName, "__init__.py"));
-    supportingFiles.add(new SupportingFile("__init__.mustache", packageName + "/models", "__init__.py"));
+    supportingFiles.add(new SupportingFile("__init.package__.mustache", "", "__init__.py"));
+    supportingFiles.add(new SupportingFile("__init.client__.mustache", packageName, "__init__.py"));
+    supportingFiles.add(new SupportingFile("__init.model__.mustache", packageName + "/models", "__init__.py"));
+    supportingFiles.add(new SupportingFile("__init.http__.mustache", "http", "__init__.py"));
```
shortcuts, line 79: for DX sake I had to make a specific init file per folder, it makes the usage much cleaner

scripts/formatter.ts
```diff
@@ -38,7 +38,7 @@ export async function formatter(language: string, folder: string): Promise<void>
       }
       break;
     case 'python':
-      cmd = `(cd ${folder} && poetry install --sync && pip freeze > requirements.txt && poetry run autopep8 -r --in-place --aggressive . && poetry run autoflake -r --remove-unused-variables --remove-all-unused-imports --in-place . && poetry run isort . && poetry run black . && poetry run flake8 --ignore=E501,W503 .)`;
+      cmd = `(cd ${folder} && poetry install --sync && pip freeze > requirements.txt && poetry run autopep8 -r --in-place --aggressive . && poetry run autoflake -r --ignore-init-module-imports --remove-unused-variables --remove-all-unused-imports --in-place . && poetry run isort . && poetry run black . && poetry run flake8 --ignore=E501,W503 --per-file-ignore='**/__init__.py:F401' .)`;
```
shortcuts, line 41: F401 is the error code for unused import, it's common to ignore it in init files

templates/python/imports.mustache
```diff
@@ -0,0 +1,38 @@
...
+  Union,
+)
+from typing_extensions import Literal
+
+from algoliasearch.http import (
+  ApiResponse,
+  RequestOptions,
+  Transporter,
+  Verb,
+)
```
shortcuts, line 36: we purposely import anything that we use so that we can keep a single import definition, the linter then removes the unused ones

millotp (approved):
looks good !

PR #2333, opened 2023-12-04
fix(ruby): add libraries in the Dockerfile

millotp:
## 🧭 What and Why

The ruby setup was broken, we need more dependencies, use the correct folder, and export the correct path.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,9 +16,10 @@ ENV DOCKER=true
 # use bash for subsequent commands
 SHELL ["/bin/bash", "--login", "-c"]
 
-# Global
+# Global (and librairies for ruby)
 RUN apt-get update && apt-get install -y --no-install-recommends \
     curl zip unzip git openssh-server \
+    build-essential bison zlib1g-dev libyaml-dev libssl-dev libgdbm-dev libreadline-dev libncurses5-dev libffi-dev \
     && apt-get clean \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -62,8 +63,9 @@ RUN sdk install sbt
 
 # Ruby with RVM, because it's too difficult with the image, dependencies are splattered everywhere
 ARG RUBY_VERSION
-RUN wget -O ruby.tar.gz https://github.com/postmodern/ruby-install/releases/download/v0.9.2/ruby-install-0.9.2.tar.gz \
-    && tar -xzvf ruby.tar.gz && cd ruby && make install && ruby-install ruby ${RUBY_VERSION} && gem install bundler
+RUN echo "export PATH=$PATH:/opt/rubies/ruby-${RUBY_VERSION}/bin" >> ~/.profile && source ~/.profile
+RUN curl -L -o ruby.tar.gz https://github.com/postmodern/ruby-install/releases/download/v0.9.2/ruby-install-0.9.2.tar.gz \
```
shortcuts, line 67: Do we still need the redirect?

shortcuts (approved):
Noiiice

Dockerfile
millotp: the `-L` yes, it redirects to some obsure raw.githubusercontent address

PR #2335, opened 2023-12-07
chore: add missing files to cache computing

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

we sometimes have to bump the cache key manually which means that we are missing some file changes when computing it.

there was some missing files in the common dependency and the dependency we compute the cache key from, in this PR we try to make it more programmatic to avoid having to add new files


aallam (approved):
Nice refacto! let's goo!

PR #2367, opened 2023-12-11
fix(python): path encoding and request options

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-1711

### Changes included:

- better dx for request options and prevent overriding the default parameters
- fix templates to prevent always using read transporter
- encode path parameters only when it's not a custom request
- validate parameters
- fix https://github.com/algolia/api-clients-automation/issues/539


templates/python/api.mustache
```diff
@@ -93,43 +93,46 @@ class {{classname}}:
...
-        if {{paramName}} is not None:
-            _path_params['{{baseName}}'] = {{paramName}}{{#isEnumRef}}.value{{/isEnumRef}}
-        {{/pathParams}}
+        {{/required}}
+        {{/allParams}}
+
+        _query_params: List[Tuple[str, str]] = []
+        _body: Optional[bytes] = None
+        {{#vendorExtensions}}
+        _path = '{{{path}}}'{{#pathParams}}.replace({{=<% %>=}}'{<%baseName%>}'<%={{ }}=%>, {{#x-is-custom-request}}{{paramName}}{{/x-is-custom-request}}{{^x-is-custom-request}}quote(str({{paramName}})){{/x-is-custom-request}}){{/pathParams}}
```
millotp, line 107: why does those variable starts with `_` ? is it required ?

templates/python/api.mustache
```diff
@@ -93,43 +93,46 @@ class {{classname}}:
...
         )
 
         response = await self._transporter.request(
             verb=Verb.{{httpMethod}},
-            path=_param[0],
-            data=_param[1],
-            request_options=_param[2],
+            path=_path,
+            data=_param[0],
+            request_options=_param[1],
```
millotp, line 130: a bit weird to have body and options in the same object, can't they be two separated variables ?

millotp (commented):
nice !

templates/python/api.mustache
shortcuts: it's common for variable that are used in the current scope only/internal

templates/python/api.mustache
shortcuts: ah it's because of tuples, the serializer returns two values and that's how we unpack it, I can split the concern at the serializer part but I believe there would be overlap

PR #2368, opened 2023-12-11
feat(csharp): add http transporter, serializer and configuration

morganleroi:
## 🧭 What and Why

🎟 JIRA Ticket: [API-728](https://algolia.atlassian.net/browse/APIC-728)

### Changes included:
- Import Http Transporter from current Algolia Client
- Reuse custom serialisation from Generated Client
- Inject `RequestOptions` in each routes.
- Adapt CallType
- Import Retry strategy

In next PR :
- CTS
- More cleaning
- DX


clients/algoliasearch-client-csharp/algoliasearch/Client/AlgoliaConfig.cs
```diff
@@ -0,0 +1,82 @@
...
+    /// <summary>
+    /// Algolia's default headers.
+    /// Will be sent for every request
+    /// </summary>
+    public Dictionary<string, string> DefaultHeaders { get; set; }
+
+    /// <summary>
+    /// Save methods batch size
+    /// </summary>
+    public int BatchSize { get; set; } = 1000;
```
shortcuts, line 60: we should precise this is only for the search client

clients/algoliasearch-client-csharp/algoliasearch/Client/AlgoliaConfig.cs
```diff
@@ -0,0 +1,82 @@
...
+
+    /// <summary>
+    /// Set the read timeout in seconds for all requests
+    /// </summary>
+    public int? ReadTimeout { get; set; }
+
+    /// <summary>
+    /// Set the read timeout in seconds for all requests
+    /// </summary>
+    public int? WriteTimeout { get; set; }
```
shortcuts, line 70: we've went with ms for the generated clients

clients/algoliasearch-client-csharp/algoliasearch/Client/ClientUtils.cs
```diff
@@ -38,9 +27,9 @@ public static string SanitizeFilename(string filename)
     /// <param name="name">Key name.</param>
     /// <param name="value">Value object.</param>
     /// <returns>A multimap of keys with 1..n associated values.</returns>
-    public static Multimap<string, string> ParameterToMultiMap(string collectionFormat, string name, object value)
+    public static Dictionary<string, string> ParameterToMultiMap(string collectionFormat, string name, object value)
```
shortcuts, line 30:
```suggestion
    public static Dictionary<string, string> ParameterToDictionary(string collectionFormat, string name, object value)
```

maybe?

clients/algoliasearch-client-csharp/algoliasearch/Client/ClientUtils.cs
```diff
@@ -80,29 +69,28 @@ public static string SanitizeFilename(string filename)
     /// Otherwise just return the string.
     /// </summary>
     /// <param name="obj">The parameter (header, path, query, form).</param>
-    /// <param name="configuration">An optional configuration instance, providing formatting options used in processing.</param>
     /// <returns>Formatted string.</returns>
-    public static string ParameterToString(object obj, IReadableConfiguration configuration = null)
+    public static string ParameterToString(object obj)
     {
       if (obj is DateTime dateTime)
```
shortcuts, line 75: I think we don't have any date parameters due to the complexity of the usage on the user side, it's a string with a respected format but TBC

clients/algoliasearch-client-csharp/algoliasearch/Client/Defaults.cs
```diff
@@ -0,0 +1,34 @@
...
+
+    /// <summary>
+    /// Read timeout in seconds
+    /// </summary>
+    public const int ReadTimeout = 5;
+
+    /// <summary>
+    /// Write timeout in seconds
+    /// </summary>
+    public const int WriteTimeout = 30;
```
shortcuts, line 21: Same here we use MS now

clients/algoliasearch-client-csharp/algoliasearch/Client/Defaults.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+  /// <summary>
+  /// Holding all default values of the library
+  /// </summary>
+  internal class Defaults
+  {
+    /// <summary>
+    /// Maximum time to wait for wait task in milliseconds
+    /// </summary>
+    public const int MaxTimeToWait = 10000;
```
shortcuts, line 11: I think this can be removed, since wait helpers doesn't exist in the specs, they are added afterward so it should be unused

clients/algoliasearch-client-csharp/algoliasearch/Http/AlgoliaHttpRequester.cs
```diff
@@ -0,0 +1,141 @@
...
+
+    /// <summary>
+    /// Don't use it directly
+    /// Send request to the REST API
+    /// </summary>
+    /// <param name="request">Request</param>
+    /// <param name="totalTimeout">Timeout in seconds</param>
+    /// <param name="ct">Optional cancellation token</param>
+    /// <returns></returns>
+    public async Task<AlgoliaHttpResponse> SendRequestAsync(Request request, int totalTimeout,
```
shortcuts, line 56: if we go with async only we can maybe remove the `async` keyword, up to you

clients/algoliasearch-client-csharp/algoliasearch/Http/AlgoliaHttpRequester.cs
```diff
@@ -0,0 +1,141 @@
...
+      }
+      catch (HttpRequestException e)
+      {
+        // HttpRequestException is thrown when an underlying issue happened such as
+        // network connectivity, DNS failure, server certificate validation.
+        return new AlgoliaHttpResponse { IsNetworkError = true, Error = e.Message };
+      }
+    }
+
+    private async Task<string> StreamToStringAsync(Stream stream)
```
shortcuts, line 126: is it from the current client? :o that's huge I didn't knew we had that

clients/algoliasearch-client-csharp/algoliasearch/Transport/HttpTransport.cs
```diff
@@ -0,0 +1,227 @@
...
+      if (data == null)
+        return null;
+
+      // MemoryStream ms = new MemoryStream();
+
+      // CompressionType compressionType = compress ? CompressionType.GZIP : CompressionType.NONE;
+      return _serializer.Serialize(data);
+
+      // ms.Seek(0, SeekOrigin.Begin);
+      // return ms;
```
shortcuts, line 164: is this commented code expected?

config/generation.config.mjs
```diff
@@ -103,8 +103,15 @@ export const patterns = [
...
+  '!clients/algoliasearch-client-csharp/Algolia.Search.sln',
+  'clients/algoliasearch-client-csharp/algoliasearch/Client/ClientUtils.cs',
+  '!clients/algoliasearch-client-csharp/algoliasearch/Client/**',
+  '!clients/algoliasearch-client-csharp/algoliasearch/Exceptions/**',
+  '!clients/algoliasearch-client-csharp/algoliasearch/Serializer/**',
+  '!clients/algoliasearch-client-csharp/algoliasearch/Utils/**',
+  'clients/algoliasearch-client-csharp/algoliasearch/Http/RequestOptions.cs',
+  '!clients/algoliasearch-client-csharp/algoliasearch/Http/**',
+  '!clients/algoliasearch-client-csharp/algoliasearch/Transport/**',
+  '!clients/algoliasearch-client-csharp/algoliasearch/Algolia.Search.csproj',
```
shortcuts, line 115: we might want to group all of the non generated code (http transporter) under the same package, so that it's easier to scope and to determine what's not generated

config/generation.config.mjs
```diff
@@ -103,8 +103,15 @@ export const patterns = [
...
+  '!clients/algoliasearch-client-csharp/.gitignore',
```
shortcuts, line 116: I think this is generated

templates/csharp/ClientUtils.mustache
shortcuts, line 1: I think I saw this file was pushed, no? if so, the generations folder is misconfigured

shortcuts (commented):
that's huge, gg

clients/algoliasearch-client-csharp/algoliasearch/Client/AlgoliaConfig.cs
morganleroi: Is the batch handling a requirement ? If yes then indeed I lost the batching feature on the search client.

shortcuts: IIRC we've not implemented this helper

morganleroi: In the current API Client, if user calls `SaveObject` then we automatically decide in we call the batch route or not.

shortcuts: yes we don't do that here, there's no helpers in generated clients, only methods to endpoints

morganleroi: Ok so indeed BatchSize is unused here. I will remove it.

clients/algoliasearch-client-csharp/algoliasearch/Client/AlgoliaConfig.cs
morganleroi: Are you sure ? I can see in several clients that we manage in Seconds

morganleroi: https://github.com/algolia/api-clients-automation/blob/d18321c1bbf01dfb5004990e07a2738820dc2b51/clients/algoliasearch-client-java/algoliasearch/src/main/java/com/algolia/config/ClientOptions.java#L127

morganleroi: Dart, Kotlin, Java, Php at least

clients/algoliasearch-client-csharp/algoliasearch/Client/ClientUtils.cs
morganleroi: Indeed. And I think this function will be killed in next PR.

clients/algoliasearch-client-csharp/algoliasearch/Client/ClientUtils.cs
morganleroi: Ok let's remove that !

clients/algoliasearch-client-csharp/algoliasearch/Http/AlgoliaHttpRequester.cs
morganleroi: Yes
https://github.com/algolia/algoliasearch-client-csharp/blob/20cf38a00433925713d7aaa9ffad6483f2d30c85/src/Algolia.Search/Http/AlgoliaHttpRequester.cs#L136

clients/algoliasearch-client-csharp/algoliasearch/Client/Defaults.cs
morganleroi: Do we have a list of Helpers we need to add in each API Client ?

clients/algoliasearch-client-csharp/algoliasearch/Client/AlgoliaConfig.cs
morganleroi: I'll go for ms. And you don't suffix the property by `ms` right ? Just `ReadTimeout` and the only way to know this is in ms it's with the comment. Right?

millotp: yes we did, we use ms everywhere, unless there is a built in Duration type, like Go or Java

millotp: I don't see any client using second, only ms or Duration

morganleroi: Ok then I'll expose a Duration. C# as a `Duration` too.

millotp: I think having `Duration.ofSeconds(2)`, it's explicit, either the unit is in the name, or we default to ms

clients/algoliasearch-client-csharp/algoliasearch/Http/AlgoliaHttpRequester.cs
morganleroi: the suffix is pretty common in C# lib but indeed I'll think about it.

config/generation.config.mjs
morganleroi: Indeed. But so far I don't have a clear idea about the folders hierarchy. If possible, I'd like to rework that later.

clients/algoliasearch-client-csharp/algoliasearch/Client/Defaults.cs
morganleroi: Removed.

shortcuts: good idea we should write one, basically anything here https://github.com/algolia/api-clients-automation/blob/main/templates/javascript/clients/client/api/helpers.mustache

config/generation.config.mjs
shortcuts: usually we group anything related to the requester in a single folder, the name is up to you. Then if you come to add other requester they have their own other folder (e.g. js has 3 requester)

clients/algoliasearch-client-csharp/algoliasearch/Client/AlgoliaConfig.cs
```diff
@@ -0,0 +1,78 @@
...
+    /// </summary>
+    public Dictionary<string, string> DefaultHeaders { get; set; }
+
+    /// <summary>
+    /// Set the read timeout for all requests
+    /// </summary>
+    public TimeSpan? ReadTimeout { get; set; }
+
+    /// <summary>
+    /// Set the read timeout in seconds for all requests
```
shortcuts, line 64:
```suggestion
    /// Set the read timeout for all requests
```

clients/algoliasearch-client-csharp/algoliasearch/Http/AlgoliaHttpRequester.cs
```diff
@@ -0,0 +1,141 @@
...
+        {
+          InnerHandler = new HttpClientHandler { AutomaticDecompression = DecompressionMethods.GZip }
+        });
+
+    /// <summary>
+    /// Don't use it directly
+    /// Send request to the REST API
+    /// </summary>
+    /// <param name="request">Request</param>
+    /// <param name="totalTimeout">Timeout in seconds</param>
```
shortcuts, line 53:
```suggestion
    /// <param name="totalTimeout">Timeout</param>
```

clients/algoliasearch-client-csharp/algoliasearch/Transport/HttpTransport.cs
morganleroi: Nope. Remove until I work on compression.

config/generation.config.mjs
morganleroi: Indeed. Fixed.

templates/csharp/ClientUtils.mustache
morganleroi: Should be fixed now

clients/algoliasearch-client-csharp/algoliasearch/Client/AlgoliaConfig.cs
morganleroi: Grrrr nice catch.

clients/algoliasearch-client-csharp/algoliasearch/Http/AlgoliaHttpRequester.cs
morganleroi: Grrr again. Nice catch, again.

templates/csharp/libraries/httpclient/RequestOptions.mustache
```diff
@@ -22,45 +20,37 @@ namespace Algolia.Search.{{packageName}}.Client
         /// Query parameters to be applied to the request.
         /// Keys may have 1 or more values associated.
         /// </summary>
-        public Multimap<string, string> QueryParameters { get; set; }
+        public Dictionary<string, string> QueryParameters { get; set; }
```
shortcuts, line 23: You can have other types than string for qp values no?

morganleroi: This comes from the generated code. I have no idea if this will works :-D Can I wait CTS to validate that ? Or CTS will not show me errors for that ?

shortcuts: I think the CTS will but if you serialize the value befor eit should be fine then

morganleroi: Same bucket as the current APi client https://github.com/algolia/algoliasearch-client-csharp/blob/5df63dec4ecc96dad25effee209de19660b85314/src/Algolia.Search/Http/RequestOptions.cs#L41

shortcuts (approved):
GGGG

PR #2369, opened 2023-12-11
fix(specs): provide non clashing names for custom methods

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

The current name of the custom methods can clash with reserved keywords of certain languages (e.g. `del` for `python`). This PR proposes a `custom` prefix to those method names.

since Dart is already released, we provide the old methods as `helpers` with a `deprecated` warning. They still leverage the custom methods under the hood.


Fluf22 (approved):
Nice 🆗

templates/dart/api.mustache
```diff
@@ -157,7 +157,45 @@ final class {{classname}} implements ApiClient {
   }  
   {{/operation}}
 
+  {{^isAlgoliasearchClient}}
```
millotp, line 160: I think this should be for all clients, not just != algoliasearch

millotp (commented):
looks great ! small comment about dart, I though we included the custom requests for algoliasearch too

templates/dart/api.mustache
shortcuts: I don't think so for dart but TBC!

shortcuts: confirmed they are not in the algoliasearch client

millotp: okay but why is it removed and not replaced in the searchClient ? https://github.com/algolia/api-clients-automation/commit/d0a0c96d195bb632c61861f46bbaa01eeda04503#diff-6e097578fa886b0b109622bbfadeda83cc72b0c60fba70d7d8a6c429b64358faL55

shortcuts: done here https://github.com/algolia/api-clients-automation/pull/2369/commits/7c3fbf6495dc479a10fac35508760dced9742a3d

PR #2372, opened 2023-12-12
feat(python): add release process

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-1713

### Changes included:

Add release process for the Python client using https://github.com/pypa/gh-action-pypi-publish

I've registered a trusted publisher with my account on PyPi, but I need to try if there's any additional changes to do


clients/algoliasearch-client-python/.github/workflows/release.yml
```diff
@@ -0,0 +1,36 @@
...
+    if: "startsWith(github.event.head_commit.message, 'chore: release')"
+    environment:
+      name: pypi
+      url: https://pypi.org/p/algoliasearch
+    permissions:
+      id-token: write
+    steps:
+    - name: Install poetry
+      shell: bash
+      run: pipx install poetry
```
millotp, line 21: don't you need to install peotry after python ? you don't know the version of pipx at this point

shortcuts: ah yes good catch!

millotp (approved):

PR #2374, opened 2023-12-13
docs: return code for objects endpoint

kai687:
## 🧭 What and Why

`POST /1/indexes/{indexName}` returns a `201 Created` on success, not a 200.

🎟 JIRA Ticket: n/a

### Changes included:

- Change return code and description for the success response for this operation.


shortcuts (approved):
Thanks!!

PR #2376, opened 2023-12-13
feat(specs): update Insights API spec

kai687:
## 🧭 What and Why

Update the API spec for the Insights API we wanted to make an update to 'Purchase' events.

We want to remove the event-level `queryID` parameter in the documentation for 'Purchase' events, and rather have object-level `queryID`s. (Otherwise, it's too confusing).

🎟 JIRA Ticket: n/a

### Changes included:

- Add the 'value' parameter to the 'AddToCart' and 'Purchase' events. 
- Remove the event level `queryID` parameter from the 'Purchase' events.
- Some other minor edits (mostly sorting and formatting, making things more consistent).

## 🧪 Test

n/a


shortcuts:
hey @kai687, thanks for the PR, taking a look!

shortcuts:
the last issue remaining is that the echo interceptor for java requires the method to return the request, but since there's no payload we can't do that, will try to find a fix for that (cc @millotp)

morganleroi:
working on a C# fix.

millotp (approved):
finally green ! gg everyone

PR #2379, opened 2023-12-15
fix(ruby): make the playground work

millotp:
## 🧭 What and Why

Add `app_id`, `api_key` and `region` to the client, make the hosts work (no retry strategy for now), and fix the playground.


shortcuts (approved):
wonderful, and 90k deletion is HUGE

PR #2381, opened 2023-12-18
chore: renovate for Python

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

just realized it did not worked because that's not the correct filename, also remove `poetry` from the automated managers because we only want to update the templates

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -14,7 +14,6 @@
     "regex",
     "pub",
     "gomod",
-    "poetry",
     "pip_requirements",
     "gradle-wrapper",
     "bundler"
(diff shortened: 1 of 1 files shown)


millotp (approved):
gg, don't forget that you can test locally with https://github.com/algolia/data-ingestion/blob/main/docs/renovate.md 😉

shortcuts:
> gg, don't forget that you can test locally with https://github.com/algolia/data-ingestion/blob/main/docs/renovate.md 😉

yup I was too lazy to setup the whole stuff D:

PR #2383, opened 2023-12-18
feat(ruby): retry strategy

millotp:
## 🧭 What and Why

🎟 JIRA Ticket: [DI-1718](https://algolia.atlassian.net/browse/DI-1718)

Bring in the retry strategy from the existing client, and try to make it work with the generated one.
The error handling is not perfect and will be improved in future PR.

[DI-1718]: https://algolia.atlassian.net/browse/DI-1718?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


templates/ruby/api.mustache
```diff
@@ -225,17 +222,14 @@ module {{moduleName}}
...
       new_options = opts.merge(
         :operation => :"{{classname}}.{{operationId}}",
-        :header_params => header_params,
-        :query_params => query_params,
+        :headers => header_params,
+        :params => query_params,
         :body => post_body,
-        :return_type => return_type
+        :return_type => return_type,
+        :use_read_transporter => {{#useReadTransporter}}true{{/useReadTransporter}}{{^useReadTransporter}}false{{/useReadTransporter}}
```
shortcuts, line 229: isn't the value of `useReadTransporter` true or false? we maybe can use that

not blocking just wondering

templates/ruby/api.mustache
```diff
@@ -225,17 +222,14 @@ module {{moduleName}}
 
       new_options = opts.merge(
         :operation => :"{{classname}}.{{operationId}}",
-        :header_params => header_params,
-        :query_params => query_params,
+        :headers => header_params,
+        :params => query_params,
```
shortcuts, line 226: I think this should stay query_parameters or query_params, not sure if we did that elsewhere but it's more descriptive

clients/algoliasearch-client-ruby/lib/algolia/user_agent.rb
shortcuts, line 1: I see some files out of the `http` folder, wdyt of having all of them there for easier scoping of what's not generated?

clients/algoliasearch-client-ruby/lib/algolia/transport/transport.rb
```diff
@@ -0,0 +1,138 @@
...
+      #
+      def request(call_type, method, path, body, opts = {})
+        @retry_strategy.get_tryable_hosts(call_type).each do |host|
+          opts[:timeout]         ||= get_timeout(call_type) * (host.retry_count + 1)
+          opts[:connect_timeout] ||= @config.connect_timeout * (host.retry_count + 1)
+
+          request_options = RequestOptions.new(@config)
+          request_options.create(opts)
+          # TODO: what is this merge for ?
+          # request_options.params.merge!(request_options.data) if method == :GET
```
shortcuts, line 36: we proxy the POST payload in QP if it's a get request IIRC

shortcuts: or the other way around but I think there's something like that

clients/algoliasearch-client-ruby/lib/algolia/transport/retry_strategy.rb
```diff
@@ -0,0 +1,116 @@
...
+        end
+
+        # Make a binary check to know whether the item contains the flag
+        #
+        # @param item [binary] item to check
+        # @param flag [binary] flag to find in the item
+        #
+        # @return [Boolean]
+        #
+        def flag?(item, flag)
```
shortcuts, line 111: do we need that?

clients/algoliasearch-client-ruby/lib/algolia/http/response.rb
```diff
@@ -0,0 +1,23 @@
+module Algolia
+  module Http
+    class Response
+      attr_reader :status, :body, :error, :headers, :has_timed_out, :network_failure
+
+      #
+      # @option status    [String]  Response status
+      # @option body    [String]  Response body
+      # @option error    [String]  Response error or caught error
+      # @option headers    [String]  Response headers
```
shortcuts, line 10: shouldn't it be an object? or we let the user deserialize it?

clients/algoliasearch-client-ruby/lib/algolia/configuration.rb
```diff
@@ -73,24 +22,35 @@ class Configuration
...
-      @configure_connection_blocks = []
-      @timeout = 60
       @params_encoder = nil
-      @debugging = false
       @inject_format = false
       @force_ending_format = false
-      @logger = defined?(Rails) ? Rails.logger : Logger.new(STDOUT)
+      @read_timeout = 30
+      @write_timeout = 5
+      @connect_timeout = 2
```
shortcuts, line 43: maybe go with ms or duration like other clients?

shortcuts (commented):
huge!!

templates/ruby/api.mustache
millotp: yep I agree, this was to avoid more breaking changes but I can do it

templates/ruby/api.mustache
millotp: the variable is empty when false, I need this logic

clients/algoliasearch-client-ruby/lib/algolia/user_agent.rb
millotp: I did it to match the [old client](https://github.com/algolia/algoliasearch-client-ruby), I think what's inside `http` is what can be exchanged for a echo requester for example

clients/algoliasearch-client-ruby/lib/algolia/transport/retry_strategy.rb
millotp: yes it's used to determine which host to pick

clients/algoliasearch-client-ruby/lib/algolia/http/response.rb
millotp: so the `requester` sends the response as a string, and the `api_client.rb` deserialize it to the correct model before sending it to the user

clients/algoliasearch-client-ruby/lib/algolia/user_agent.rb
shortcuts: not a blocker anyway but I feel like it's better if we scope everything related to the transporter in the same folder

shortcuts (approved):
clean!

PR #2399, opened 2023-12-18
chore: correctly delete python on pregen

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

the snake cased folders were not deleted, resulting in the query_suggestions client being outdated

### DIFF
--- a/scripts/common.ts
+++ b/scripts/common.ts
@@ -261,6 +261,13 @@ export function capitalize(str: string): string {
   return str.charAt(0).toUpperCase() + str.slice(1);
 }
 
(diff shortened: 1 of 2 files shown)


millotp (approved):
good catch, I have the same issue for ruby actually

shortcuts:
> good catch, I have the same issue for ruby actually

I guess you can leverage the same stuff!

millotp:
yep I'll do it in my open PR

shortcuts:
removing automerge just to check if the CI deletes the code correctly

PR #2400, opened 2023-12-18
feat(ruby): make hit generic

millotp:
## 🧭 What and Why

There is no generic in ruby, so 1 way to make the Hit object accept anything is to add a any getter and setter when serializing/deserializing.


templates/ruby/base_object.mustache
```diff
@@ -21,6 +21,10 @@
           transformed_hash["#{key}"] = _deserialize(type, attributes[attribute_map[key]])
         end
       end
+      {{#vendorExtensions.x-is-generic}}
+
+      transformed_hash.merge!(attributes.reject { |k, _| attribute_map.key?(k.to_sym) })
```
shortcuts, line 26: that's an horrible syntax ahah but... if it works

shortcuts (commented):
looks great but is it expected that it only has effect on the hit model?

millotp:
there should be `getObjectsResponse` too, but maybe it doesn't work because it's an inline model

templates/ruby/base_object.mustache
millotp: Thankfully copilot wrote it for me !

shortcuts:
> there should be `getObjectsResponse` too, but maybe it doesn't work because it's an inline model

is it something we want to fix in this PR?

millotp:
The type is already `Array<Object>` so there is no issue with serialization/deserialization here, it will work as expected

shortcuts (approved):
gg!

PR #2407, opened 2023-12-19
feat(scripts): support Alpha releases

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-1713

### Changes included:

python alpha releases have a different pattern than what semver supports, this adds support for it plus a few tests


Fluf22 (approved):
Nice 👍🏻

PR #2409, opened 2023-12-19
fix(ruby): customize linter

millotp:
## 🧭 What and Why

Use a different set of linter rules to better match our repository and the generated code.


shortcuts (approved):
ternaries everywherreeeeee

looks clean

PR #2410, opened 2023-12-19
feat(generators): differentiate requests/clients test files

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

test file names should differ for Python otherwise the runner won't start the tests due to ambiguity, I only applied this to python but wdyt of adding such suffix to every files just for simplicity?

### DIFF
--- a/generators/src/main/java/com/algolia/codegen/cts/tests/TestsClient.java
+++ b/generators/src/main/java/com/algolia/codegen/cts/tests/TestsClient.java
@@ -35,6 +35,11 @@ public class TestsClient extends TestsGenerator {
     if (!available()) {
       return;
     }
+
+    if (this.language.equals("python")) {
+      extension = "_client" + extension;
+    }
+
     supportingFiles.add(
       new SupportingFile("client/suite.mustache", outputFolder + "/client", Helpers.createClientName(client, language) + extension)
     );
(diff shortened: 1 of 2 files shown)


Fluf22 (approved):
It's a good idea, as you wish!
Overall, it's just that you'll have to handle language casing differences 😄 
Maybe just do it in another PR so it doesn't block you from doing other things in Python?

millotp:
having them in different folders isn't enough ?

shortcuts:
> It's a good idea, as you wish! Overall, it's just that you'll have to handle language casing differences 😄 

yup we already have an helper for that, it can just be adding the suffix and let the helper convert

> having them in different folders isn't enough ?

nope it doesn't look like so, even if you have different test classes

millotp:
Also some languages have a strict file pattern like `go`, where the test should end in `_test.go` otherwise it won't run them

shortcuts:
> Also some languages have a strict file pattern like `go`, where the test must end in `_test.go` otherwise it won't run them

I guess if we do like the python client, e.g. `Requests|Client + extension` it should work everywhere

shortcuts:
oki we can abort, definiting each folder as a separate module works, I can just add a template file for that

PR #2412, opened 2023-12-20
chore(javascript): remove some non needed props from generators

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

we can use lambdas instead of generating a property for it


millotp (approved):
I liked the `capitalizedApiName`, but why not !

shortcuts:
> I liked the `capitalizedApiName`, but why not !

Yup in that case it moves the responsibility to the template but I kind of want to remove things to maintain in the generators D:

PR #2413, opened 2023-12-20
fix(generators): add generic to top level

morganleroi:
## 🧭 What and Why

🎟 JIRA Ticket:

### Changes included:

- List changes

### DIFF
--- a/generators/src/main/java/com/algolia/codegen/utils/GenericPropagator.java
+++ b/generators/src/main/java/com/algolia/codegen/utils/GenericPropagator.java
@@ -67,10 +67,6 @@ public class GenericPropagator {
 
   private static boolean markPropagatedGeneric(IJsonSchemaValidationProperties model) {
     CodegenProperty items = model.getItems();
-    // Skip one-of types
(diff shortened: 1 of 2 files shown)


morganleroi:
If you have time guys, could be cool to have help here to fix and check this change on Java and potentially other lang @millotp @shortcuts @damcou @aallam @Fluf22

shortcuts:
JS is good

shortcuts:
I think java too, @aallam or @millotp feel free to take a look at https://github.com/algolia/api-clients-automation/pull/2413/commits/de5acaf25d83b64c6b05cc114cd73ebaf0e18db5

shortcuts:
ok now it's scala that fails

millotp (approved):
this is looking good !

morganleroi:
Do you think, after this change, if we need to double-check manually that we do not break anything ?

shortcuts:
> Do you think, after this change, if we need to double-check manually that we do not break anything ?

only for java, js has e2e

morganleroi:
You guys are 👍🏼

PR #2401, opened 2023-12-18
feat(ruby): add CTS

millotp:
## 🧭 What and Why

🎟 JIRA Ticket: [DI-1728](https://algolia.atlassian.net/browse/DI-1728)

Generate the CTS for ruby, using the standard unit test.

[DI-1728]: https://algolia.atlassian.net/browse/DI-1728?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2414, opened 2023-12-20
chore(cts): add index to default test name

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

for simplicity, we can append the test index to the operationID when no test name are provided, this avoid tests having the same name and/or forcing the contributor to provide a testName (description)

discovered in https://github.com/algolia/api-clients-automation/pull/2411 and https://github.com/algolia/api-clients-automation/pull/2401


tests/output/javascript/yarn.lock
shortcuts, line 1: weird that it bumps it

millotp (approved):
safe change !

PR #2415, opened 2023-12-20
chore(cli): add apic alias to bypass yarn scripts

millotp:
## 🧭 What and Why

Add a bash alias for `yarn docker` named `apic`, you have to install it manually because we can't run bash built in like `source` in yarn script https://github.com/yarnpkg/berry/issues/2088.

The time goes from ~5s to launch the CLI to instantaneous !

Note: this runs the js directly, if you change the cli code you need to recompile with `tsc`, with I added to `yarn install` directly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "docker:setup": "yarn docker:clean && yarn docker:build && yarn docker:mount",
     "fix:json": "eslint --ext=json . --fix",
     "github-actions:lint": "eslint --ext=yml .github/",
-    "postinstall": "husky install && yarn workspace eslint-plugin-automation-custom build",
+    "postinstall": "husky install && yarn workspace eslint-plugin-automation-custom build && yarn workspace scripts tsc",
     "playground:browser": "yarn workspace javascript-browser-playground start",
     "release": "yarn workspace scripts createReleasePR",
     "scripts:lint": "yarn workspace scripts lint",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -20,7 +20,7 @@
     "docker:setup": "yarn docker:clean && yarn docker:build && yarn docker:mount",
     "fix:json": "eslint --ext=json . --fix",
     "github-actions:lint": "eslint --ext=yml .github/",
-    "postinstall": "husky install && yarn workspace eslint-plugin-automation-custom build",
+    "postinstall": "husky install && yarn workspace eslint-plugin-automation-custom build && yarn workspace scripts tsc",
```
shortcuts, line 23: love it!! can we add it at the end of `yarn docker:setup` too?

shortcuts (commented):
such nice addition

package.json
millotp: sure, I think we run `yarn install` more often that `yarn docker:setup` though

millotp: but it's best to have it in multiple places you're right

shortcuts (approved):
wonderful!!

scripts/package.json
```diff
@@ -3,18 +3,18 @@
   "version": "1.0.0",
   "type": "module",
   "scripts": {
```
shortcuts, line 5: alternatively we can have a yarn script that checks if a dist folder exist otherwise run tsc, but I guess this is safe to assume it will always be built

shortcuts (approved):
huge!!

scripts/package.json
millotp: yep we just have to be careful when editing the cli

PR #2417, opened 2023-12-21
feat(cts): add e2e

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

The goal of this PR is to provide support for e2e tests in our CTS generators. e2e aims at increasing our confidence level when shipping the generated API clients.

This first iteration of this PoC only focuses on two part of the assertion:
- the raw json response (`body`)
- the status code (`statusCode`)

Ideally, we should, in a later iteration, focus on the deserialization part, but thanks to how our CTS is designed, we can do that incrementally.

~_This PoC is only demonstrated with Python, but shouldn't be too different with others_~
This PoC is demonstrated with Python, JavaScript and PHP


generators/src/main/java/com/algolia/codegen/cts/tests/TestsRequest.java
```diff
@@ -107,6 +107,21 @@ public void run(Map<String, CodegenModel> models, Map<String, CodegenOperation>
             test.put("assertNullBody", true);
           }
 
+          if (req.response != null) {
+            bundle.put("hasE2E", true);
+            test.put("response", req.response);
+            test.put("assertStatusCode", language.equals("ruby") || language.equals("python"));
```
millotp, line 113: I don't understand why this is needed, each template can choose to check the statusCode no ?

tests/output/python/.env.example
millotp, line 1: I think the .env should be global, and live in `tests/output` maybe

templates/javascript/tests/requests/requests.mustache
```diff
@@ -8,6 +11,18 @@ const apiKey = process.env.ALGOLIA_SEARCH_KEY || 'test_api_key';
 
 const client = {{client}}(appId, apiKey, {{#hasRegionalHost}}'{{{defaultRegion}}}', {{/hasRegionalHost}}{ requester: echoRequester() });
 
+{{#hasE2E}}
```
millotp, line 14: always reading the credentials from the .env would make it much easier in every template

tests/CTS/methods/responses/.gitkeep
millotp, line 1: so long my friend !

millotp (commented):
insane

tests/output/python/.env.example
shortcuts: make sense, it's the same everywhere, I'll update it

generators/src/main/java/com/algolia/codegen/cts/tests/TestsRequest.java
shortcuts: it's not needed anymore actually, we can do #statusCode indeed!

templates/javascript/tests/requests/requests.mustache
shortcuts: i'm fine with that!

generators/src/main/java/com/algolia/codegen/cts/tests/TestsRequest.java
```diff
@@ -107,6 +107,20 @@ public void run(Map<String, CodegenModel> models, Map<String, CodegenOperation>
             test.put("assertNullBody", true);
           }
 
+          if (req.response != null) {
+            bundle.put("hasE2E", true);
+            test.put("response", req.response);
```
millotp, line 112: you don't need to call `enhanceParameters` on response ?

millotp (approved):
gg !

generators/src/main/java/com/algolia/codegen/cts/tests/TestsRequest.java
shortcuts: only JSON is asserted right now so not yet but indeed for JVM languages it will be required I believe, next step!

PR #2418, opened 2023-12-21
fix(ruby): support additionalProperties: true

millotp:
## 🧭 What and Why

Some models can have arbitrary keys, my previous solution only worked for `x-is-generic`, but this works for everything.


shortcuts (approved):
gg!

PR #2438, opened 2023-12-26
chore(cts): remove `methods` folder

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-1735

### Changes included:

since we have the e2e at the `requests` level we can remove all of the `methods` folders and just keep the `requests` one


shortcuts:
since it changes the path of the CTS folder, the CI doesn't delete the right one, so I had to remove it manually

tests/output/scala/src/test/scala/algoliasearch/methods/requests/QuerySuggestionsTest.scala
millotp, line 1: moving the files by hand is not enough ?

millotp (commented):
nice

tests/output/scala/src/test/scala/algoliasearch/methods/requests/QuerySuggestionsTest.scala
shortcuts: I can do that too but the outcome would be pretty similar, maybe it would keep the history? not sure

millotp (approved):

PR #2441, opened 2023-12-27
chore(cts): partial assert on e2e

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-1736

### Changes included:

Because some values of the API can vary, we can change the e2e tests a bit to only assert the fields we want. e.g. for the `search` method, `processingTimeMS` can vary, so we remove it from the expected body of the CTS and only assert the remaining fields.

I did not found native ways to do it in PHP, JavaScript and Python, so I wrote a small helper


templates/php/tests/requests/requests.mustache
```diff
@@ -36,6 +36,33 @@ class {{clientPrefix}}Test extends TestCase implements HttpClientInterface
      */
     private $recordedRequests = [];
 
+    {{#hasE2E}}
+    protected function union($expected, $received)
```
shortcuts, line 40: idk how namespaces etc. works so it's inlined for now, there's a loooot of duplicated stuff in the PHP test (e.g. `assertRequests`) etc. would be great it we move all of that to an helper file cc @damcou

Fluf22:
🆗 
It would mean one more manual helper to do in each language?
We definitely need a list of stuff required to do when adding a new language

Fluf22 (dismissed):

shortcuts:
> 🆗 It would mean one more manual helper to do in each language?

In the tests yes only if there's no native way to do so, however since e2e will be required at some point, we'll know right away but it's a one time write

> We definitely need a list of stuff required to do when adding a new language

I'm currently writing a small doc that lists helpers, where they live and what they do

shortcuts:
great example of the necessity of this addition: https://github.com/algolia/api-clients-automation/actions/runs/7337323488/job/19978111243?pr=2442

.github/workflows/check.yml
```diff
@@ -261,7 +261,7 @@ jobs:
         run: cd clients/algoliasearch-client-javascript && yarn workspace algoliasearch test
 
       - name: Remove previous CTS output
-        run: rm -rf tests/output/javascript/src || true
+        run: rm -rf tests/output/javascript/src/requests tests/output/javascript/src/client || true
```
millotp, line 264: maybe we could use `fromJSON(needs.setup.outputs.JAVASCRIPT_DATA).testsToDelete`, this was just a simplification

templates/python/tests/requests/requests.mustache
```diff
@@ -4,6 +4,7 @@ from algoliasearch.http.transporter import EchoTransporter
 from algoliasearch.{{{import}}}.client import {{#lambda.pascalcase}}{{{client}}}{{/lambda.pascalcase}}
 from algoliasearch.{{{import}}}.config import Config
 {{#hasE2E}}
+from ..helpers import Helpers
```
millotp, line 7: never saw that syntax before !

millotp (commented):
this is very nice !

templates/python/tests/requests/requests.mustache
shortcuts: python is full of different syntax.... that's confusing ahah

.github/workflows/check.yml
shortcuts: ah yes I forgot about that good idea

millotp: hope it works without the `|| true`

millotp (approved):

PR #2442, opened 2023-12-27
fix(python): less breaking changes

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

A bit of DX improvements by removing some unnecessary breaking changes, which will hopefully help user migrate


tests/output/python/requirements.txt
```diff
@@ -1,6 +1,6 @@
 argcomplete==3.1.6
 click==8.1.7
 packaging==23.2
-pipx==1.3.3
-platformdirs==4.1.0
+pipx==1.3.2
+platformdirs==4.0.0
```
millotp, line 5: it doesn't work with the latest version ?

millotp (commented):
nice cleanup !

tests/output/python/requirements.txt
shortcuts: hm great catch idk why it pushed this, I believe I ran before main was updated with the deps upgrade?

shortcuts:
will keep it flaky until https://github.com/algolia/api-clients-automation/pull/2441 is merged

millotp (approved):
gg !

PR #2444, opened 2023-12-27
chore(cli): bash autocomplete for apic

millotp:
## 🧭 What and Why

Bringing our cli to the next level with bash autocomplete ! We could event remove `inquirer` now.

Screenshot 2023-12-27 at 16 42 18


scripts/install.sh
```diff
@@ -4,7 +4,92 @@
...
 # [[ -s "$HOME/<path to api-clients-automation>/scripts/install.sh" ]] && source "$HOME/<path to api-clients-automation>/scripts/install.sh"
 
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null && pwd)"
+
 export apic() {
   docker exec -it api-clients-automation bash -lc "cd scripts && NODE_NO_WARNINGS=1 node dist/scripts/cli/index.js $*";
 }
 
+_list_languages() {
+  cat $ROOT/config/clients.config.json | jq -r 'keys[]'
```
shortcuts, line 14: Maybe it should be executed in docker so users that don't have jq (idk if that exists) doesn't have to install it

shortcuts (commented):
Thats so nice!!!!!!

scripts/install.sh
millotp: I'm not sure if it's possible, autocomplete has to execute really fast to feel natural, and this is called everytime someone presses tab.
Maybe I can make it conditional, is jq is not installed I fallback to docker ?

shortcuts: That make sense, lets not bother then, jq is required anyway elsewhere

shortcuts (approved):
Ggg

PR #2445, opened 2023-12-28
feat(ruby): client tests

millotp:
## 🧭 What and Why

Clients tests for ruby, with a few fixes along the way


Fluf22 (approved):
Nice work!

PR #2447, opened 2023-12-28
fix(ruby): support ruby alpha format

millotp:
## 🧭 What and Why

Syntax like `3.0.0.alpha.0` was not supported by `getNextVersion`.


scripts/release/createReleasePR.ts
```diff
@@ -179,7 +179,7 @@ export function getNextVersion(current: string, releaseType: semver.ReleaseType
   // python alpha releases have a pattern like X.Y.ZaN
   // where a means alpha and N can be any digit representing the alpha version
   // see https://peps.python.org/pep-0440/
```
Fluf22, line 181: Isn't this comment not related anymore?

Fluf22 (approved):

scripts/release/createReleasePR.ts
shortcuts: It can be expanded to ruby indeed

millotp: it's still relevant for python, but also support ruby now, I can add a small com

shortcuts (commented):
Sheeesh

scripts/release/createReleasePR.ts
Fluf22: Or maybe remove it? I feel like the regex is self explanatory 😁 
It's true we could document every language specificity to know why the regex get more complex each time.

millotp: hopefully we won't stumble across new version format ! this should support everything

millotp: quoi que, ruby has `[2.0.0.pre.alpha.2](https://rubygems.org/gems/algolia/versions/2.0.0.pre.alpha.2)`

Fluf22 (approved):
Let's go

shortcuts (approved):
lovely

PR #2456, opened 2023-12-29
chore(python): code cleanup

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

some regular code cleanup for the API clients

- [x] only generate `__properties` when used in the model
- [x] default `data` to `raw_data` to prevent assign at the method level
- [x] only generate `_query_parameters` and `_headers` when there's use for in the method, leverage default value otherwise


.github/workflows/check.yml
```diff
@@ -367,7 +367,7 @@ jobs:
         run: yarn cli cts run ${{ matrix.client.language }}
 
       - name: Zip artifact before storing
-        run: zip -r -y clients-${{ matrix.client.language }}.zip ${{ matrix.client.path }} ${{ matrix.client.testsToStore }} -x "**/node_modules**" "**/.yarn/cache/**" "**/build/**" "**/dist/**" "**/.gradle/**" "**/bin/**" "**/vendor/**" "**/target/**" "**/.dart_tool/**"
+        run: zip -r -y clients-${{ matrix.client.language }}.zip ${{ matrix.client.path }} ${{ matrix.client.testsToStore }} -x "**/node_modules**" "**/__pycache__" "**/.yarn/cache/**" "**/build/**" "**/dist/**" "**/.gradle/**" "**/bin/**" "**/vendor/**" "**/target/**" "**/.dart_tool/**"
```
millotp, line 370: this didn't work, I think you need `**/__pycache__/**`

shortcuts: ah yes maybe good idea

shortcuts: it worked!

millotp (approved):

PR #2458, opened 2023-12-29
chore(ci): prevent unwanted ci runs

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

changes like https://github.com/algolia/api-clients-automation/pull/2456 that only touch python file forces the CI to run every languages, I assume this is because the templates folder is considered as common, but since it's scoped to a language we can move it to the lang deps


millotp (dismissed):
this is a revert of 4b82a06f8ede2c7df6f098865de854204b5505c0 but why not

shortcuts:
> 4b82a06

partially only yes

millotp (approved):

PR #2475, opened 2024-01-04
chore(ci): compute shouldRun after filtering

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

see ci failure on main: https://github.com/algolia/api-clients-automation/actions/runs/7400572570/attempts/1

the ci thinks it should run `client_gen` because `shouldRun` is `true` since it is computed computed before we filter javascript from the matrix to put it in its own


millotp (approved):
good catch

PR #2478, opened 2024-01-05
fix(python): provide community requested changes

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

It's pretty counter intuitive to have to call a method to instantiate the class when we only want to provide the `appID` and `apiKey`, this has been reported a few times on the Python repo but since it's a breaking change, it couldn't be applied


millotp (approved):
I was gonna say you could remove the `raise ValueError` but you already did it 😁

morganleroi (approved):
Make sense !

PR #2481, opened 2024-01-05
chore(scripts): remove openapitools.json

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

most of the logic now lives in the generator or the clients.config.json, we can remove the `openapitools.json` config file in order to only keep the on the fly one


config/clients.config.json
shortcuts, line 1: maybe we can have this file as a JS one to have variables for less redundancy but that's enough for now

millotp: it would make it hard to parse in java and bash

generators/src/main/java/com/algolia/codegen/utils/Helpers.java
```diff
@@ -179,13 +179,13 @@ public static String getClientConfigField(String language, String... fields) thr
...
 
-  /** Get the `field` value in the `config/openapitools.json` file for the given language */
+  /** Get the `field` value in the `openapitools.json` file for the given language */
   public static String getOpenApiToolsField(String language, String client, String... fields) throws ConfigException {
     if (fields.length == 0) {
       throw new ConfigException("getOpenApiToolsField requires at least one field");
     }
     if (cacheOpenApiToolsConfig == null) {
-      cacheOpenApiToolsConfig = readJsonFile("config/openapitools.json");
+      cacheOpenApiToolsConfig = readJsonFile("openapitools.json");
```
millotp, line 188: this is dangerous, because now it only contains the information for the current run, not everything, like thomas want

config/clients.config.json
shortcuts: good point

scripts/common.ts
```diff
@@ -35,23 +35,35 @@ const ROOT_DIR = path.resolve(process.cwd(), '..');
...
+        key = `${language}-${client}`;
+      }
+
       // eslint-disable-next-line no-param-reassign
-      current[key].additionalProperties.packageName = output.substring(output.lastIndexOf('/') + 1);
+      current[key] = {
+        additionalProperties: {},
+        ...gen,
+        output,
+        client: key.slice(key.indexOf('-') + 1),
```
millotp, line 56: I dont' think you need the slice now, the key is already the client name

scripts/install.sh
```diff
@@ -22,12 +22,12 @@ _list_clients_for_language() {
...
     _list_clients
   else
-    echo "all $(cat $ROOT/config/openapitools.json | jq --arg lang "$1-" '."generator-cli".generators | keys[] | select(startswith($lang)) | sub($lang; "")')"
+    echo "all $(cat $ROOT/config/clients.config.json | jq --arg lang "$1" 'with_entries(select(.key == $lang)) | .[] | .clients')"
   fi
 }
 
 _list_clients() {
-  echo "all $(cat $ROOT/config/openapitools.json | jq '."generator-cli".generators | keys[] | sub(".*?-"; "")' | sort | uniq)"
+  echo "all $(cat $ROOT/config/clients.config.json | jq --arg lang "$1" 'with_entries(select(.key == "java")) | .[] | .clients')"
```
millotp, line 30: java ?

scripts/release/updateAPIVersions.ts
```diff
@@ -146,9 +145,6 @@ async function updateDartPackages(changelog: string, nextVersion: string): Promi
   // Version is sync'd on every clients so we set it once.
   clientsConfig.dart.packageVersion = nextVersion;
 
-  // update `openapitools.json` config file.
-  await writeJsonFile(toAbsolutePath('config/openapitools.json'), openapiConfig);
```
millotp, line 150: do we still need to modify the variable `openapiConfig` if it's not used ?

website/docs/contributing/release-process.md
```diff
@@ -77,11 +77,3 @@ dependencies {
 ```
 
 If it's not a snapshot, you can ignore the sonatype repository.
-
-### JavaScript
```
millotp, line 81: why is this gone ?

millotp, line 81: why is this gone ?

millotp (commented):
so nice !

generators/src/main/java/com/algolia/codegen/utils/Helpers.java
shortcuts: Yes but if the command contains all languages for java for example, there will be 10 elements

scripts/common.ts
shortcuts: Ah true

scripts/install.sh
shortcuts: Yes it has all clients and so we don't have to get all clients + uniq etc so I guess that's fine?

scripts/release/updateAPIVersions.ts
shortcuts: Hmmmm here I don't think so indeed

website/docs/contributing/release-process.md
shortcuts: We shouldn't recommend it and it's pretty reliable right now

generators/src/main/java/com/algolia/codegen/utils/Helpers.java
shortcuts: just re-read your msg, actually what @Fluf22 wants now is just the `clients` array in the clients config UUC

millotp: yes so there is no point in reading this file anymore, it should all come from the `config` folder

scripts/install.sh
millotp: not critical anyway

generators/src/main/java/com/algolia/codegen/utils/Helpers.java
shortcuts: for the use case that is in the codebase today it's fine it prevents us from adding more logic since it's just reading a variable

millotp:
the file is also referenced in `scripts/ci/githubActions/utils.ts` btw

shortcuts:
> scripts/ci/githubActions/utils.ts

good catch! fixed in https://github.com/algolia/api-clients-automation/pull/2481/commits/e0926c6665d078c4a5d06de59437a964c28dbdc4

millotp (approved):
looks great ! so much simpler !

PR #2497, opened 2024-01-08
chore(clients): add more e2e

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

add more e2e to some frequently methods used on the clients, I'll add some save e2e later on


millotp (approved):
so little change to create so powerful tests !

scripts/cts/runCts.ts
```diff
@@ -40,7 +40,7 @@ async function runCtsOne(language: string): Promise<void> {
       });
       break;
     case 'ruby':
-      await run(`bundle install && bundle exec rake test`, { cwd: 'tests/output/ruby' });
+      await run(`bundle install && bundle exec rake test --trace`, { cwd: 'tests/output/ruby' });
```
millotp, line 43: maybe we don't want to keep this

shortcuts: It helped me debugging but it's really verbose, up to you!

PR #2500, opened 2024-01-09
feat(clients): deserialize in e2e and SFFV response

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-1748

### Changes included:

We now deserialize responses in the generated models for Python, pydantic would raise an error if it failed

Also fix the search for facet values response which was missing properties

### DIFF
--- a/specs/search/common/schemas/SearchForFacetValuesResponse.yml
+++ b/specs/search/common/schemas/SearchForFacetValuesResponse.yml
@@ -3,6 +3,7 @@ searchForFacetValuesResponse:
   additionalProperties: false
   required:
     - facetHits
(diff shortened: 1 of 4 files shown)


shortcuts:
**2500**

millotp (commented):
the statusCode assertion has disappeared from the CTS ?

shortcuts:
> the statusCode assertion has disappeared from the CTS ?

only for endpoints with body assertions since the deserialized object doesn't contain it, but the `set_settings` still has it for example!

millotp (approved):
nice

PR #2501, opened 2024-01-09
feat(python): improve docstring and fix transporter errors

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

_This PR applies the feedback from @kai687's review_

- [x] provide some docstring comments to the main classes (I've followed https://github.com/google/styleguide/blob/gh-pages/pyguide.md#38-comments-and-docstrings)
- [x] fix deserialize error in the transporter when the call fails
- [x] add a `py.typed` file at the root of the package for libraries like `mypy`

Screenshot 2024-01-09 at 12 54 02

### DIFF
--- a/templates/python/api.mustache
+++ b/templates/python/api.mustache
@@ -13,13 +13,14 @@ class {{classname}}:
     Args:
     app_id (str): The Algolia application ID to retrieve information from.
     api_key (str): The Algolia api key bound to the given `app_id`.
+    {{#hasRegionalHost}}region ({{#allowedRegions}}"{{.}}"{{^-last}} | {{/-last}}{{/allowedRegions}}): The region of your Algolia application.{{/hasRegionalHost}}
 
     Returns:
     The initialized API client.
 
     Usage:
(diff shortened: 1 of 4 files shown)


templates/python/py.typed.mustache
millotp, line 1: this doesn't need to be a template, you can just add it in the clients folder

millotp (commented):
looks good !

templates/python/py.typed.mustache
shortcuts: > this doesn't need to be a template, you can just add it in the clients folder

true I put it there to avoid accidental deletion since this is something we can't really assert it but it shouldn't happen

templates/python/api.mustache
```diff
@@ -26,12 +43,33 @@ class {{classname}}:
...
 
+    async def __aenter__(self) -> None:
+        return self
+
+    async def __aexit__(self, exc_type, exc_value, traceback) -> None:
+        """Closes the underlying `transporter` of the API client."""
+        await self.close()
+
     async def close(self) -> None:
+        """Closes the underlying `transporter` of the API client."""
```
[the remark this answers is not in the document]
shortcuts: it's a bit redundant indeed but not a huge deal!

PR #2502, opened 2024-01-09
chore(ci): remove gradle wrapper validation from all languages

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket:

I don't think it improves the CI time but we download the bundle validation for java on every clients which it isn't needed

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -28,7 +28,7 @@ runs:
         cache: gradle
 
     - name: Validate gradle wrapper
-      if: inputs.type != 'minimal'
+      if: ${{ inputs.language == 'java' }}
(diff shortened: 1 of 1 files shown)


morganleroi (approved):
nice !

millotp (commented):
the wrapper is used for all language, without the validation we are open to code injection

shortcuts:
> the wrapper is used for all language, without the validation we are open to code injection

ah? isn't it what validates the java bundle with the current one?

millotp:
no this is another step https://github.com/algolia/api-clients-automation/blob/main/.github/workflows/check.yml#L348

shortcuts:
oh shoot aborting then D: great catch ty

PR #2504, opened 2024-01-09
fix(ruby): release using Trusted Publishing

millotp:
## 🧭 What and Why

Use the new feature of [Trusted Publishers] to simplify the release process.
The release will also push it's own tag so we don't need to do it from the APIC CI.


shortcuts (commented):
looks great! waiting for the release CI to finish

shortcuts (approved):
ah bah

PR #2510, opened 2024-01-09
chore(scripts): replace config path

millotp:
## 🧭 What and Why

By default, we need to rebuild the scripts everytime we change the the config files, but this is less than ideal.
The only way I found to fix it was by using a library, `tsc-alias`, that will replace after the transpilation every reference to `../config` with `../../../config` to fetch the original file.
Using `paths`, or `dynamic import` didn't work


shortcuts (approved):
I'm fine with that! great job finding an alternative, such a huge case D:

PR #2511, opened 2024-01-09
feat(clients): generate code snippets from cts

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-1749

### Changes included:

This PR registers a new generator in order to generate code snippets for each method of each languages, it is completely based on the CTS request generator and will take the first test case of each method as the example.

The snippets are generated for
- java
- javascript
- python
- php

#### Generators change

I had to change the output folder and template dir in order to fully reuse the `TestsRequest` generator, which means each template has been edited to include the `tests` folder when using partial templates

#### clients.config.json

A new field named `snippets` allows us to store the output folder and the file extension, similarly to what we do for tests

#### CI

We now store the snippets in the artifacts and restore them in order to be pushed by the CI

### What's next

- Add snippets for missing clients
- Improve first test case variable values and descriptions


Fluf22 (approved):
Ooh nice

generators/src/main/java/com/algolia/codegen/cts/AlgoliaCTSGenerator.java
```diff
@@ -63,6 +63,19 @@ public void processOpts() {
...
+    SnippetsGenerator snippetGenerator = new SnippetsGenerator(language, client);
+
+    if (!snippetGenerator.available()) {
+      return;
+    }
+
+    String snippetsExtension = Helpers.getClientConfigField(language, "snippets", "extension");
+    String snippetsOutputFolder = Helpers.getClientConfigField(language, "snippets", "outputFolder");
+
+    snippetGenerator.addSupportingFiles(supportingFiles, snippetsOutputFolder, snippetsExtension);
```
millotp, line 76: the `SnippetsGenerator` should do that logic itself

scripts/cts/generate.ts
```diff
@@ -43,6 +43,9 @@ export async function ctsGenerateMany(generators: Generator[]): Promise<void> {
       await run('YARN_ENABLE_IMMUTABLE_INSTALLS=false yarn install', {
         cwd: 'tests/output/javascript',
       });
+      await run('YARN_ENABLE_IMMUTABLE_INSTALLS=false yarn install', {
```
millotp, line 46: I think this should be in a totally different file, like `scripts/snippets/generate.ts`

millotp (commented):
so cool, almost nothing to add !

generators/src/main/java/com/algolia/codegen/cts/AlgoliaCTSGenerator.java
shortcuts: ah yes it can

scripts/cts/generate.ts
shortcuts: I did not yet created the CLI stuff, I know you want to do it :D

millotp (approved):
let's start with that !

PR #2512, opened 2024-01-09
fix(ruby): exclude vendor from bundled gem

millotp:
## 🧭 What and Why

The published gem was 5.4MB instead of the expected `300KB`, because I forgot to remove files excluded by git.


millotp:
if I don't specify it it won't !

PR #2513, opened 2024-01-10
fix(cts): generate tests in the kotlin fashion

millotp:
## 🧭 What and Why

In order to generate pretty code snippets, we can use the same technic as @aallam did for kotlin, and don't generate any variables, just inline everything.
This simplifies everything, and we don't need to deal with any generics anymore !


shortcuts (approved):
I love it! we need it for csharp too right?

PR #2514, opened 2024-01-10
fix(go): use better parameters

millotp:
## 🧭 What and Why

Make the go CTS use the real models, and not just the json.
This involves reworking a lot of the types because it's a mess between pointer and not pointer.


morganleroi:
Huuuuuge

shortcuts (approved):
feat fixes and chore

in-sane gg

PR #2515, opened 2024-01-11
chore(scripts): remove interactive option

millotp:
## 🧭 What and Why

Now that we have bash completion, the interactive cli doesn't make sense anymore, and it was never used anyway


shortcuts (approved):
Nice cleanup!

PR #2516, opened 2024-01-11
chore(scripts): fix install and remove deps

millotp:
## 🧭 What and Why

APIC autocomplete is a bit broken for some functions.
Also remove unused deps.


shortcuts (approved):
lovely

PR #2517, opened 2024-01-11
fix(csharp): enforce dotnet sdk version

morganleroi:
## 🧭 What and Why

Enforce use of dotnet SDK specific version when using the CLI (CI).


shortcuts (approved):

generators/src/main/java/com/algolia/codegen/AlgoliaCSharpGenerator.java
```diff
@@ -41,10 +45,20 @@ public String getClientName(String client) {
...
     CLIENT = (String) additionalProperties.get("client");
+
+    Stream<String> stream = null;
+    try {
+      stream = Files.lines(Paths.get("config/.csharp-version"));
+    } catch (IOException e) {
+      e.printStackTrace();
+      System.exit(1);
+    }
+
```
millotp, line 56: maybe just `Files.readString(fileName);`, and remove trailing `\n`, as you want

millotp (commented):
thanks for finding a fix so quickly !

shortcuts (commented):
can you edit https://github.com/algolia/api-clients-automation/blob/main/scripts/buildClients.ts#L23-L26 to be `find .` to avoid searching `cwd` in `cwd `pls?

shortcuts (dismissed):
gg guys

millotp (approved):

PR #2519, opened 2024-01-11
fix(scripts): own cli for snippets

millotp:
## 🧭 What and Why

Split the cli for snippets because it will have it's own custom logic and we don't want to mix everything, will still reuse as much of the CTS as possible.

Also stop using the generated file `openapitools.json` because it was creating incorrect `tests/output/javascript/package.json` when generating for a single client, use the original `clients.config.json` file instead.


generators/src/main/java/com/algolia/codegen/cts/manager/JavaCTSManager.java
```diff
@@ -14,7 +14,7 @@ public JavaCTSManager(String client) {
   }
 
   @Override
-  public void addSupportingFiles(List<SupportingFile> supportingFiles) {
+  public void addTestsSupportingFiles(List<SupportingFile> supportingFiles) {
```
shortcuts, line 17: here we add snippets supporting files

generators/src/main/java/com/algolia/codegen/cts/manager/JavascriptCTSManager.java
```diff
@@ -26,41 +26,26 @@ public void addDataToBundle(Map<String, Object> bundle) throws GeneratorExceptio
...
-    while (fields.hasNext()) {
-      Map.Entry<String, JsonNode> field = fields.next();
-
-      if (!field.getKey().startsWith("javascript-")) {
-        continue;
-      }
-
-      Map<String, String> client = new HashMap<>();
-      String output = field.getValue().get("output").asText();
+    Iterator<JsonNode> clientIterator = Helpers.getClientConfig("javascript").get("clients").elements();
```
shortcuts, line 32: cc @Fluf22 IIRC you made a different implementation one of you will have conflicts :D

scripts/snippets/generate.ts
```diff
@@ -0,0 +1,53 @@
...
+
+async function snippetsGenerate(gen: Generator): Promise<void> {
+  const spinner = createSpinner(`generating code snippets for ${gen.key}`);
+
+  await run(
+    `yarn openapi-generator-cli --custom-generator=generators/build/libs/algolia-java-openapi-generator-1.0.0.jar generate \
+     -g algolia-cts -i specs/bundled/${gen.client}.yml --additional-properties="language=${gen.language},client=${gen.client},mode=snippets"`
+  );
+  spinner.succeed();
+}
```
shortcuts, line 17: this could be common for the CTS and the Snippets and take the mode as parameter?

scripts/snippets/generate.ts
```diff
@@ -0,0 +1,53 @@
...
+        cwd: 'snippets/javascript',
+      });
+    }
+
+    const snippetsPath = toAbsolutePath(`snippets/${lang}`);
+    if (await exists(snippetsPath)) {
+      await formatter(lang, snippetsPath);
+    }
+  }
+}
```
shortcuts, line 53: same for this I guess

shortcuts (commented):
lovelyyyy

generators/src/main/java/com/algolia/codegen/cts/manager/JavascriptCTSManager.java
Fluf22: Nope, I added a new method, so it won't conflict, I think

shortcuts: but I mean they do the same thing I believe

Fluf22: I could test this method, but I don't know if it will work with mustache?

scripts/snippets/generate.ts
millotp: yes but this is just the start, snippets will have more logic for file splitting and formatting maybe, I didn't want to have `if(snippets)` everywhere

shortcuts: they will definitely be different but I think it will be post-processing the currently generated files so it should stay pretty similar, at least for the lines 20 to 35 it will for sure

millotp: okay I added `setupAndGen` that does all the common stuff

generators/src/main/java/com/algolia/codegen/AlgoliaJavascriptGenerator.java
```diff
@@ -135,6 +136,7 @@ public String getPackageName(String client) throws ConfigException {
   /** Set default generator options */
   private void setDefaultGeneratorOptions() {
     String apiName = CLIENT + Helpers.API_SUFFIX;
+    System.out.println(CLIENT);
```
shortcuts, line 139:
```suggestion
```

shortcuts (dismissed):
love it

shortcuts (dismissed):

shortcuts (approved):
GGGGG

PR #2522, opened 2024-01-11
test(csharp): inline parameters

morganleroi:
## 🧭 What and Why

🎟 JIRA Ticket: DI-1716

### Changes included:

Inline parameters for CTS and Snippets.


scripts/formatter.ts
```diff
@@ -5,7 +5,11 @@ export async function formatter(language: string, cwd: string): Promise<void> {
   const spinner = createSpinner(`running formatter for '${language}' in '${cwd}'`);
   switch (language) {
     case 'csharp':
-      await run('dotnet format', { cwd, language });
+      if (cwd.includes('tests')) {
```
shortcuts, line 8: this needs `snippets` now too

tests/output/csharp/.editorconfig
shortcuts, line 1: can you apply this to the snippets too?

shortcuts (commented):
❤️

scripts/formatter.ts
morganleroi: Yes ok ! I'll try to generate snippet and take a look at the result

scripts/formatter.ts
```diff
@@ -5,7 +5,11 @@ export async function formatter(language: string, cwd: string): Promise<void> {
   const spinner = createSpinner(`running formatter for '${language}' in '${cwd}'`);
   switch (language) {
     case 'csharp':
-      await run('dotnet format', { cwd, language });
+      if (cwd.includes('tests') || cwd.includes('snippets')) {
```
millotp, line 8: why not use csharpier for the client too ?

millotp (commented):
looks way better gg !

PR #2525, opened 2024-01-12
chore(ci): use docker

millotp:
## 🧭 What and Why

Instead of initizalizing everything locally and on the CI, we can reuse the same docker image to save a lot of time.


shortcuts (approved):
I LOVE IT GG

millotp:
this doesn't seem to be faster, I think we need to shrink the image size a lot

shortcuts:
> this doesn't seem to be faster, I think we need to shrink the image size a lot

If fine with it not being faster personally as long as it reduces the friction between local/CI it's still a huge improvement

morganleroi:
Wooo this is indeed a big improvment if it does not kills the CI duration !

PR #2545, opened 2024-01-15
chore: support fish for setup

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

you can't setup docker with fish actually because it sets the env var from substitutions, but with bass (fish plugin) it works if we use export directly in the script we are running

### DIFF
--- a/scripts/docker/setup.sh
+++ b/scripts/docker/setup.sh
@@ -1,5 +1,10 @@
 #! /bin/bash
 
 export NODE_VERSION=$(cat .nvmrc)
-find config -name '.*-version' | xargs -I{} sh -c 'l=$(echo "{}" | sed -e "s/-/_/;s/config\/\.//" | tr "[a-z]" "[A-Z]");echo "export $l=$(cat {})"'
+
+while read line; do
+  arr=($line)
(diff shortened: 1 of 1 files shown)


scripts/docker/setup.sh
```diff
@@ -1,5 +1,10 @@
 #! /bin/bash
 
 export NODE_VERSION=$(cat .nvmrc)
-find config -name '.*-version' | xargs -I{} sh -c 'l=$(echo "{}" | sed -e "s/-/_/;s/config\/\.//" | tr "[a-z]" "[A-Z]");echo "export $l=$(cat {})"'
+
+while read line; do
+  arr=($line)
+  export $(echo ${arr[1]} | sed -e "s/-/_/;s/config\/\.//" | tr "[a-z]" "[A-Z]")=${arr[0]}
+done < <(find config -name '.*-version' -exec jq --raw-input -r '. + " " + input_filename' {} \;)
```
shortcuts, line 8: i can't remember who to do a while on 2 variables outputted by jq so I used a space to split it 🤷🏻 that's fineeee

millotp (approved):

PR #2546, opened 2024-01-15
chore(scripts): check the generated files for cache

millotp:
## 🧭 What and Why

The cache only checked for the source file, not the generated ones, so if you deleted the bundled spec for example the cache would consider it valid.

Fix it by also included the hash of the generated files.


shortcuts:
> ## 🧭 What and Why
> The cache only checked for the source file, not the generated ones, so if you deleted the bundled spec for example the cache would consider it valid.
> 
> Fix it by also included the hash of the generated files.

alternatively we can ensure the file we are checking the cache from exists but that's fine too

millotp:
> alternatively we can ensure the file we are checking the cache from exists but that's fine too

that's what we did before, but it's not enough when the generated files changed

shortcuts (approved):
looks clean!

PR #2548, opened 2024-01-16
fix(javascript): lock

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

checksum isn't computed the same way on the standalone js repo and the automation repo, the only thing that changed was the `.yarnrc`, now that they are the same there's no diff anymore

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -1,7 +1,7 @@
 compressionLevel: mixed
 
-enableGlobalCache: false
-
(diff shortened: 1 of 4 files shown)


millotp (approved):
that's weird, great find !

shortcuts:
> that's weird, great find !

I assume the `compressionLevel` changes the checksu

PR #2550, opened 2024-01-16
fix(swift): rewrite networking layer

Fluf22:
## 🧭 What and Why

🎟 JIRA Ticket: [DI-1722]

### Changes included:

- broader use of async/await instead of legacy queueing mechanism
- similar transporter/requester pattern to the one used in the other APIC

[DI-1722]: https://algolia.atlassian.net/browse/DI-1722?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


morganleroi:
Huuuuge work !

clients/algoliasearch-client-swift/Sources/Core/Common/RequestOptions.swift
```diff
@@ -28,6 +28,47 @@ public class RequestOptions {
     self.body = body
   }
 
+  public static func + (lhs: RequestOptions, rhs: RequestOptions?) -> RequestOptions {
```
shortcuts, line 31: _maybe_ we can do like `baseRequestOptions` and `userRequestOptions` which returns `finalRequestOptions` or something like this

clients/algoliasearch-client-swift/Sources/Core/Networking/HTTP/HTTPMethod.swift
```diff
@@ -0,0 +1,20 @@
...
+public enum HTTPMethod: String {
+  case options = "OPTIONS"
+  case get = "GET"
+  case head = "HEAD"
+  case post = "POST"
+  case put = "PUT"
+  case patch = "PATCH"
+  case delete = "DELETE"
+  case trace = "TRACE"
+  case connect = "CONNECT"
```
shortcuts, line 19: we don't support all of them but I'm fine with that

clients/algoliasearch-client-swift/Sources/Core/Networking/Transporter.swift
```diff
@@ -8,128 +8,141 @@ import Foundation
...
+      self.requestBuilder =
+        requestBuilder
+        ?? URLSessionRequestBuilder(
+          sessionConfiguration: sessionConfiguration
+        )
+
+      return
+    }
+
+    self.requestBuilder = requestBuilder
```
shortcuts, line 42: maybe early return if requestBuilder so we don't initialize without using the values?

not sure IIUC though

templates/swift/Package.mustache
```diff
@@ -23,7 +23,7 @@ let package = Package(
   name: "{{projectName}}",
   platforms: [
     .iOS(.v13),
-    .macOS(.v10_13),
+    .macOS(.v10_15),
```
shortcuts, line 26: is this the minimal version to support?

shortcuts (commented):
looks great! some nitpick comments but I don't know much about swift

templates/swift/Package.mustache
Fluf22: Yes!

clients/algoliasearch-client-swift/Sources/Core/Networking/Transporter.swift
Fluf22: Yes, it's what the `guard` does: if the `requestBuilder` is not `nil` it instantiates it and return, otherwise, it just returns the given value

clients/algoliasearch-client-swift/Sources/Core/Common/RequestOptions.swift
Fluf22: You mean just updating the variable name?
I can do that

templates/swift/Package.mustache
shortcuts: ok make sense! we might want to set that in stone at some point I guess or is it automatic with sift?

Fluf22: We will definitely need to document that!
It's the SDK version that allows us to use some methods or tells us some old ones are deprecated 
But as it's in the Swift package, users are warned (it's like using browser versions)

shortcuts (approved):
ggggg

PR #2553, opened 2024-01-16
fix(specs): rename clearAllSynonyms to clearSynonyms

kai687:
## 🧭 What and Why

To be consistent with `clearObjects()` and `clearRules()` methods, rename this method to `clearSynonyms()`.

🎟 JIRA Ticket: N/A ([Slack](https://algolia.slack.com/archives/C05C3DRTVFU/p1705419628079899))

### Changes included:

- Rename filename and `operationId`


shortcuts (commented):
looks great! you also need to rename the test file of the same name

PR #2554, opened 2024-01-16
snippets: remove index from csharp

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

snippets include index of the test which doesn't have much context here, we can use the client name instead to prevent duplicated defintions


morganleroi (approved):
Thanks !

PR #2556, opened 2024-01-16
test(csharp): add e2e test suite

morganleroi:
## 🧭 What and Why

🎟 JIRA Ticket: DI-1766

### Changes included:
- Add e2e tests


Fluf22 (approved):
Nice work!

templates/csharp/tests/requests/requests.mustache
```diff
@@ -42,7 +60,7 @@ public class {{client}}RequestTests
      Assert.Equal("{{{path}}}",req.Path);
      Assert.Equal("{{{method}}}",req.Method.ToString());
     {{#body}}
-        JsonAssert.EqualOverrideDefault("{{#lambda.escapeQuotes}}{{{body}}}{{/lambda.escapeQuotes}}", req.Body, new JsonDiffConfig(true));
+        JsonAssert.EqualOverrideDefault("{{#lambda.escapeQuotes}}{{{body}}}{{/lambda.escapeQuotes}}", req.Body, new JsonDiffConfig(false));
```
millotp, line 63: what does `JsonDiffConfig` do ? I can't find the doc

millotp (approved):
nice !

templates/csharp/tests/requests/requests.mustache
morganleroi: The `JsonDiffConfig` allows additional properties to be added without error during assertion. (if `true`).

For the request with Echo mock, it was a mistake to allow additional properties to be added without failing. For the e2e, it's still necessary as some properties are added to the payload but we don't want to assert then (Request duration, .Processing time, ...). Does it make sense ?

PR #2557, opened 2024-01-17
feat(scripts): lint ACLs in the specs

millotp:
## 🧭 What and Why

Check that the `x-acl` property is compliant with Algolia ACLs.


shortcuts (approved):
wonderful, love the idea!!

PR #2558, opened 2024-01-17
fix(python): full overwrite `data` with request_options

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-1784

### Changes included:

we should fully overwrite the `data` when the user provide it in the `request_options`


Fluf22 (dismissed):
Yep, matches the other clients!

shortcuts:
> Yep, matches the other clients!

Thanks for catching it!

shortcuts:
btw I don't think we assert this somewhere

millotp (approved):

PR #2559, opened 2024-01-17
fix(ruby): correct oneOf deserialization

millotp:
## 🧭 What and Why

The ruby client was broken because of the change of namespaces.


shortcuts (approved):
great catch!

PR #2560, opened 2024-01-17
chore(ci): cache golangci-lint

millotp:
## 🧭 What and Why

Just like [golangci-lint-action](https://github.com/golangci/golangci-lint-action), we can cache the results of `golangci-lint` (about 10MB) and have a huge speed up when linting.
Of course this is dwarfed by swift :(


shortcuts (approved):
Once swiftformat is fixed i'm sure this will be HUGE

PR #2561, opened 2024-01-17
fix(generators): add uniqueHost property

morganleroi:
## 🧭 What and Why

Monitoring API configuration is not correctly generated. We use the `{appId}-dsn.algolia.net` format but this host does not exist for the Monitoring API. 

[the spec is correct](https://github.com/algolia/api-clients-automation/blob/679077d0e8a9954fa8d3fa2bc0e846139ec50dcc/specs/bundled/monitoring.yml#L13) and only define [the unique available host](https://www.algolia.com/doc/rest-api/monitoring/#status-endpoints). 

We need to introduce a new property `uniqueHost ` to avoid fallback to `{appId}-dsn.algolia.net` format in our configuration files. 

C# example is provided.

🎟 JIRA Ticket:


Fluf22 (approved):
Thank you for this fix!
I knew something was wrong with my Monitoring client 😄

millotp (approved):
looks good ! maybe it could be simplified with more variables, like `isSingleHost`, `isRegionalHosts`, `isAppIDHosts`... I'm not sure

shortcuts:
that's nice good catch! do we contribute to your pr directly?

Fluf22 (approved):
I would say we should each do our PRs separately

millotp:
The CTS should be aware of that change too, the client test for monitoring is not correct right now

shortcuts:
> The CTS should be aware of that change too, the client test for monitoring is not correct right now

I don't think we assert hosts yet but we can add one of course

millotp:
> I don't think we assert hosts yet

yes we do ! in all clients ! except monitoring

shortcuts:
> > I don't think we assert hosts yet
> 
> yes we do ! in all clients ! except monitoring

:ah: let's fix that once all clients have implemented the uniqueHost so we don't have to block everyone until they fix their CTS

PR #2564, opened 2024-01-17
feat(csharp): add playground for all APIs

morganleroi:
## 🧭 What and Why

🎟 JIRA Ticket:

### Changes included:

- List changes


shortcuts:
makes me think we could generate playgrounds... and just never push changes we made on them

cc @millotp I'm sure you'll want to implement that right away

morganleroi:
It's done @millotp  !

playground/csharp/Playground/appsettings.example.json
```diff
@@ -0,0 +1,8 @@
+{
+  "AppID": "XXX",
+  "AdminApiKey": "XXXXXXXXXXXXXX",
```
millotp, line 3: you can't use the `.env` ?

millotp (commented):
nice !

playground/csharp/Playground/appsettings.example.json
morganleroi: `.env` is not the standard way to pass secret in dotnet. I can switch to `.env` file if you want. Dotnet dev are aware of the `appsetting` feature.

millotp: we are not Dotnet devs ahah, we don't want specific things for specific language, everything should use the same thing. Maybe you can leave both and default to `appsettings` if there is no .env ??

morganleroi: ok, let me add .env instead. Indeed in the head, the playground was mainly for our customers. But maybe not.

morganleroi: done

millotp (commented):
Thanks for the change ! last comment my bad

playground/csharp/Playground/Utils/Configuration.cs
```diff
@@ -0,0 +1,31 @@
+using dotenv.net;
+
+namespace Algolia.Utils;
+
+public static class Config
+{
+  public static Configuration Load()
+  {
+    DotEnv.Load(options: new DotEnvOptions(ignoreExceptions: false, envFilePaths: new[] { "./Playground/.env"}));
```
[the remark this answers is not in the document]
morganleroi: ok done. For now, I still need other env var so I use both .env files.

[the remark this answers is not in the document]
millotp: you could add them to the `.env.example` :)

millotp (approved):
gg !

PR #2566, opened 2024-01-17
fix(cts): reuse same logic for all client tests

millotp:
## 🧭 What and Why

this was way too hard, but we can simulate dynamic templates with lambda even though the mustache library that openapi generator is using doesn't support dynamic templates !
This allows us to remove the `step.mustache` template, and just rely on a variable.

Dart and Kotlin are too different to use the same format for now, If I have time I'll try to convert them too.

I'm not sure if this PR is worth it but it was a nice challenge, and if we add more steps in the future in will make a difference.


tests/output/python/requirements.txt
```diff
@@ -0,0 +1,6 @@
+argcomplete==3.2.1
```
Fluf22, line 1: I feel like it's not the first time I've seen this file pushed 🤔

Fluf22 (dismissed):
I didn't understand much, but I trust you about the utility 😄 You'll have to make us a tutorial, though 😂

shortcuts (commented):
Looks great! Too bad it's not native :(

I like the idea but I'm not sure if it's better to have less mustache logic or document a custom implementation, up to you!

millotp:
I'll try to add more client tests to show the merit of this approach :)

tests/output/python/requirements.txt
millotp: I remove it, it's generated

Fluf22 (approved):

PR #2569, opened 2024-01-18
fix(java): correct url for monitoring client

millotp:
## 🧭 What and Why

Following #2561, use the correct url for java, and rework the generators to use the already read spec file, and the correct types for servers.


generators/src/main/java/com/algolia/codegen/AlgoliaScalaGenerator.java
```diff
@@ -90,6 +90,12 @@ private String typeToName(String content) {
     return content.trim().replace("[", "Of").replace("]", "").replace(".", "").replace(", ", "");
   }
 
+  @Override
```
morganleroi, line 93: How did the client worked without this ?

millotp: it was called line 80 before, I just changed the way the function work so I had to move it here

morganleroi (approved):
Puuurfect !

PR #2570, opened 2024-01-18
fix(cts): client test for monitoring

millotp:
## 🧭 What and Why

following #2561, we need to add tests to make sure every implementation is correct


millotp:
this is expecting to fail until all languages support it.

Fluf22 (approved):
One more to go

PR #2576, opened 2024-01-19
docs(clients): add ACL section to code doc

millotp:
## 🧭 What and Why

As suggested by @morganleroi and @shortcuts, we can add the required ACLs to the clients doc, above each function.
We can't do that by simply adding it to the `{{notes}}` because each language has it's own way to handle multiline comments.

I didn't do java for now, because the javadoc is messed up, will do it in another PR


morganleroi:
Huuuuuuge

templates/csharp/libraries/httpclient/api.mustache
```diff
@@ -102,7 +102,17 @@ namespace Algolia.Search.Clients
         {{#supportsAsync}}
 
         /// <summary>
-        /// {{{summary}}} {{notes}}
+        /// {{{summary}}}
```
shortcuts, line 105: maybe just the `description` is enough, IIRC it's required so it removes the redundancy of having both

shortcuts: (applies to all languages IMHO)

templates/swift/api.mustache
```diff
@@ -79,7 +79,15 @@ typealias Client = {{classname}}
...
-     {{{.}}}{{/notes}}{{#subresourceOperation}}
+     {{{.}}}{{/notes}}
+     {{#vendorExtensions}}
+     {{#x-acl.0}}
+
+     Required API Key ACLs:{{/x-acl.0}}
+     {{#x-acl}}
+       - {{.}}
+     {{/x-acl}}
+     {{/vendorExtensions}}{{#subresourceOperation}}
```
shortcuts, line 90: this seems to add newlines even when there's no ACLs to display

shortcuts (commented):
so nice

templates/csharp/libraries/httpclient/api.mustache
millotp: there is no description on operations, just summary and notes, maybe I can just leave the notes yes

Fluf22 (commented):
I'm not sure that's the best option for Swift, because it removes a useful short description for some operations.
But this can be looked at later!

Fluf22 (approved):

PR #2595, opened 2024-01-22
fix(specs): correct `searchSynonyms` parameters

morganleroi:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

The parameters of the `searchSynonyms` method are in fact body parameters and not query parameters, this PR also fixes the browse helpers


morganleroi:
@millotp @shortcuts do you an idea why this change broke the CTS generation ?

specs/search/paths/synonyms/searchSynonyms.yml
```diff
@@ -24,6 +21,12 @@ post:
           properties:
             query:
               $ref: '../../../common/schemas/SearchParams.yml#/query'
+            type:
+              $ref: './common/parameters.yml#/SynonymType'
```
shortcuts, line 25: I guess `type` is reserved in many languages and fails?

morganleroi: But it's only the CTS that fails to generate. `Error in searchSynonyms with all parameters: Parameter type not found in the root parameter`

I'll try to dig into the generator.

shortcuts: yup that's weird but if you try to gen the python client it doesn't work 🤔

shortcuts (approved):
great catch!! gg

PR #2600, opened 2024-01-24
fix(scripts): check for last commit in js

millotp:
## 🧭 What and Why

This step didn't trigger on the last release, maybe because the condition was wrong, let's try in js instead.


shortcuts (commented):
same for the spread to new world it did not triggered so I guess that's a problem elsewhere :/

millotp:
ah for push to new world you have to set `setVerbose(false)` otherwise the output of the git functions is messed up, I think we had this issue in other scripts

shortcuts (approved):
good catch

PR #2603, opened 2024-01-25
chore(swift): use another formatter

millotp:
## 🧭 What and Why

This removes java and node from swift and ruby images, because they don't use it locally, and it's too hard to make it work on the CI.
This also uses another formatter for swift, which doesn't require compiling the entire universe to install it.


Fluf22 (approved):
Let's go!

PR #2604, opened 2024-01-25
feat(specs): add acls to API endpoints

alphonseb:
## 🧭 What and Why

Adding the ACLs to the QS api endpoints spec. Reference is https://github.com/algolia/go/blob/master/cmd/query-suggestions/query-suggestions-api/service/service.go


millotp (approved):
thanks for the contribution !

PR #2605, opened 2024-01-25
feat(cts): add helpers to the specs

millotp:
## 🧭 What and Why

This is a POC to add the search helpers to the OpenAPI Specs, which has several advantages:
- Common place for the method definition, we know what they expect and what they return.
- It automatically generates every enum ! Same in every languages (this is +++)
- We can generates tests with strict types !
- Automatic documentation of our helpers ?

Inconvenients:
- Specs are not made to write helpers, we have to provide a few useless properties, but we can work around that with tooling.


millotp:
Actually we need all e2e before this can work...

scripts/pre-gen/generateOpenapitools.ts
```diff
@@ -25,10 +25,12 @@ export async function generateOpenapitools(gens: Generator[]): Promise<void> {
       gitRepoId: getClientsConfigField(language, 'gitRepoId'),
       glob: `specs/bundled/${client}.yml`,
       templateDir,
-      generatorName: `algolia-${language}`,
-      output: `#{cwd}/${rest.output}`,
+      generatorName: `algolia-${mode !== 'client' ? 'cts' : language}`,
+      output: `#{cwd}/${mode === 'client' ? rest.output : ''}`,
```
shortcuts, line 29: same ternary operator maybe just for consistency

```suggestion
      generatorName: `algolia-${mode === 'client' ? language : 'cts'}`,
      output: `#{cwd}/${mode === 'client' ? rest.output : ''}`,
```

specs/search/helpers/generateSecuredApiKey.yml
```diff
@@ -0,0 +1,71 @@
...
+          $ref: '#/securedApiKeyRestriction'
+    responses:
+      '200':
+        description: OK
+        content:
+          application/json:
+            schema:
+              type: string
+      '400':
+        $ref: '../../common/responses/IndexNotFound.yml'
```
shortcuts, line 36:
```suggestion
```

I guess we can omit those

shortcuts: just keep the happy path and specify it's unused?

templates/javascript/clients/client/model/clientMethodProps.mustache
```diff
@@ -109,7 +109,7 @@ export type GenerateSecuredApiKeyOptions = {
   /**
    * A set of properties defining the restrictions of the secured API key.
    */
-  restrictions?: SecuredApiKeyRestrictions;
+  restrictions?: SecuredApiKeyRestriction;
```
shortcuts, line 112: there's many so maybe keep the plural form?

shortcuts (commented):
awesome, really nice addition!

specs/search/helpers/generateSecuredApiKey.yml
millotp: redocly is not happy if I remove it, I left the bare minimum to pass to the linters

scripts/playground.ts
```diff
@@ -64,6 +64,9 @@ export async function playground({
     case 'scala':
       // run scala playground
       break;
+    case 'swift':
+      await run(`swift run`, { cwd: 'playground/swift/playground/playground', language });
```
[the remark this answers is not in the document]
millotp: yeah and it does't work actually, I'll remove it for now

templates/javascript/clients/client/model/clientMethodProps.mustache
shortcuts: not every clients, js is in plural for example

millotp: 1 out of 11 ahah

shortcuts: yes but...... he is right :D

millotp: renamed, you won

shortcuts (approved):
nicer with an S!!!!!!!!!!

PR #2607, opened 2024-01-26
chore(ci): disable yarn hardened mode

millotp:
## 🧭 What and Why

disable yarn hardened mode because it takes 30-60s more time to run that a simple `yarn install`, and we don't run untrusted CI anyway.

Also try swift on linux, to see the speed difference


Fluf22 (approved):
More speed 🤩

Fluf22 (approved):

PR #2609, opened 2024-01-26
feat(cts): add e2e for go

millotp:
## 🧭 What and Why

Added e2e for go, once again it helped to find a lot of bug, including a very big one that caused the http request to close before reading the body, so it would fail but only on large bodies !

Also improved drastrically the error reporting by wrapping useful errors.

Removed the debug config because we already have the `debug.Enable` which works better (but not thread safe).


Fluf22 (approved):
Everything looks good!

PR #2627, opened 2024-01-29
feat(csharp): add compression

morganleroi:
## 🧭 What and Why

🎟 JIRA Ticket: DI-1731

### Changes included:
- Add GZip Compression implementation


tests/output/csharp/src/tests/CompressionTests.cs
millotp, line 1: this folder is not meant for hand written test, maybe this should be in the client folder directly, like we do for js

millotp (commented):
looks great !

tests/output/csharp/src/tests/CompressionTests.cs
morganleroi: In dotnet we separate tests and production code in two different projects. Maybe I can create a separate folder for theses tests, but inside the tests assembly ? WDYT ?

millotp: yes, or maybe just outside of src here

morganleroi: done

millotp (dismissed):

Fluf22 (approved):
Let's go!

PR #2631, opened 2024-01-29
feat(csharp): add custom user-agent segment

morganleroi:
## 🧭 What and Why

🎟 JIRA Ticket:

### Changes included:

- List changes


tests/output/csharp/src/AddSegmentToUserAgentTests.cs
millotp, line 1: this could be part of the clients test suite too

millotp (approved):
nice

PR #2632, opened 2024-01-29
fix(scripts): push origin for new docs

millotp:
## 🧭 What and Why

The CI failed on main, because the git command is malformed.


shortcuts (approved):
Tyyy

PR #2639, opened 2024-01-30
fix(go): support http

millotp:
## 🧭 What and Why

Go path encoding is messed up, and it doesn't support http


Fluf22 (approved):
We do this to support localhost testing, right?
Because swift doesn't support `http` either, right now 😢 I'll have to do like you

millotp:
Yes this is a fix for #2633

PR #2673, opened 2024-02-06
fix(cts): add e2e for abtesting

millotp:
## 🧭 What and Why

Add an e2e for abtesting that uses a space in the query parameters, to make sure it works with the new encoding.
AB Tests can only last 90 days however, so we need to recreate an abtest an update this test every 3 month, might be worth to automate it.

Here is the request to create the abtest with the `addABTests` endpoint:
```json
 {
    "endAt": "2024-05-06T09:04:26.469Z", // update this to be NOW + 90 days
    "name": "cts_e2e_abtest",
    "variants": [
      {
        "index": "cts_e2e_search_facet",
        "trafficPercentage": 25
      },
      {
        "index": "cts_e2e abtest",
        "trafficPercentage": 75
      }
    ]
  }
```


Fluf22 (approved):
🆗

Fluf22 (dismissed):
This is totally gonna conflict with my current PR :lolsob:

Fluf22 (approved):
finally

PR #2674, opened 2024-02-07
feat(csharp): add sync method overload

morganleroi:
## 🧭 What and Why

After discussing with C# contributors (Legacy), we decided to keep Synchronous support methods for compatibility reason. 

### Changes included:
- Add Async Wrapper helper


Fluf22 (approved):
Clean!

PR #2675, opened 2024-02-07
remove all warns from assembly.

morganleroi:
## 🧭 What and Why

Remove all warns from the assembly.
- We were using Nullable reference type.
- Missing GetHashCode Override when Equals override.

image


millotp (approved):
nice !

PR #2681, opened 2024-02-08
fix(cts): strict comparaison of query parameters

millotp:
## 🧭 What and Why

Unknown query parameters are not allowed in most APIs, we need to make sure that the client only sends what was requested, and nothing else.
Enforce strict check in all languages, and fix python by initializing every member of a class, otherwise they are treated as static member ! (so weird)
No need to initialize models for python, because they all extend the `BaseModel` which initializes everything properly (from what I understand, maybe I'm wrong).

also some small fixes in other languages to make the CI pass, and locally too.


morganleroi:
Do I need to double check if all assertion are correctly done in C# ?

millotp:
I checked for all languages, but maybe I got some wrong !

morganleroi (approved):
GéGé

PR #2704, opened 2024-02-12
feat(cts): add e2e tests

Fluf22:
## 🧭 What and Why

🎟 JIRA Ticket: [DI-1791]

### Changes included:

- add e2e tests support
- enable formatter rules
- fix qp encoding
- add qp key encoding
- remove Builder protocol as it's not compatible with Linux (it triggers encoding-computing on the other URLComponents' properties)

[DI-1791]: https://algolia.atlassian.net/browse/DI-1791?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2708, opened 2024-02-13
fix(cts): headers should not be encoded

millotp:
## 🧭 What and Why

This PR makes sure that headers are not encoded the same way as query parameters, and it also check for recursive encoding of query parameters.


Fluf22:
#2704 should fix Swift here

Fluf22:
💙

PR #2709, opened 2024-02-14
feat(csharp): add requestion option builder helper

morganleroi:
## 🧭 What and Why

When working on Guides, I added this helper to help user to build request options. It wasn't really convenient. The "old" 
 way is still valid ofc, user can still forge himself the `RequestOption` object.

Guides, Snippets and CTS Updated.


morganleroi:
Forgot to update the CTS

millotp (approved):
nice, I fixed the issue differently in #2708

PR #2710, opened 2024-02-14
fix(csharp): add nextDelay parameter in Wait helpers

morganleroi:
## 🧭 What and Why
Add `nextDelay` parameter to all `Wait` helpers. Detected this lack during Guides writting.
Also fix a typo on default parameter.


clients/algoliasearch-client-csharp/algoliasearch/Utils/ClientExtensions.cs
```diff
@@ -24,14 +24,16 @@ public static class ClientExtensions
   /// <param name="indexName">The `indexName` where the operation was performed.</param>
   /// <param name="taskId">The `taskID` returned in the method response.</param>
   /// <param name="maxRetries">The maximum number of retry. 50 by default. (optional)</param>
+  /// <param name="nextDelayFunc">The function to decide how long to wait between retries. Math.Min(retryCount * 200, 5000) by default. (optional)</param>
```
shortcuts, line 27: this is called `timeout` in other clients no? we should align on naming

morganleroi: Yes it's called timeout indeed. But I thought it wasn't clear about that it's actually doing. Will change it to timeout to align.

shortcuts: Or rename timeout everywhere to something that suits best! I don't mind personally

clients/algoliasearch-client-csharp/algoliasearch/Utils/ClientExtensions.cs
```diff
@@ -328,8 +336,10 @@ public static class ClientExtensions
     AsyncHelper.RunSync(() => client.SearchForFacetsAsync(requests, searchStrategy, options, cancellationToken));
 
   private static async Task<T> RetryUntil<T>(Func<Task<T>> func, Func<T, bool> validate,
-    int maxRetries = DefaultMaxRetries, CancellationToken ct = default)
+    int maxRetries = DefaultMaxRetries, Func<int, int> timeout = null, CancellationToken ct = default)
   {
+    timeout ??= NextDelay;
```
millotp, line 341: can you put the default in the function definition instead of null ?

millotp (commented):
looks good !

clients/algoliasearch-client-csharp/algoliasearch/Utils/ClientExtensions.cs
morganleroi: Nope I cannot add a function as a default value

millotp: too bad

millotp (approved):

PR #2712, opened 2024-02-15
fix(swift): QuerySuggestionsConfigurationResponse appID casing

Fluf22:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

- make sure we use the same `appID` casing everywhere


millotp (approved):
nice

generators/src/main/java/com/algolia/codegen/AlgoliaSwiftGenerator.java
```diff
@@ -122,6 +122,7 @@ public void processOpts() {
     );
 
     reservedWords.add("LogLevel");
+    nameMapping.put("appId", "appID");
```
millotp, line 125: does it make sense to have it only for `appID` ? maybe there are a lot more variable in the code with the same format (userID maybe ?)
Maybe there is a more global setting in openapi generator to change the casing for all variables ?

millotp (approved):
looks good ! 1 more e2e that finds lots of bugs !

tests/output/swift/Tests/Utils/Utils.swift
```diff
@@ -35,59 +35,57 @@ public extension Data {
     }
 }
 
-public func XCTLenientAssertEqual(received: Data?, expected: Data?) throws {
+public func XCTLenientAssertEqual(received: Data?, expected: Data?, path: String = "$") throws {
```
millotp, line 38: maybe this could use the `union` function too, it's quite a nice pattern

generators/src/main/java/com/algolia/codegen/AlgoliaSwiftGenerator.java
Fluf22: I think we can adapt case by case, it's easy

tests/output/swift/Tests/Utils/Utils.swift
Fluf22: I'm gonna try again in my next PR

PR #2713, opened 2024-02-15
fix(cts): assert that exceptions are thrown

millotp:
## 🧭 What and Why

When we want to assert an exception, we also need to make sure the catch clause is reached, and this was not the case in Python, PHP and Ruby, so the tests were basically useless.


Fluf22 (approved):
Ah bah oui, indeed

damcou (approved):
Shoot ?

PR #2714, opened 2024-02-15
fix(php): revert restriction for psr/http-message dependency

damcou:
## 🧭 What and Why

🎟 JIRA Ticket: DI-1822

### Changes included:

Following the feedback from the Magento team, we want to rollback the `psr/http-message` dependency by including the 1.1 version to allow them to use the generated PHP client in Magento.


millotp (approved):
looks good

PR #2719, opened 2024-02-16
fix(kotlin): return `replaceAllObjects`'s intermediate operations 

aallam:
## 🧭 What and Why

Simplify the return type to return the list of operations.

### Changes included:

Return the list of intermediate task IDs.


millotp (approved):
should we do this in other languages as well ?

aallam:
> should we do this in other languages as well ?

most likely, yes

PR #2737, opened 2024-02-19
fix(clients): add e2e tests for analytics API

millotp:
## 🧭 What and Why

Simple e2e to make sure the analytics API is called correctly, and that query params encoding is working properly.

Note: the scope of the PR is not very correct but it's just to trigger a release of all languages.


Fluf22 (approved):
🤞🏻

PR #2739, opened 2024-02-19
fix(cts): add e2e for ingestion API

millotp:
## 🧭 What and Why

Add a bit more tests for ingestion API, it can be flaky if someones create a authentication however


shortcuts (approved):
love it!

millotp:
I have very low confidence for this PR

tests/CTS/requests/ingestion/getAuthentications.json
```diff
@@ -27,12 +27,42 @@
...
+            "createdAt": "2024-01-08T16:47:31Z",
+            "updatedAt": "2024-01-08T16:47:31Z"
+          },
+          {},
+          {},
+          {},
+          {},
+          {},
+          {},
+          {}
```
Fluf22, line 63: Is this to only validate the first element thanks to the `union` helper?

Fluf22 (approved):

tests/CTS/requests/ingestion/getAuthentications.json
millotp: exactly, maybe we should make the helper allow for partial match in lists ? like skip remaining elements by looping only over `min(received.length, expected.length)` ?

Fluf22: That's a good idea

Fluf22 (approved):
Amazing job 🙌🏻

PR #2750, opened 2024-02-20
fix(python): remove default values

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

remove default values for python, I've also updated the e2e settings index in order to assert we don't fallback to the default


morganleroi (approved):
Looks 👍🏼

PR #2752, opened 2024-02-20
feat(swift): waitFor helpers

Fluf22:
## 🧭 What and Why

🎟 JIRA Ticket: [DI-1746]

### Changes included:

- `waitForTask` helper
- `waitForApiKey` helper

## 🧪 Test
- manual tests for each helper and their error cases

[DI-1746]: https://algolia.atlassian.net/browse/DI-1746?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/clients/algoliasearch-client-swift/Sources/Core/Helpers/Retry.swift
+++ b/clients/algoliasearch-client-swift/Sources/Core/Helpers/Retry.swift
@@ -1,5 +1,5 @@
 //
-//  Waiters.swift
+//  Retry.swift
 //
 //
 //  Created by Algolia on 19/02/2024.
--- a/config/generation.config.mjs
+++ b/config/generation.config.mjs
@@ -153,5 +153,6 @@ export const patterns = [
   'clients/algoliasearch-client-swift/Sources/Core/Helpers/Version.swift',
 
   'tests/output/swift/Package.swift',
+  '!tests/output/swift/handmade/**',
   '!tests/output/swift/Utils/**',
 ];
--- a/scripts/formatter.ts
+++ b/scripts/formatter.ts
@@ -65,11 +65,7 @@ export async function formatter(language: string, cwd: string): Promise<void> {
       await run('sbt -Dsbt.server.forcestart=true scalafmtAll scalafmtSbt', { cwd, language });
       break;
     case 'swift':
-      if (cwd.includes('tests')) {
-        await run(`swiftformat Tests`, { cwd, language });
-      } else {
-        await run(`swiftformat Sources`, { cwd, language });
-      }
+      await run(`swiftformat .`, { cwd, language });
       break;
     default:
       spinner.warn(`no formatter for '${language}'`);
--- a/templates/swift/api.mustache
+++ b/templates/swift/api.mustache
@@ -149,130 +149,141 @@ typealias Client = {{classname}}
     }{{/vendorExtensions}}
 {{/operation}}{{#isSearchClient}}
 
-    public enum ApiKeyOperation: String {
-        case add
-        case update
-        case delete
-    }
+  public enum ApiKeyOperation: String {
+      case add
+      case update
+      case delete
+  }
 
-    /// Wait for a task to complete
-    /// - parameter taskID: The id of the task to wait for
-    /// - parameter indexName: The name of the index to wait for
-    /// - parameter maxRetries: The maximum number of retries
-    /// - parameter initialDelay: The initial delay between retries
-    /// - parameter maxDelay: The maximum delay between retries
-    /// - returns: Void
-    open func waitForTask(
-        with taskID: Int64,
-        for indexName: String,
-        maxRetries: Int? = 50,
-        initialDelay: TimeInterval? = 0.2,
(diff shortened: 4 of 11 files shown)


clients/algoliasearch-client-swift/Sources/Core/Helpers/Retry.swift
```diff
@@ -0,0 +1,43 @@
...
+    retry: () async throws -> T,
+    until: (T) -> Bool,
+    maxRetries: Int? = 50,
+    initialDelay: TimeInterval? = 0.2,
+    maxDelay: TimeInterval? = 5.0
+) async throws -> T {
+    var currentRetries = 0
+    var currentDelay = initialDelay ?? 0.2
+
+    let maxRetriesValue = maxRetries ?? 50
```
millotp, line 20: why do you need to specify the default value twice ?

config/generation.config.mjs
```diff
@@ -153,5 +153,6 @@ export const patterns = [
   'clients/algoliasearch-client-swift/Sources/Core/Helpers/Version.swift',
 
   'tests/output/swift/Package.swift',
+  '!tests/output/swift/handmade/**',
```
millotp, line 156: handwritten ?

millotp:
thanks for taking the time to add tests !

millotp:
oops wrong button

scripts/formatter.ts
```diff
@@ -65,11 +65,7 @@ export async function formatter(language: string, cwd: string): Promise<void> {
       await run('sbt -Dsbt.server.forcestart=true scalafmtAll scalafmtSbt', { cwd, language });
       break;
     case 'swift':
-      if (cwd.includes('tests')) {
-        await run(`swiftformat Tests`, { cwd, language });
-      } else {
-        await run(`swiftformat Sources`, { cwd, language });
-      }
+      await run(`swiftformat .`, { cwd, language });
```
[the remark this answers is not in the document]
Fluf22: Yes with the rewrite of the config file, it’s cleaner

clients/algoliasearch-client-swift/Sources/Core/Helpers/Retry.swift
Fluf22: My bad
I had to do this because the user could force pass nil
But I shouldn’t allow this since we’ll default anyway
I’ll fix this

clients/algoliasearch-client-swift/Sources/Core/Helpers/Retry.swift
```diff
@@ -0,0 +1,43 @@
...
+//
+
+import Foundation
+
+public func retryUntil<T>(
+    retry: () async throws -> T,
+    until: (T) -> Bool,
+    maxRetries: Int? = 50,
+    initialDelay: TimeInterval? = 0.2,
+    maxDelay: TimeInterval? = 5.0
```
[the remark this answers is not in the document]
Fluf22: I followed the Go example
Would it be better with a function?
I find it cleaner with variables in swift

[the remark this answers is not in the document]
Fluf22: Updated to match `createIterable` method signature used in other clients

templates/swift/api.mustache
```diff
@@ -147,6 +147,182 @@ typealias Client = {{classname}}
...
+                        return false
+                    }
+
+                    if apiKey.maxQueriesPerIPPerHour != response.maxQueriesPerIPPerHour {
+                        return false
+                    }
+
+                    if apiKey.validity != response.validity {
+                        return false
+                    }
```
shortcuts, line 248: I just realized we compare validity but it's meant to be decreased every seconds no? idk how that behave in this compare scenario

millotp: yeah it's an issue in every language, but it was already the same in legacy clients https://github.com/algolia/api-clients-automation/pull/2715#discussion_r1494217048

Fluf22: I thought we could do something like `receivedValidity > expectedValidity + maxDelay`?
Wdyt?

Fluf22: Yes totally, I forgot to remove it, but @millotp mentioned it to me already

millotp: except if the guy only update the validity

shortcuts: > I thought we could do something like `receivedValidity > expectedValidity + maxDelay`? Wdyt?

time might have elapsed in the meantime still :/

> except if the guy only update the validity

I guess this is the kind of edge case we can just mitigate with a comment

Fluf22: > time might have elapsed in the meantime still :/

If time elapsed more than `maxDelay`, then the operation will fail anyway

Fluf22: Let's go with this and check back when we have an actual issue from a customer.

shortcuts (approved):
gg gg

PR #2757, opened 2024-02-21
fix(clients): browse synonyms page not in response

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket:

### Changes included:

Catched by fluffman the page parameter doesn't exist in the synonyms response


Fluf22 (approved):
Let’s go

PR #2753, opened 2024-02-20
feat(python): add `generate_secured_api_key` and `get_secured_api_key_remaining_validity`

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-1774

### Changes included:

PR #2760, opened 2024-02-22
fix(cts): remove flaky monitoring tests

millotp:
## 🧭 What and Why

The monitoring API seems to be very slow for us, and return a lot of Connection timeout which makes the CTS flaky, the easiest solution is to remove the test.


shortcuts:
It's already removed in my PR :D

shortcuts:
https://github.com/algolia/api-clients-automation/pull/2753

Just waiting for your approval

PR #2762, opened 2024-02-22
fix(clients): replace all objects

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

fix wait for python and chsarp, kotlin is already right, php is less wrong


millotp (approved):
I hope this is the correct arrangement !

shortcuts:
> I hope this is the correct arrangement !

so from the tests I've made................. weirdly it doesn't matter much in that case but I guess it's because it's pretty fast still

however looking at the current JS implementation, this is the correct way

morganleroi:
image

🤣

morganleroi (approved):
Thanks A LOT

PR #2763, opened 2024-02-22
feat(python): add `chunked_batch`

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

I realized while implementing the `replace_all_objects` method that we were calling the native `batch` method, which ultimately supports a size of 1000 records (you can provide more if they are smaller, but 1000 is what was done in old clients)

so in this PR we add a new `chunked_batch` method to leverage it there


templates/python/search_helpers.mustache
```diff
@@ -207,7 +207,38 @@
...
+                        batch_write_params=BatchWriteParams(requests=requests),
+                        request_options=request_options,
+                    )
+                )
+                requests = []
+        if wait_for_tasks:
+            for response in responses:
+                await self.wait_for_task(
+                    index_name=index_name, task_id=response.task_id
+                )
```
Fluf22, line 238: Shouldn't we wait only for the last taskID?
In which case, we can drop the `wait_for_task` bool and just return the response so the caller can wait for it if he wants to

Fluf22 (approved):
🆗

PR #2769, opened 2024-02-23
feat(python): add `batch_size` to chunked batch

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-1837

### Changes included:

- [x] add `batch_size` to allow user controlling the split chunk logic
- [x] fix some formatting and naming
- [x] add tests for `replace_all_records`
- [x] match return type to JS and Swift implementations


templates/python/search_helpers.mustache
```diff
@@ -213,6 +213,7 @@
         objects: List[Dict[str, Any]],
         action: Action = "addObject",
         wait_for_tasks: bool = False,
+        batch_size: int = 1000,
```
millotp, line 216: should you add a check on this ? it cannot go above 1000 right ? but maybe the API already returns an error

millotp (approved):
nice !

templates/python/search_helpers.mustache
shortcuts: it can go above if the records are pretty small, 1000 is a safe limit that the previous API clients team (or maybe search directly?) added based on the average allowed record size

PR #2788, opened 2024-02-26
fix(javascript): allow `batchSize` on `ReplaceAllObjectsOptions`

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

the automerge merged before I fix, this option should be available on both method not just the chunked one


Fluf22 (approved):
I'll allow it

PR #2789, opened 2024-02-26
chore: run all gen on main

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

CI failed multiple times on `main` but the last commit merged will only trigger the JavaScript client to run. We should make `main` run every steps of the gen so that we know it's always the most up to date version.


millotp (approved):
nice

millotp (approved):
let's go !

shortcuts:
lovely

PR #2792, opened 2024-02-27
fix(specs): add missing attributes for CT Source update

damcou:
## 🧭 What and Why

🎟 JIRA Ticket: DI-1788

### Changes included:

- Add missing attributes `url` and `fallbackIsInStockValue` for CT Source Update

### DIFF
--- a/specs/ingestion/common/schemas/source.yml
+++ b/specs/ingestion/common/schemas/source.yml
@@ -373,6 +373,13 @@ SourceUpdateCommercetools:
         Array of locales that must match the following pattern: ^[a-z]{2}(-[A-Z]{2})?$. For example ["fr-FR", "en"].
       items:
         type: string
+    url:
+      type: string
+    fallbackIsInStockValue:
+      type: boolean
+      default: true
+      description: >
+        Determines the value that will be stored in the Algolia record if there's no inventory information on the product.
     customFields:
       $ref: '#/CommercetoolsCustomFields'
 
--- a/specs/ingestion/common/schemas/source.yml
+++ b/specs/ingestion/common/schemas/source.yml
@@ -381,8 +381,6 @@ SourceUpdateCommercetools:
         Determines the value that will be stored in the Algolia record if there's no inventory information on the product.
     customFields:
       $ref: '#/CommercetoolsCustomFields'
-  required:
-    - url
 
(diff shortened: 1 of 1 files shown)


specs/ingestion/common/schemas/source.yml
```diff
@@ -373,6 +373,13 @@ SourceUpdateCommercetools:
         Array of locales that must match the following pattern: ^[a-z]{2}(-[A-Z]{2})?$. For example ["fr-FR", "en"].
       items:
         type: string
+    url:
+      type: string
+    fallbackIsInStockValue:
+      type: boolean
+      default: true
+      description: >
+        Determines the value that will be stored in the Algolia record if there's no inventory information on the product.
```
Fluf22, line 382: This means during an update, it will default to true...
Are you sure this is the wanted behavior?

Fluf22: Then you should add it to the required array

damcou: done

specs/ingestion/common/schemas/source.yml
```diff
@@ -373,8 +373,16 @@ SourceUpdateCommercetools:
...
         type: string
+    url:
+      type: string
+    fallbackIsInStockValue:
+      type: boolean
+      description: >
+        Determines the value that will be stored in the Algolia record if there's no inventory information on the product.
     customFields:
       $ref: '#/CommercetoolsCustomFields'
+  required:
```
millotp, line 384: the goal of the update is not to have required properties

damcou: Removed, please, I just want to close my PR about the custom fields...

specs/ingestion/common/schemas/source.yml
damcou: (and removed, see below)

shortcuts: > This means during an update, it will default to true... Are you sure this is the wanted behavior?

We don't (and shouldn't) use default values in the API client, if the API defaults to true it's normal to mention it here

shortcuts: ah it's an update nvm

shortcuts (approved):
wonderful

PR #2794, opened 2024-02-27
chore(specs): remove JSON specs and new worlds logic

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

we sadly don't need this logic anymore so we can remove anything related to those specs


millotp (approved):
😢

PR #2796, opened 2024-02-27
feat(php): add `chunkedBatch` method

damcou:
## 🧭 What and Why

🎟 JIRA Ticket: DI-1832

### Changes included:

- Add method `chunkedBatch` and use it in `replaceAllObjects`


morganleroi (commented):
Can you add a unit test to check that we do call X times method and correctly chunk the batch ?

shortcuts:
> > Can you add a unit test to check that we do call X times method and correctly chunk the batch ?
> 
> Where are we regarding helper tests ? I see that only a few languages have them, shouldn't we wait for the helpers to be added to the specs before adding them ?

you can follow python and javascript they should be fairly close to reality, errors will be fixed in an upcoming pr from @millotp

morganleroi:
Seems weird to me to push code without associated tests. Even if helpers signature will be generated, you will have to implement the test. Up to you.

damcou:
Since it's not only about this method, I'll dedicate a whole ticket/PR about helper tests (it may take a while)

morganleroi (approved):

PR #2798, opened 2024-02-28
feat(cts): generate tests for helpers

millotp:
## 🧭 What and Why

Use the helper spec to generate tests, starting with the simplest: generateSecuredApiKey


tests/CTS/client/search/helpers.json
```diff
@@ -0,0 +1,25 @@
+[
+  {
+    "testName": "generate secured api key basic",
+    "steps": [
+      {
+        "type": "method",
+        "object": "$client",
+        "path": "generateSecuredApiKey",
+        "parameters": {
+          "parentAPIKey": "2640659426d5107b6e47d75db9cbaef8",
```
millotp, line 10: no worries this is our search api key, this means the generate key is usable for e2e

damcou:
I was about to start adding those test for PHP by hand, is it still useful or should I wait for this PR to be merged ?

clients/algoliasearch-client-javascript/packages/client-common/src/transporter/helpers.ts
```diff
@@ -42,16 +42,14 @@ export function serializeUrl(
...
   return Object.keys(parameters)
+    .filter((key) => parameters[key] !== undefined)
+    .sort()
     .map(
       (key) =>
         `${key}=${encodeURIComponent(
-          isObjectOrArray(parameters[key])
-            ? JSON.stringify(parameters[key])
+          Object.prototype.toString.call(parameters[key]) === '[object Array]'
+            ? parameters[key].join(',')
```
shortcuts, line 52: I feel like this will break something

clients/algoliasearch-client-python/algoliasearch/http/serializer.py
```diff
@@ -12,21 +13,30 @@ class QueryParametersSerializer:
     query_parameters: Dict[str, Any] = {}
 
     def parse(self, value) -> Any:
-        if isinstance(value, dict):
-            return dumps(value)
-        elif isinstance(value, list):
+        if isinstance(value, list):
```
shortcuts, line 16: does list validates dict? that's why you changed the order?

templates/go/api.mustache
```diff
@@ -19,6 +19,7 @@ import (
 	"crypto/sha256"
 	"encoding/base64"
 	"encoding/hex"
+  "sort"
```
shortcuts, line 22: 📏

shortcuts (commented):
looks awesome

clients/algoliasearch-client-javascript/packages/client-common/src/transporter/helpers.ts
millotp: it did break something, it wasn't working before with generateSecuredApiKey, but you agree with me that never ever do we have a query param that looks like a serialezed object right ? we only allow primitives and array to stringified

shortcuts: Yup but IIRC this implem was exactly the same as the legacy client but since I never found the reason why it had an object case :/

millotp: okay but we don't have any tests that expect that, and we don't have the same behavior in other languages, this was an outlier

clients/algoliasearch-client-python/algoliasearch/http/serializer.py
millotp: same thing as for js, we never want to stringify a dict

clients/algoliasearch-client-javascript/packages/client-common/src/transporter/helpers.ts
millotp: I can revert this if you want, but I think this is more correct

shortcuts: I also think so but since it worked like that for years I think there's an edge case we are not aware of :/ we can leave it and see if there's complains

millotp: reverted in https://github.com/algolia/api-clients-automation/pull/2798/commits/b9ff92cb96993653281f769543a87f931e192ded

millotp: ah bah it breaks the test

shortcuts: I mostly wonder how it could've been wrong for 4 years in the current JS client, maybe the engine supports this format?

Your initial solution is indeed correct to me, we can go with it and if user complain we fix it (we can also for example try it on the Crawler migration PoC too)

millotp: sounds good, the PR should be fine then

shortcuts (dismissed):
🤞🏻 🍀

PR #2799, opened 2024-02-28
docs: integrates generated snippet

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-1848

### Changes included:

This PR aims at integrating the generated snippets in the documentation, which is [natively supported by Redoc](https://github.com/Redocly/redoc/blob/main/docs/redoc-vendor-extensions.md#x-codeSamples).


templates/csharp/snippets/method.mustache
shortcuts, line 1: I guess all of those separators could be directly added by the generator but having them  manually defined allows us to properly encapsulate the snippet part we want to showcase

scripts/types.ts
```diff
@@ -66,7 +66,42 @@ type Tag = {
...
 };
 
+type Method = 'delete' | 'get' | 'options' | 'patch' | 'post' | 'put';
+
+export type CodeSamples = {
+  lang:
+    | 'C'
+    | 'C++'
+    | 'CoffeeScript'
+    | 'csharp'
```
shortcuts, line 76: `c#` and `C#` notations doesn't seem properly forwarded from Redoc -> Linguist -> Prismjs, using `csharp` shows a correct highlighting though

scripts/buildSpecs.ts
```diff
@@ -3,12 +3,86 @@ import fsp from 'fs/promises';
...
+      return 'csharp';
+    case 'javascript':
+      return 'JavaScript';
+    default:
+      return capitalize(language) as CodeSamples['lang'];
+  }
+}
+
+// For a given `clientName`, reads the matching snippet file for every available clients and builds an hashmap of snippets per operationId per language.
+async function transformSnippetsToCodeSamples(
```
shortcuts, line 24: this is kind of a mess which I'd like to improve but does the job for the POC

scripts/types.ts
morganleroi: Thanks !

shortcuts (commented):
another question is: do we want to keep the standalone snippets if they are integrated in the docs? do you see value at having those files in the repo?

scripts/common.ts
```diff
@@ -289,7 +289,7 @@ export async function setupAndGen(
 ): Promise<void> {
   if (!CI) {
     const clients = [...new Set(generators.map((gen) => gen.client))];
-    await buildSpecs(clients, 'yml', true);
+    await buildSpecs(clients, 'yml', false, true);
```
millotp, line 292: might be worth to use an object as the parameter now, it's not obvious what `false, true` mean in this case

scripts/cli/index.ts
```diff
@@ -40,6 +40,10 @@ const flags = {
     flag: '-json, --output-json',
     description: 'outputs the spec in JSON instead of yml',
   },
+  forDocs: {
```
millotp, line 43: just `--doc` is enough ?

scripts/buildSpecs.ts
millotp, line 1: this file is getting huge, you think it would be possible to split it between doc and normal stuff ?

scripts/buildSpecs.ts
```diff
@@ -3,12 +3,88 @@ import fsp from 'fs/promises';
...
+  clientName: string,
+  forDocs: boolean,
+): Promise<Record<string, Record<string, string>>> {
+  const snippetFiles: Record<string, Record<string, string>> = {};
```
millotp, line 30: is this the same type as `Path` ?

scripts/buildSpecs.ts
```diff
@@ -3,12 +3,88 @@ import fsp from 'fs/promises';
...
+
+    // find snippets for each operationId in the current gen.language + clientName combo
+    const snippetFileContent = await fsp.readFile(
+      toAbsolutePath(
+        `snippets/${gen.language}/${gen.snippets.outputFolder}/${createClientName(clientName, gen.language)}${gen.snippets.extension}`,
+      ),
+      'utf8',
+    );
+    // iterate over every matches (operationId) and store it in the hashmap for later use
+    for (const match of snippetFileContent.matchAll(/SEPARATOR (.+)\n([\s\S]*?)SEPARATOR/g)) {
```
millotp, line 49: can you include a weird char in the separator pls just in case this words appear in a test ?

scripts/buildSpecs.ts
```diff
@@ -3,12 +3,88 @@ import fsp from 'fs/promises';
...
+      const lines: string[] = match[0].split('\n');
+      if (!lines?.[0]) {
+        throw new Error(`No snippet found for ${gen.language} ${gen.client}`);
+      }
+
+      const operationIdMatch = lines[0].match(/SEPARATOR (.+)/);
```
millotp, line 55: you could also get this from `match[1]`

scripts/buildSpecs.ts
```diff
@@ -3,12 +3,88 @@ import fsp from 'fs/promises';
...
+      }
+
+      snippetFiles[gen.language][operationIdMatch[1]] = '';
+
+      let indent = 0;
+      for (const [i, line] of lines.entries()) {
+        // skip first and last lines because they contain the SEPARATOR or operationId
+        if (i === 0 || i === lines.length - 1) {
+          continue;
+        }
```
millotp, line 71:
```suggestion
      for (const line of lines.slice(1, -1)) {
```

scripts/buildSpecs.ts
```diff
@@ -3,12 +3,88 @@ import fsp from 'fs/promises';
...
+
+        // best effort to determine how far the snippet is indented so we
+        // can have every snippets in the documentation on the far left
+        // without impacting the formatting
+        if (indent === 0) {
+          indent = line.length - line.trim().length;
+        }
+
+        snippetFiles[gen.language][operationIdMatch[1]] += `${line.slice(indent)}\n`;
```
millotp, line 80: maybe you can replace tabs by 2 spaces also, I think go is broken

scripts/buildSpecs.ts
```diff
@@ -33,43 +109,55 @@ async function transformBundle({
   }
 
   const bundledSpec = yaml.load(await fsp.readFile(bundledPath, 'utf8')) as Spec;
-
-  let bundledDocSpec: Spec | undefined;
-  if (withDoc) {
-    bundledDocSpec = yaml.load(await fsp.readFile(bundledPath, 'utf8')) as Spec;
-  }
   const tagsDefinitions = bundledSpec.tags;
+  const snippetFiles = await transformSnippetsToCodeSamples(clientName, forDocs);
```
millotp, line 113:
```suggestion
  const snippetSamples = await transformSnippetsToCodeSamples(clientName, forDocs);
```
?

millotp (commented):
this is insanely good congrats !

millotp:
can you update the apic autocompletion too ? I can do it if you want

website/redocly.yaml
```diff
@@ -0,0 +1,14 @@
...
+  - recommended
+
+theme:
+  # See https://redocly.com/docs/api-reference-docs/configuration/functionality/
+  openapi:
+    disableSearch: true
+    downloadDefinitionUrl: https://github.com/algolia/api-clients-automation/tree/main/specs/bundled
+    hideRequestPayloadSample: true
+    # See https://redocly.com/docs/api-reference-docs/configuration/theming/
+    theme:
```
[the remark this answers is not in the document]
shortcuts: I was trying options and nothing seems to be necessary right now, I can remove the prop if you want

scripts/cli/index.ts
shortcuts: I don't mind!

scripts/buildSpecs.ts
shortcuts: I can better type it actually but no it's for languages instead of methods!

scripts/buildSpecs.ts
shortcuts: Sure I can at least extract the docs logic

website/redocly.yaml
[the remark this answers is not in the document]
millotp: no biggy

scripts/buildSpecs.ts
shortcuts: I'll do the cleanup in a second time because I believe we have to move a few things around which would add a lot of complexity to knowing what really changed in this PR

shortcuts:
> can you update the apic autocompletion too ? I can do it if you want

it doesn't include flags so there's nothing to update, no?

millotp:
nope I didn't try with flags

templates/javascript/snippets/method.mustache
```diff
@@ -17,11 +18,12 @@ export async function snippetFor{{method}}(): Promise<void> {
     {{#requestOptions.headers.parameters}}headers: {{{.}}}{{/requestOptions.headers.parameters}}
   };{{/hasRequestOptions}}
 
-  // Call the API
+  // call the API
```
millotp, line 21: this is the only snippets using `call` in lowercase 😛

shortcuts: Good catch!

millotp (approved):
so nice !

millotp:
you got a cache issue, bumping should do it

shortcuts:
> you got a cache issue, bumping should do it

Ah yes indeed, I also have an addition for this afternoon I'll do both later

millotp (approved):
the acls are not very pretty but the snippets are 💯

PR #2800, opened 2024-02-28
fix(swift): improve DX

Fluf22:
## 🧭 What and Why

🎟 JIRA Ticket: [DI-1724]

### Changes included:

- remove GzipSwift from deps
- make CocoaPods compatible with the project
- use custom gzipping
- use OSLog package
- only import swift-log for linux

## 🧪 Test

[DI-1724]: https://algolia.atlassian.net/browse/DI-1724?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clients/algoliasearch-client-swift/Sources/Core/Networking/Compression/Gzip.swift
shortcuts, line 1: Not saying the implem is bad but isn't it worth letting the user know using gzip requires an external dep so that we don't have to maintain this kind of implem?

shortcuts (commented):
Lovely cleanup

clients/algoliasearch-client-swift/Sources/Core/Networking/Compression/Gzip.swift
Fluf22: The only implem that supports Linux makes us unable to support cocoapods
This is an attempt at supporting it, and seeing what are the downsides
First one is obv to maintain gzipping + Linux compat ourselves,  but on the other hand, it's a known open source algorithm a lot of people explained everywhere on the internet, and which shouldn't change much

Fluf22: Until this, I was using an external dep that worked well (GzipSwift), but it breaks compat with CocoaPods because it targets a value for some platform that's too low and not supported by Xcode anymore (which is used behind the scene to build pods by the cocoapods tooling suite)

clients/algoliasearch-client-swift/Sources/Core/Common/Logging.swift
```diff
@@ -2,26 +2,32 @@
 //  Logging.swift
 //
 //
-//  Created by Algolia on 15/12/2023.
+//  Created by Algolia on 20/02/2020.
```
millotp, line 5: whoah you have a time machine !

clients/algoliasearch-client-swift/Sources/zlib/zlib.h
millotp, line 1: let try to make C the most used lang in this repo !

millotp (approved):
meant to approve

clients/algoliasearch-client-swift/Sources/Core/Networking/Compression/Gzip.swift
Fluf22: Still trying to make cocoapods working, and not looking good... I'll put the PR back to draft and ping you when it'll be good (if it ever is)

clients/algoliasearch-client-swift/Sources/Core/Common/Logging.swift
Fluf22: Copied back the update made by Vlad on master 😄

clients/algoliasearch-client-swift/Sources/Core/AnyCodable/AnyCodable.swift
```diff
@@ -0,0 +1,149 @@
...
+    public let value: Any
+
+    public init(_ value: (some Any)?) {
+        self.value = value ?? ()
+    }
+}
+
+extension AnyCodable: _AnyEncodable, _AnyDecodable {}
+
+extension AnyCodable: Equatable {
```
millotp, line 28: worst language ever

clients/algoliasearch-client-swift/Sources/Core/AnyCodable/AnyEncodable.swift
```diff
@@ -0,0 +1,299 @@
...
+        default:
+            let context = EncodingError.Context(
+                codingPath: container.codingPath,
+                debugDescription: "AnyEncodable value cannot be encoded"
+            )
+            throw EncodingError.invalidValue(value, context)
+        }
+    }
+
+    #if canImport(Foundation)
```
millotp, line 114: you need to recode the entire std library ?

generators/src/main/java/com/algolia/codegen/AlgoliaSwiftGenerator.java
```diff
@@ -29,6 +32,78 @@
 
 public class AlgoliaSwiftGenerator extends Swift5ClientCodegen {
 
+  private static final List<String> reservedModelNames = List.of(
```
millotp, line 35: could be nice to get this list from the spec, but very hard

generators/src/main/java/com/algolia/codegen/AlgoliaSwiftGenerator.java
```diff
@@ -42,6 +117,40 @@ public static String formatIdentifier(String text) {
     return INSTANCE.isReservedWord(text) ? INSTANCE.escapeReservedWord(text) : text;
   }
 
+  public static Boolean isReservedModelName(String name) {
+    return (
+      AlgoliaSwiftGenerator.reservedModelNames.contains(name) || AlgoliaSwiftGenerator.reservedModelNames.contains(Helpers.capitalize(name))
```
millotp, line 122: or just store the word in lowercase and always compare with lowecase

generators/src/main/java/com/algolia/codegen/AlgoliaSwiftGenerator.java
```diff
@@ -336,4 +454,25 @@ public String toEnumVarName(String value, String datatype) {
       LOWERCASE_FIRST_LETTER
     );
   }
+
+  @Override
+  public String toModelName(String name) {
+    var sanitizedName = this.sanitizeName(name);
+    var camelizedName = camelize(sanitizedName);
+    if (AlgoliaSwiftGenerator.isReservedModelName(camelizedName)) {
```
millotp, line 462: I don't think you need to have `AlgoliaSwiftGenerator.` everytime.

generators/src/main/java/com/algolia/codegen/cts/tests/ParametersWithDataType.java
```diff
@@ -132,6 +135,10 @@ private Map<String, Object> traverseParams(
       finalParamName = "type_";
     }
 
+    if (language.equals("swift")) {
```
millotp, line 138: can you make a function instead of adding if everywhere ?

Fluf22: There are two different things done in those `if`: do you want to have two different method called?
Also, don't you find it clearer to have a transparent if condition using a specific language name like this, instead of having a function call that does conditional logic, some other place in the file?

clients/algoliasearch-client-swift/Sources/Core/AnyCodable/AnyEncodable.swift
Fluf22: I'm not sure to understand the question?
`canImport` checks that the lib exists on the system. In that case, it adds an overload to the encode method.
In the other case, it does nothing

generators/src/main/java/com/algolia/codegen/AlgoliaSwiftGenerator.java
Fluf22: Okay

clients/algoliasearch-client-swift/Sources/Core/AnyCodable/AnyEncodable.swift
millotp: I meant the comment for the entire file, the code looks like something the standard lib should offer

generators/src/main/java/com/algolia/codegen/cts/tests/ParametersWithDataType.java
millotp: I find it clearer with less if, the code is already very complex and it would be nice to make it more streamlined, you can create 2 functions yes, as you want.
We should not have language or client specific code inside the main functions like this

clients/algoliasearch-client-swift/Sources/Core/AnyCodable/AnyEncodable.swift
Fluf22: No, it's not in the standard library 😄

generators/src/main/java/com/algolia/codegen/cts/tests/ParametersWithDataType.java
Fluf22: Okay

generators/src/main/java/com/algolia/codegen/cts/tests/ParametersWithDataType.java
```diff
@@ -297,6 +301,11 @@ private void handleModel(
       // find a discriminator to handle oneOf
       CodegenModel model = (CodegenModel) spec;
       IJsonSchemaValidationProperties match = findMatchingOneOf(param, model);
+
+      if (language.equals("swift")) {
```
millotp, line 305: you could do the same with this, maybe not as nice

millotp (dismissed):
gg !

generators/src/main/java/com/algolia/codegen/cts/tests/ParametersWithDataType.java
Fluf22: Ah yes, I wanted to and I forgot...
Sorry, let me do it

generators/src/main/java/com/algolia/codegen/cts/tests/ParametersWithDataType.java
```diff
@@ -326,6 +333,11 @@ private void handleModel(
         useExplicitName = Helpers.shouldUseExplicitOneOfName(model.oneOf);
       }
 
+      if (language.equals("swift")) {
+        typeName = AlgoliaSwiftGenerator.prefixReservedModelName(typeName, client);
+        baseType = AlgoliaSwiftGenerator.prefixReservedModelName(baseType, client);
+      }
```
millotp, line 339: can you use `getOriginalParamName` here too ?

millotp (approved):
looks good !

PR #2801, opened 2024-02-29
fix(search): add `queryID` to search response

aallam:
## 🧭 What and Why

`queryID` is missing from search responses.
https://www.algolia.com/doc/guides/sending-events/concepts/event-types/#what-is-the-query-id

### Changes included:

Add `queryID` to `SearchResponse`

### DIFF
--- a/specs/search/common/schemas/SearchResponse.yml
+++ b/specs/search/common/schemas/SearchResponse.yml
@@ -174,8 +174,12 @@ baseSearchResponse:
       type: string
       description: Host name of the server that processed the request.
       example: 'c2-uk-3.algolia.net'
-    userData:
+    userData: 
       $ref: '../../../common/schemas/IndexSettings.yml#/userData'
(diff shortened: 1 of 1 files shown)


specs/search/common/schemas/SearchResponse.yml
```diff
@@ -174,8 +174,12 @@ baseSearchResponse:
       type: string
       description: Host name of the server that processed the request.
       example: 'c2-uk-3.algolia.net'
-    userData:
+    userData: 
       $ref: '../../../common/schemas/IndexSettings.yml#/userData'
+    queryID:
```
millotp, line 179: is it always returned ?

aallam: No, it's returned only if the request had `clickAnalytics: true`

millotp (approved):
perfect !

PR #2804, opened 2024-02-29
fix(cts): test encoding of special characters

millotp:
## 🧭 What and Why

The `filters` property can use some special char, we need to make sure that every language has the same encoding

### DIFF
--- a/tests/CTS/requests/common/customGet.json
+++ b/tests/CTS/requests/common/customGet.json
@@ -38,7 +38,8 @@
         "query": "parameters with space",
         "and an array": [
           "array",
-          "with spaces"
+          "with spaces",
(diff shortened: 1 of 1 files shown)


shortcuts (changes requested):
good idea! just blocking in case of miss merge during release

millotp:
the CI will fail don't worry

Fluf22:
Swift being green feelsgoodman.gif

millotp:
I give up, algolia accept both so I'll just adapt my tests

shortcuts:
> I give up, algolia accept both so I'll just adapt my tests

oneof cts oneof cts!!!!!!!!!!!!!!!!

PR #2807, opened 2024-03-01
fix(javascript): move node related helpers at node bundle level

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

because crypto is not available on the browser, and it doesn't make sense to expose it there, we move those helpers to the `node` bundle only which makes the mustache changes a big heavy but for the best!


generators/src/main/java/com/algolia/codegen/AlgoliaJavascriptGenerator.java
```diff
@@ -79,6 +79,7 @@ public void processOpts() {
       // `lite` builds
       supportingFiles.add(new SupportingFile("client/builds/browser.mustache", "lite/builds", "browser.ts"));
       supportingFiles.add(new SupportingFile("client/builds/node.mustache", "lite/builds", "node.ts"));
+      supportingFiles.add(new SupportingFile("client/builds/liteNode.mustache", "lite/builds", "node.ts"));
```
shortcuts, line 82: I need to do more partial mustache here as it's basically the same file without the helper mention

millotp (approved):

Fluf22 (approved):
Nice, let's go

PR #2824, opened 2024-03-04
fix(scripts): lint the correct bundled spec

millotp:
## 🧭 What and Why

The normal spec was linted when building the doc spec, resulting in absolute chaos


morganleroi (approved):
Nice !

PR #2826, opened 2024-03-04
chore: remove custom caching logic

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

cache flakyness already fooled us by merging bugs on main a few times, for the few seconds we gain, we should remove it completely to increase our confidence


scripts/cache.ts
shortcuts, line 1: I just realized I've removed the local one too, the time increase isn't that bad since it's only for bundling java or the specs, but lmk if you want me to restore it

shortcuts: reverted

scripts/common.ts
```diff
@@ -157,25 +156,9 @@ export async function gitCommit({
 
 async function buildCustomGenerators(): Promise<void> {
   const spinner = createSpinner('building custom generators');
-  const cache = new Cache({
```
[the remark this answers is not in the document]
shortcuts: ah I forgot those

millotp (approved):
looks good ! I think its a good trade off

shortcuts:
> looks good ! I think its a good trade off

yup at least we shouldn't be afraid of having invalid cache anymore

PR #2831, opened 2024-03-05
feat(specs): improve API reference documentation

kai687:
## 🧭 What and Why

To make the OpenAPI-generated docs ready to replace the main API reference docs,
add more information from the API methods and API parameters docs.

🎟 JIRA Ticket: 

### Changes included:

- Update the Search API's description. Especially parameters from the OLD API parameter reference.
- Update redocly config for some better sorting (required before optional parameters, alphabetical sorting)
- **Fix**: Search dictionaries endpoint has `searchDictionaryEntries` response
- **Fix**: `attributesForFaceting` can only be used as index setting, not as search parameter.
- **Fix**: the `explain` API parameter is not supposed to be exposed.


millotp (approved):
thanks for the hard work, this is huge !

shortcuts (commented):
We might want additionalProperties: true if explain is not exposed because IIRC Kotlin throw on unknown props, @aallam ?

shortcuts:
This is a massiveeeee improvement thanks a lot @kai687

specs/common/parameters.yml
```diff
@@ -30,33 +30,34 @@ Index:
...
   name: forwardToReplicas
-  description: Indicates whether changed index settings are forwarded to the replica indices.
+  description: Whether changes are applied to replica indices.
   schema:
     type: boolean
 
 Page:
   in: query
   name: page
   description: >
```
shortcuts, line 58: I think we can replace all > with pipes to preserve formatting, wdyt?

specs/search/paths/dictionaries/common/schemas/SearchDictionaryEntriesResponse.yml
```diff
@@ -24,43 +29,28 @@ dictionaryEntry:
...
+        - cheval
+        - cheveaux
       items:
         type: string
     decomposition:
       type: array
-      description: For compound entries, governs the behavior of the `word` parameter.
-      example: ['kopf','schmerz','tablette']
+      description: Invividual components of a compound word in the `compounds` dictionary.
+      example: [kopf, schmerz, tablette]
```
shortcuts, line 53: We should keep quotes since it's an array of string, no?

shortcuts (commented):
HUGE

millotp (approved):
thanks !

shortcuts (approved):
love it!

PR #2832, opened 2024-03-06
fix(scripts): use glob to apply wildcard before rm

Fluf22:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

- use glob to apply wildcard and get a list of files to remove
- fix error message in `check.yml`

## 🧪 Test
Tested locally by following this path:
- in `AlgolaSwiftGenerator`, change the destination file name for the `client_configuration.mustache` template to `ConfigurationClient.swift`
- in `removeExistingCodegen.ts`, change the wildcard to `*Client.swift`
- run `apicb generate swift attesting -v`
- confirm that the script shows it will remove only the `AbtestingClient.swift` file and the `Models` folder
- confirm that the old `Configuration.swift` file is still there
- fix everything
- ???
- profit!


scripts/package.json
```diff
@@ -45,6 +45,7 @@
     "express": "4.18.3",
     "folder-hash": "4.0.4",
     "fs-extra": "11.2.0",
+    "glob": "^10.3.10",
```
millotp, line 48: can you pin this pls ?

scripts/pre-gen/removeExistingCodegen.ts
```diff
@@ -79,15 +81,18 @@ export async function removeExistingCodegen({
...
   }
 
-  // Delete client model folder/file
-  await fsp.rm(toAbsolutePath(path.resolve('..', output, baseModelFolder, clientModel)), {
-    force: true,
-    recursive: true,
-  });
+  // Delete client model and API folder/files
+  const filesToRemove = await glob([
+    toAbsolutePath(path.resolve('..', output, baseModelFolder, clientModel)),
```
millotp, line 86: where does the `..` comes from ?

millotp (commented):
nice !

scripts/package.json
Fluf22: Removed

scripts/pre-gen/removeExistingCodegen.ts
Fluf22: It was from the old code
Since we used `fsp`, we were in the `scripts` folder and had to exit it, first

Anyway, it has been removed

millotp: that's weird because toAbsolutePath already reset the cwd to the root folder

Fluf22: Ah 🤷🏻‍♂️

millotp (approved):
gg !

PR #2833, opened 2024-03-07
chore(scala): fix scala release ci

aallam:
## 🧭 What and Why

At Scala client release time, gpg key is not used correctly.

### Changes included:

- Attempt to fix gpg signing 
- Add logging (list installed pgp keys)


Fluf22 (approved):
Tricky issue

PR #2856, opened 2024-03-11
fix(swift): wait for tasks while sending chunk batches

Fluf22:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

- fix `replaceAllObjects` helper by waiting for tasks in the `chunkedBatch` method


millotp (approved):
Thanks for the fix !

Fluf22:
Thank you for spotting the bug!

PR #2858, opened 2024-03-12
feat(specs): add ga4BigqueryExport, streaming and append options to task inputs

JasonBerry:
## 🧭 What and Why
GA4 BigQuery Export task type was added.

🎟 JIRA Ticket: EEX-1055

### Changes included:

- Add `ga4BigqueryExport` task type
- Add `streaming` trigger type
- Add `append` action type

### DIFF
--- a/specs/ingestion/common/schemas/source.yml
+++ b/specs/ingestion/common/schemas/source.yml
@@ -94,7 +94,7 @@ SourceSearch:
 
 SourceType:
   type: string
-  enum: ['bigcommerce', 'commercetools', 'json', 'csv', 'bigquery', 'docker']
+  enum: ['bigcommerce', 'commercetools', 'json', 'csv', 'bigquery', 'docker', 'ga4BigqueryExport']
 
 SourceCommercetools:
   type: object
@@ -261,6 +261,28 @@ SourceCSV:
   required:
     - url
 
+SourceGA4BigQueryExport:
+  type: object
+  additionalProperties: false
+  properties:
+    projectID:
+      type: string
+      description: GCP project ID that the BigQuery Export writes to.
+    datasetID:
+      type: string
+      description: BigQuery dataset ID that the BigQuery Export writes to.
+    tablePrefix:
+      type: string
+      description: Prefix of the tables that the BigQuery Export writes to (i.e. events_intraday_ for streaming, events_ for daily).
+  required:
+    - projectID
+    - datasetID
+    - tablePrefix
+  x-discriminator-fields:
+    - projectID
+    - datasetID
+    - tablePrefix
+
 SourceBigQuery:
   type: object
   additionalProperties: false
@@ -356,6 +378,7 @@ SourceInput:
     - $ref: '#/SourceJSON'
     - $ref: '#/SourceCSV'
     - $ref: '#/SourceBigQuery'
+    - $ref: '#/SourceGA4BigQueryExport'
     - $ref: '#/SourceDocker'
 
 SourceUpdateCommercetools:
@@ -388,6 +411,7 @@ SourceUpdateInput:
     - $ref: '#/SourceJSON'
     - $ref: '#/SourceCSV'
     - $ref: '#/SourceBigQuery'
+    - $ref: '#/SourceGA4BigQueryExport'
     - $ref: '#/SourceUpdateDocker'
 
 DockerSourceStream:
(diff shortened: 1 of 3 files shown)


specs/ingestion/common/schemas/source.yml
```diff
@@ -261,6 +261,28 @@ SourceCSV:
...
+      description: Dataset ID of the BigQuery Source.
+    table:
+      type: string
+      description: Table name (for default BQ).
+    tablePrefix:
+      type: string
+      description: Table prefix (for Google Analytics).
+  required:
+    - projectID
+    - datasetID
```
jsok, line 282: table prefix is required too

specs/ingestion/common/schemas/source.yml
```diff
@@ -261,6 +261,28 @@ SourceCSV:
...
+  type: object
+  additionalProperties: false
+  properties:
+    projectID:
+      type: string
+      description: Project ID of the BigQuery Source.
+    datasetID:
+      type: string
+      description: Dataset ID of the BigQuery Source.
+    table:
```
jsok, line 274: There is no table field on the source type, only table prefix.

specs/ingestion/common/schemas/source.yml
```diff
@@ -261,6 +261,28 @@ SourceCSV:
...
+      type: string
+      description: Table name (for default BQ).
+    tablePrefix:
+      type: string
+      description: Table prefix (for Google Analytics).
```
jsok, line 279: This could be rewritten to match the GA4 docs a bit better.

Maybe "table prefix, e.g. events_ or events_intraday_ without the YYYYMMDD"

jsok (changes requested):

specs/ingestion/common/schemas/source.yml
```diff
@@ -261,6 +261,28 @@ SourceCSV:
   required:
     - url
 
+SourceGA4BigQueryExport:
+  type: object
+  additionalProperties: false
+  properties:
+    projectID:
+      type: string
+      description: Project ID of the BigQuery Source.
```
jonathaningram, line 270:
```suggestion
      description: GCP project ID that the BigQuery Export writes to.
```

specs/ingestion/common/schemas/source.yml
```diff
@@ -261,6 +261,28 @@ SourceCSV:
...
+    datasetID:
+      type: string
+      description: Dataset ID of the BigQuery Source.
```
jonathaningram, line 273:
```suggestion
      description: BigQuery dataset ID that the BigQuery Export writes to.
```

specs/ingestion/common/schemas/source.yml
jonathaningram: If helpful, referring to https://support.google.com/analytics/answer/9358801?hl=en will help you write this better per jsok's suggestion.

specs/ingestion/common/schemas/source.yml
```diff
@@ -261,6 +261,28 @@ SourceCSV:
...
+    table:
+      type: string
+      description: Table name (for default BQ).
+    tablePrefix:
+      type: string
+      description: Table prefix (for Google Analytics).
+  required:
+    - projectID
+    - datasetID
+  x-discriminator-fields:
```
jonathaningram, line 283: Out of interest, do you know how this works / what it does?

shortcuts: Strongly typed languages have a custom serializer/deserializer for oneOf schemas and the discriminator helps it matching the raw payload/reponse with the correct type if the key is found in it

jonathaningram: Right. In that case, I think we'd need more than just `projectID` here because `SourceBigQuery` also only has `projectID` in it?

JasonBerry: Yes, I think we should add `tablePrefix`?

jonathaningram: Yeah sounds like we should do that. I guess we can check what the generated clients look like and update it in a follow up if it's wrong.

jonathaningram (changes requested):
build issue?

jonathaningram (approved):

specs/ingestion/common/taskParameters.yml
```diff
@@ -18,12 +18,12 @@ action:
...
     items:
       $ref: './schemas/task.yml#/ActionType'
-    example: save,replace
+    example: save,replace,partial,append
 
 sourceID:
   name: sourceID
   in: query
-  description: The sourceIDs of the tasks to retrive.
+  description: The sourceIDs of the tasks to retrieve.
```
jsok, line 26: ❤️

jsok (approved):

shortcuts (approved):
Sorry I fell asleep ahah ty for the contribution!

PR #2859, opened 2024-03-12
docs: insights api reference

kai687:
## 🧭 What and Why

Update the insights spec for the Insights API reference documentation.


shortcuts:
I've uncommented your change because locally it worked 🤔

shortcuts (approved):
awesome!!

PR #2861, opened 2024-03-13
chore(CODEOWNERS): add owners for /specs/ingestion/

jonathaningram:
## 🧭 What and Why

Goal: Allow EX to merge our changes without having to wait for a round trip review. Post-merge reviews are always appreciated!

### Changes included:

- Add EX and Data Ingestion teams as owners of /specs/ingestion/

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -2,3 +2,5 @@
 # See https://help.github.com/articles/about-code-owners/
 
 * @algolia/api-clients-automation
+
+/specs/ingestion/ @algolia/api-clients-automation @algolia/data-ingestion @algolia/event-experiences


JasonBerry (approved):

jonathaningram:
These teams aren't public, closing.

shortcuts:
we also still enforce code owners to the API clients team for now as changing a spec also impacts the documentation and the clients themselves

jonathaningram:
@shortcuts thanks. Yeah we're just mildly concerned if we don't manage to get https://github.com/algolia/api-clients-automation/pull/2858 (for example) ready in this day cycle that we'll miss the boat on it and have to wait a round trip to get it merged. Don't suppose you have any ideas? We could try to get added to the `@algolia/api-clients-automation` team but it's not exactly appropriate since we only really want to be reviewing things we can review :)

shortcuts:
> @shortcuts thanks. Yeah we're just mildly concerned if we don't manage to get https://github.com/algolia/api-clients-automation/pull/2858 (for example) ready in this day cycle that we'll miss the boat on it and have to wait a round trip to get it merged. Don't suppose you have any ideas? We could try to get added to the `@algolia/api-clients-automation` team but it's not exactly appropriate since we only really want to be reviewing things we can review :) 

Yup that make sense, if it's only for Ingestion I guess that's fine considering the _low_ usage but we have to be careful of potential breaking changes/renames since our frontend dashboard uses it

BTW the DI and API clients team are basically the same so it's fine if you don't list it as code owners since it's private

jonathaningram:
> > @shortcuts thanks. Yeah we're just mildly concerned if we don't manage to get #2858 (for example) ready in this day cycle that we'll miss the boat on it and have to wait a round trip to get it merged. Don't suppose you have any ideas? We could try to get added to the `@algolia/api-clients-automation` team but it's not exactly appropriate since we only really want to be reviewing things we can review :)
> 
> Yup that make sense, if it's only for Ingestion I guess that's fine considering the _low_ usage but we have to be careful of potential breaking changes/renames since our frontend dashboard uses it
> 
> BTW the DI and API clients team are basically the same so it's fine if you don't list it as code owners since it's private

@shortcuts in the change I included DI and EX teams just for "correctness" since it may not always be true that "BTW the DI and API clients team are basically". But, it's a moot point anyway because both teams are private so this won't work. 

I requested access to join: `@algolia/api-clients-automation`. I wonder if it would be OK to allow me/us in that team, under the condition that we use our powers of approval sparingly and keep in mind "careful of potential breaking changes/renames since our frontend dashboard uses it"?

shortcuts:
> I requested access to join: `@algolia/api-clients-automation`. I wonder if it would be OK to allow me/us in that team, under the condition that we use our powers of approval sparingly and keep in mind "careful of potential breaking changes/renames since our frontend dashboard uses it"?

I guess it might be better to have a new public team for events so you can co-own the ingestion part? The project isn't super mature in term of external team contributions which could lead to unwanted changes slipping through

PR #2864, opened 2024-03-14
feat(swift): release process

Fluf22:
## 🧭 What and Why

🎟 JIRA Ticket: [DI-1724]

### Changes included:

- CI worflow for automatic new version release


clients/algoliasearch-client-swift/.github/workflows/release.yml
```diff
@@ -0,0 +1,83 @@
+name: Release package
+
+on:
+  push:
+    branches:
+      - next
+
+jobs:
+  test_macos:
```
shortcuts, line 9: do we have tests in the package directly?

clients/algoliasearch-client-swift/.github/workflows/release.yml
```diff
@@ -0,0 +1,83 @@
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Get versions
+        id: versions
+        shell: bash
+        run: |
+          echo "RELEASE_VERSION=$(cat clients/algoliasearch-client-swift/AlgoliaSearchClient.podspec | grep -Po "(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?")" >> $GITHUB_OUTPUT
```
shortcuts, line 50:
```suggestion
          echo "RELEASE_VERSION=$(cat AlgoliaSearchClient.podspec | grep -Po "(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?")" >> $GITHUB_OUTPUT
```

this will run in the client repo directly right?

clients/algoliasearch-client-swift/.github/workflows/release.yml
```diff
@@ -0,0 +1,83 @@
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Get versions
+        id: versions
+        shell: bash
+        run: |
+          echo "SWIFT_VERSION=$(cat config/.swift-version)" >> $GITHUB_OUTPUT
```
shortcuts, line 22: same you won't have this since it's in the repo directly

clients/algoliasearch-client-swift/.github/workflows/release.yml
Fluf22: Yes, but to avoid using too much credit, we only test on linux.
Here, I run a macos test to be sure it's compatible with all platforms ☺️

clients/algoliasearch-client-swift/.github/workflows/release.yml
Fluf22: Good catch omg 😱

clients/algoliasearch-client-swift/.github/workflows/release.yml
Fluf22: I'm gonna use something else, thanks for pointing that out

.github/workflows/test-release-swift.yml
shortcuts, line 1: we can remove this now, no?

shortcuts (commented):
looks great but I think we shouldn't run tests there if they don't run on the monorepo, if they run in both places it's fine but otherwise all the feedback should be at the monorepo level

.github/workflows/check.yml
```diff
@@ -357,15 +360,56 @@ jobs:
           name: clients-${{matrix.client.language }}
           path: clients-${{matrix.client.language }}.zip
 
+  swift_cts_macos:
```
shortcuts, line 363: can't we have a job between client_gen and codegen which only restore artifacts and run the CTS?

otherwise I think it should be named `client_gen_swift_macos` or something like the JS one since it does more than the CTS

shortcuts (commented):
noice

.github/workflows/check.yml
Fluf22: If you wish

clients/algoliasearch-client-swift/.github/workflows/release.yml
```diff
@@ -0,0 +1,45 @@
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Get versions
+        id: versions
+        shell: bash
+        run: |
+          echo "RELEASE_VERSION=$(cat AlgoliaSearchClient.podspec | grep s.version | sed 's/ //g' | cut -d \' -f 2)" >> $GITHUB_OUTPUT
```
shortcuts, line 22:
```suggestion
          echo "RELEASE_VERSION=$(grep "s.version = " AlgoliaSearchClient.podspec | cut -d "'" -f2)" >> $GITHUB_OUTPUT
```

I guess we can simplify since the format is guaranteed up to you

.github/workflows/check.yml
shortcuts: We don't need to generate on both platform I think

clients/algoliasearch-client-swift/.github/workflows/release.yml
Fluf22: It's perfect!

.github/workflows/check.yml
Fluf22: I understand, I'm making the change 😁

.github/workflows/check.yml
```diff
@@ -357,15 +360,54 @@ jobs:
...
+
+      - name: Setup
+        uses: ./.github/actions/setup
+        with:
+          type: minimal
+          language: swift
+
+      - name: Generate CTS
+        run: yarn cli cts generate swift ${{ fromJSON(needs.setup.outputs.SWIFT_DATA).toRun }}
+
```
shortcuts, line 396:
```suggestion
```

_I think_ they are also in the artifact

Fluf22: They are not retrieved by the custom action:
```
    await io.rmRF(`clients/algoliasearch-client-${language}`);
    await exec(`unzip -q -o clients-${language}.zip`);
    await io.rmRF(`clients-${language}.zip`);
```

Fluf22:
Let's try without them 😁

.github/workflows/check.yml
shortcuts: since the CI pushes them they do indeed, and then you can just update the needs to `client_gen`

shortcuts (approved):
nothing to add on my side!

shortcuts:
(gg)

Fluf22:
I'll have to add back the condition for the cts test to run only on the main branch 😁 
Just want to check it runs well, first

.github/workflows/check.yml
```diff
@@ -357,15 +360,49 @@ jobs:
...
+      ALGOLIA_APPLICATION_ID: ${{ secrets.ALGOLIA_APPLICATION_ID }}
+      ALGOLIA_ADMIN_KEY: ${{ secrets.ALGOLIA_ADMIN_KEY }}
+      MONITORING_API_KEY: ${{ secrets.MONITORING_API_KEY }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Download artifacts
+        uses: ./scripts/ci/actions/restore-artifacts
+        with:
+          type: all
```
millotp, line 384: no need to download all artifacts, you just need the spec and the swift client, but you need to edit the action to achieve that :(

scripts/ci/actions/restore-artifacts/src/index.ts
```diff
@@ -56,6 +70,12 @@ async function run(): Promise<void> {
     } else if (actionType === 'all') {
       await restoreSpecs();
       await restoreLanguages();
+    } else if (actionType === 'languages') {
+      const languages = core.getMultilineInput('languages');
```
millotp, line 74: can you reflect this in `action.yml` pls ?

scripts/ci/actions/restore-artifacts/src/index.ts
```diff
@@ -36,6 +36,20 @@ async function restoreSpecs(): Promise<void> {
...
+  const artifacts = await artifact.listArtifacts();
+  const langArtifact = artifacts.artifacts.find((a) => a.name === `clients-${language}`);
+  if (langArtifact === undefined) {
+    throw new Error(`No ${language} artifact found`);
+  }
+
+  await download(artifact, langArtifact.id);
+  await io.rmRF(`clients/algoliasearch-client-${language}`);
+  await exec(`unzip -q -o clients-${language}.zip`);
+  await io.rmRF(`clients-${language}.zip`);
```
millotp, line 50: could be nice to reuse the code from the other func

millotp (commented):
insane !

millotp:
why was the job skipped ?

Fluf22:
@millotp The Swift CTS on macOS? Because we are not on the `main` branch

millotp:
ah bah yes obviously, did it run successfully before ?

millotp (approved):
looks good !

Fluf22:
Yes, I waited for it to run successfully before adding back the condition :)

PR #2865, opened 2024-03-15
fix(specs): make languages an enum

kai687:
## 🧭 What and Why

Restrict allowed languages for settings like `queryLanguages` and `indexLanguages` to the list of supported languages, as currently the clients suggest that all string values are accepted, which would lead to errors when using.

### Changes included:

- Introduce a new enum `supportedLanguage`
- Restrict items for `indexLanguages`, `queryLanguages`, `removeStopWords`, `ignorePlurals`, and the dictionary response to be of `supportedLanguage` type.

## 🧪 Test

- Kotlin client fails to build.
- Swift tests fail with some type error I don't quite understand.


Fluf22:
Inre-run the Kotlin one to check it wasn't just a one-off error
The Swift error is interesting, let me take a look in the generated code

Fluf22:
It seems kotlin doesn't understand the oneOf anymore, now that there is a $ref in one of the element...
I can also take a look

Fluf22:
I found the issue for Swift, I'm gonna push a fix

Fluf22:
I found the issue for Kotlin, I'm gonna _try to_ push a fix 😄

Fluf22:
@aallam This [commit](https://github.com/algolia/api-clients-automation/pull/2865/commits/8b4ff1a9bcb2eb635b9f4989baaad3f1f5de601f) fixes the [issue](https://github.com/algolia/api-clients-automation/actions/runs/8299340295/job/22714930318#step:7:24), but I would love to have you to take a look at it, to confirm

Fluf22:
@kai687 All good, everything is fixed for now, you can merge!
@aallam will take a look next week, and we will push a fix in another PR if needed.

Fluf22 (approved):
Nice 🆗

PR #2883, opened 2024-03-17
docs: analytics api

kai687:
## 🧭 What and Why

Update the API reference for the Analytics API.

🎟 JIRA Ticket:

### Changes included:

- **New**: added missing endpoints: `/2/conversions/revenue`, `/2/conversions/addToCartRate`, `/2/conversions/purchaseRate`.
- **New**: added `revenueAnalytics` parameter and related response properties to `/2/searches` and `/2/hits` endpoints.
- **Fixed**: made some response properties nullable (if there are no events) to make it easier to see which properties can be null and which can be 0.


shortcuts:
huge!!

Fluf22:
You're doing an awesome job on those docs, it's only a small help from us!

Fluf22 (approved):
🆗

generators/src/main/java/com/algolia/codegen/AlgoliaJavaGenerator.java
```diff
@@ -107,6 +107,20 @@ public String toEnumVarName(String value, String datatype) {
...
+      case "!=":
+        return "NOT_EQUALS";
+      case "<":
+        return "LESS_THAN";
+      case "<=":
+        return "LESS_EQUALS";
+      case ">":
+        return "GREATER_THAN";
+      case ">=":
+        return "GREATER_EQUALS";
```
shortcuts, line 123: I wonder if we should not do that to every clients instead for the sake of harmonizing stuff

specs/analytics/common/parameters.yml
shortcuts, line 1: I'm sure there is parameters here that are also used in other specs and therefore should be in commons

I guess for now we can ignore it and update once all specs are finals

specs/analytics/paths/click/getPurchaseRate.yml
```diff
@@ -0,0 +1,73 @@
...
+  responses:
+    '200':
+      description: OK
+      headers:
+        x-ratelimit-limit:
+          $ref: '../../common/parameters.yml#/x-ratelimit-limit'
+        x-ratelimit-remaining:
+          $ref: '../../common/parameters.yml#/x-ratelimit-remaining'
+        x-ratelimit-reset:
+          $ref: '../../common/parameters.yml#/x-ratelimit-reset'
```
shortcuts, line 26: This makes me think only python and ruby provides a way to retrieve those at the response level, maybe all clients should have that capability @algolia/api-clients-automation wdyt

shortcuts (approved):
An other awesome contribution, thanks a ton @kai687 !!

PR #2887, opened 2024-03-19
chore(scala): set sonatype target

aallam:
## 🧭 What and Why

Broken scala client release process

### Changes included:

Set sonatype target


shortcuts (changes requested):
let's try it! (but wait for the current release to finish)

shortcuts (approved):
gg!

PR #2894, opened 2024-03-21
docs(php): Added an entry for php migration guide

damcou:
## 🧭 What and Why

🎟 JIRA Ticket: DI-1867

### Changes included:

- Added en entry for the php migration guide related to Configuration classes


millotp (approved):
good idea ! is it too late to improve the naming ? because it's all over the place in the `Configuration` directory, half of the time it's Config and the rest is Configuration

damcou:
> good idea ! is it too late to improve the naming ? because it's all over the place in the `Configuration` directory, half of the time it's Config and the rest is Configuration

I'm just dealing with the feedbacks we had for the PHP client (this missing change was mentioned in the migration guide for example). But I agree this is not consistent, I'll create a ticket to deal with it once the more pressing topics are handled ;)

PR #2895, opened 2024-03-21
fix(scripts): correctly remove codegen for js and dart

millotp:
## 🧭 What and Why

removeExistingCode was not targeting the correct folder for js and dart.


Fluf22 (approved):
Sorry for not checking those ones out

millotp:
no problem !

PR #2900, opened 2024-03-22
fix(swift): harden macOS CI

Fluf22:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

- fix the blocking issue (problems below will be adressed in a later PR)

- ~check for forgotten source file duplicates~
- ~change directory before running the podspec linting~

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -368,7 +368,6 @@ jobs:
       - client_gen
     if: |
       always() &&
-      github.ref == 'refs/heads/main' &&
       needs.setup.outputs.RUN_GEN == 'true' &&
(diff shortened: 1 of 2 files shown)


.github/workflows/check.yml
```diff
@@ -396,8 +396,44 @@ jobs:
       - name: Run CTS
         run: yarn cli cts run swift ${{ fromJSON(needs.setup.outputs.SWIFT_DATA).toRun }}
 
+  swift_podspec_check:
```
shortcuts, line 399: I think you can leave that in the matrix with a if language swift, so that you don't have to do all the setup steps, wdyt?

shortcuts (approved):

PR #2919, opened 2024-03-25
docs: separate labels for code samples

kai687:
## 🧭 What and Why

Add a separate `label` property to the `x-codeSamples` array.
This lets us use `lang` for the syntax highlighting grammar and an arbitrary `label`.
For now, this is only useful to render `C#` instead of `CSharp`.

See: [`x-codeSamples` (Redocly)](https://redocly.com/docs/api-reference-docs/specification-extensions/x-code-samples/).

🎟 JIRA Ticket:

### Changes included:

- Rename `mapLanguagesToCodeSampleSupporter` to `getCodeSampleLabel` and return `C#` when the language is `csharp`.

### DIFF
--- a/scripts/buildSpecs.ts
+++ b/scripts/buildSpecs.ts
@@ -9,16 +9,16 @@ import type { CodeSamples, Language, SnippetSamples, Spec } from './types.js';
 
 const ALGOLIASEARCH_LITE_OPERATIONS = ['search', 'customPost'];
 
-function mapLanguageToCodeSampleSupporter(language: Language): CodeSamples['lang'] {
+function getCodeSampleLabel(language: Language): CodeSamples['label'] {
   switch (language) {
     case 'csharp':
-      return 'CSharp';
+      return 'C#';
     case 'javascript':
       return 'JavaScript';
     case 'php':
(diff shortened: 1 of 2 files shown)


scripts/buildSpecs.ts
```diff
@@ -9,16 +9,16 @@ import type { CodeSamples, Language, SnippetSamples, Spec } from './types.js';
 
 const ALGOLIASEARCH_LITE_OPERATIONS = ['search', 'customPost'];
 
-function mapLanguageToCodeSampleSupporter(language: Language): CodeSamples['lang'] {
+function getCodeSampleLabel(language: Language): CodeSamples['label'] {
   switch (language) {
     case 'csharp':
-      return 'CSharp';
+      return 'C#';
```
shortcuts, line 15: it wasn't adding the highlighting when I added it but it seems good now 🤔 that's nice

shortcuts (approved):

scripts/buildSpecs.ts
shortcuts: ahhh ok I just realized what you did, NICE!!

PR #2920, opened 2024-03-26
feat(specs): support other timestamp formats to match API impl

jonathaningram:
## 🧭 What and Why

🎟 JIRA Ticket:

### Changes included:

- List changes


jonathaningram:
Unfortunately this is proving more difficult than it should be. 

We cannot use one of because it creates a `Timestamp` Java class but then the generator imports `java.sql.Timestamp`. Seems to be a long-standing bug: https://github.com/swagger-api/swagger-codegen/issues/10440

I'm sure there's a way around this by updating the template, but closing this for now as won't do.

PR #2924, opened 2024-03-26
fix(specs): query suggestions api

kai687:
## 🧭 What and Why

Update the spec for the Query Suggestions API.

🎟 JIRA Ticket:

### Changes included:

- Updated the `/1/configs` API response (there are no API keys returned in the response)
- Marked fields as `nullable: true` instead of using a `oneOf` with `null`


shortcuts:
> Don't really know how to fix the failing CTS. The API response for GET Query Suggestions config includes the `appId`, so I added the appId as expected response parameter in the CTS. Don't know why it still fails.

I tried calling the API and `appId` was wrongly cased it seems like https://github.com/algolia/api-clients-automation/pull/2924/commits/3028a521aa08dc6efe328b2341c0adaef1a8973c

specs/query-suggestions/common/responses/QuerySuggestionsConfigurationResponse.yml
```diff
@@ -0,0 +1,20 @@
+QuerySuggestionsConfigurationResponse:
+  type: object
+  description: API response for retrieving Query Suggestions configurations.
+  allOf:
+    - $ref: '#/AppID'
+    - $ref: '../schemas/QuerySuggestionsConfiguration.yml#/QuerySuggestionsConfigurationWithIndex'
+  required:
+    - appId
```
shortcuts, line 8:
```suggestion
    - appID
```

I forgot that

shortcuts (dismissed):
Incredible work

shortcuts (approved):
Awesome nothing to add!!

PR #2949, opened 2024-04-03
fix(specs): crawler api

kai687:
## 🧭 What and Why

Update the specs for the Crawler API to be suited as API reference documentation.

🎟 JIRA Ticket:

### Changes included:

- Add spec for deleting a crawler
- Complete crawler config and action parameters and descriptions
- Introduce tags
- Re-use schema objects if possible.

Only remaining issue is `recordExtractor` and `linkExtractor` since those are JavaScript functions as strings. I think it'll be better to document them outside the OpenAPI spec on a separate page.


specs/crawler/common/parameters.yml
```diff
@@ -1,18 +1,18 @@
 CrawlerIdParameter:
   name: id
   in: path
-  description: The Id of the targeted Crawler.
+  description: Crawler ID.
   required: true
   schema:
-    type: string
+    $ref: '#/CrawlerId'
```
shortcuts, line 7:
```suggestion
    $ref: '#/CrawlerID'
```

specs/crawler/common/parameters.yml
```diff
@@ -1,18 +1,18 @@
...
 TaskIdParameter:
-  name: tid
+  name: taskID
   in: path
-  description: The Id of the targeted Task.
+  description: Task ID.
   required: true
   schema:
-    type: string
+    $ref: '#/TaskId'
```
shortcuts, line 15:
```suggestion
    $ref: '#/TaskID'
```

idk what other thinks but I'm team capital D and I think we do that already elsewhere

specs/crawler/common/parameters.yml
```diff
@@ -22,138 +22,138 @@ CrawlerVersionParameter:
...
+  in: query
+  description: Algolia application ID for filtering the API response.
+  schema:
+    $ref: '#/applicationID'
+
+applicationID:
+  type: string
+  description: |
+    Algolia application ID where the crawler creates and updates indices.
+    The Crawler add-on must be enabled for this application.
```
shortcuts, line 57: I think some of those exist in `specs/common/parameters.yml`

specs/crawler/spec.yml
```diff
@@ -1,45 +1,97 @@
...
 components:
   securitySchemes:
     BasicAuth:
       type: http
       scheme: basic
 servers:
-  - url: https://crawler.algolia.com/api/1
-  - url: https://crawler-dev.algolia.com/api/1
-  - url: http://localhost:8000/api/1
+  - url: https://crawler.algolia.com/api
```
shortcuts, line 43: _I think_ we can keep the dev one too wdyt

shortcuts: I was more thinking for Algolians but since there's no API clients right now we might not have a need for it

specs/crawler/common/parameters.yml
shortcuts: No problem, not a blocker at all, if there's none than can be used let's leave it like that for now

specs/crawler/paths/crawler.yml
```diff
@@ -98,3 +61,30 @@ patch:
...
+#               success:
+#                 type: boolean
+#                 description: Whether the deletion was successful.
+#               data:
+#                 type: object
+#                 properties:
+#                   taskId:
+#                     $ref: '../common/parameters.yml#/TaskId'
+#     '403':
+#       description: Forbidden
```
shortcuts, line 90:
```suggestion
```

I guess we can remove it for now as we don't expose the client, just the documentation

shortcuts (dismissed):
LOVELY

gg!!

shortcuts (approved):
thanks a lot again for this!!

PR #2954, opened 2024-04-04
fix(specs): usage api

kai687:
## 🧭 What and Why

Update of the Usage API spec.
Unfortunately, I can't test the Usage API (can't make any requests) so I can't independently verify the information like I did for the other APIs, so I only did a few style updates.

### Changes included:

- Use an enum for the `{statistics}` parameter
- Increased reuse of existing schemas.


specs/usage/spec.yml
```diff
@@ -2,26 +2,61 @@ openapi: 3.0.2
...
-    url: 'https://www.algolia.com/doc/rest-api/usage/'
 components:
   securitySchemes:
-    BasicAuth:
-      type: http
-      scheme: basic
+    appId:
+      $ref: '../common/securitySchemes.yml#/appId'
+    apiKey:
+      $ref: '../common/securitySchemes.yml#/apiKey'
```
shortcuts, line 45: it's really this and not basic? if so we will be able to generate clients then!

shortcuts (commented):
this looks wonderful!

shortcuts:
nice my fix worked :D

shortcuts (approved):
thanks a ton!!

PR #2958, opened 2024-04-04
fix(specs): recommend api

kai687:
## 🧭 What and Why

Update the spec for the Recommend API to make it suitable as API reference

### Changes included:

- Added Looking Similar model to the "Get recommendations" response and remove "Recommended for you" (didn't find any reference anywhere for this).
- Recommend Rules != Search Rules, so I typed the schema, trying to reuse what I can
- Small changes to common parameters: 
  - made `x-algolia...` headers lowercase. HTTP headers are case-insensitive anyway. It just looks better IMO.
  - changed the description of the Search filters parameter to make it usable for recommendations as well.

### Things that will still be wrong

Fixing this requires some remodeling in the "common" parameters/Search API response,
which might break things unrelated to Recommend. 

- Recommendation API response != Search API response.
- `queryParameters` for getting recommendations does not accept `hitsPerPage`, `page`, `offset`, `length`. 

## 🧪 Test

TODO: fix tests for the new responses.


shortcuts:
> No idea why Dart fails, but Java, Kotlin, Scala fails because `threshold` is tested with a 42, but the `threshold` parameter can be a float. What's the right way of fixing this? Testing with a float?

I guess so yes if it's expected to have this format, but it might be an issue with the CTS template

shortcuts (commented):
maybe @raed667 @marialungu @francoischalifour  can help for the review on this one

for the recommended for you, I know it's available in the JavaScript client https://github.com/search?q=repo%3Aalgolia%2Falgoliasearch-client-javascript%20recommended-for-you&type=code

shortcuts:
> And indeed, `recommended-for-you` seems to be some (private) beta model. The Recommend team can decide if that should be documented here or not.

since it has been added by the recommend team on the javascript client we should keep it for now and let them remove it if necessary

shortcuts (approved):
purrrrfect

PR #2960, opened 2024-04-05
feat(clients): add `usage` client

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-1917

### Changes included:

follow up of https://github.com/algolia/api-clients-automation/pull/2954, we can generate the usage client now that the specs is validated


generators/src/main/java/com/algolia/codegen/AlgoliaKotlinGenerator.java
```diff
@@ -195,6 +195,15 @@ public OperationsMap postProcessOperationsWithModels(OperationsMap objs, List<Mo
   @Override
   public String toEnumVarName(String value, String datatype) {
     if (!"String".equals(datatype)) return super.toEnumVarName(value, datatype);
+    // In some cases, the API might accept characters instead of the textual notation, we will
+    // replace it internally so that it doesn't output the character itself.
+    switch (value) {
+      case "90p_processing_time":
+        return "NinetypProcessingTime";
```
Fluf22, line 202: Shouldn't it be `NinetyPProcessingTime`? 😬
Non-blocking, it's naming 😉

scripts/pre-gen/removeExistingCodegen.ts
```diff
@@ -87,14 +87,14 @@ export async function removeExistingCodegen({
...
+    recursive: true,
   });
 
   // Delete client api folder/file
-  await run(`rm -rf ${path.join(baseApiFolder, clientApi)}`, {
-    cwd: folder,
-    language,
+  await fsp.rm(path.join(folder, path.join(baseApiFolder, clientApi)), {
+    force: true,
+    recursive: true,
```
Fluf22, line 98: This was changed on purposes, because fsp doesn't work with glob patterns

Fluf22 (commented):
Just a confirmation about the file removing process

generators/src/main/java/com/algolia/codegen/AlgoliaKotlinGenerator.java
shortcuts: Actually I'm doing an other PR that will add this enum var logic to all clients in order to have something common, but yes I can capitalize the P

scripts/pre-gen/removeExistingCodegen.ts
shortcuts: Ahhh oki I did not knew, I think there's an issue for new clients with rm -rf when the folder doesn't exist though, I'll try again

Fluf22: Interesting

shortcuts: ah maybe it's just the `run` utility since giving a folder makes a cd

Fluf22: Aaaaah, so in that case we should not put the `cwd` to avoid it trying to `cd` in a non-existant folder
Instead, using `path.join` instead will use `rm` directly, and not fail if the directory doesn't exist

shortcuts: yup I just tried that it should work

shortcuts: https://github.com/algolia/api-clients-automation/pull/2960/commits/abcd4b0cbee6a20ba1275ffead3949461af53512

Fluf22: noice.gif

Fluf22 (approved):
Seems perfect
The whole toolchain is awesome

PR #2962, opened 2024-04-06
docs: contribution guide for docs

kai687:
## 🧭 What and Why

Improve contribution guidelines for adding new clients and the documentation aspects of it.

### Changes included:

- Edit the "Add new clients" guide to give it a clearer structure and edit for style.
- Add a new "docs" guide to define elements where we want to achieve consistent docs.
- Add the `remark-smartypants` plugin to render straight quotes as "correct" quotes.
- Left-align table headers (looks better IMO)

I'll update the specs to follow these guidelines in a separate PR.

PR #2976, opened 2024-04-07
fix(specs): update documentation for consistency

kai687:
## 🧭 What and Why

Follow up with PR #2962 to align the APIs (except the in progress Recommend, Personalization, and A/B testing APIs) with the guidelines established by that PR.

### Changes included:

- No HTML
- Consistent descriptions of date-time and timestamps
- Consistent intro sections with `h2` rather than `h1` headings

I didn't update the `X-Algolia-User-ID` header, because making it all lowercase breaks the tests. This parameter is only used in the deprecated Multi-cluster management endpoints, so I thought it's ok to leave it as it is.

### DIFF
--- a/specs/analytics/common/parameters.yml
+++ b/specs/analytics/common/parameters.yml
@@ -316,7 +316,7 @@ x-ratelimit-remaining:
     type: integer
 
 x-ratelimit-reset:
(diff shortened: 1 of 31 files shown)


shortcuts (approved):
perfect, nothing to add!

PR #2990, opened 2024-04-11
fix(specs): attribute and index names case sensitive

kai687:
## 🧭 What and Why

It's sometimes not straightforward to know which parts are case-sensitive or not and catches people by surprise.

- `indexName` in request bodies are case-sensitive
- attribute names referring to records, such as `searchableAttributes`

This change makes this more visible.


millotp (approved):
nice addition !

PR #3013, opened 2024-04-17
fix(specs): remove periods on summaries

kai687:
## 🧭 What and Why

Add a new ESLint rule to remove trailing `.` on `summary` fields.
This will make the docs look a tiny bit better, since the `summary` entries are used as headers.

### Changes included:

- Add new rule `no-final-dot`
- Modify existing rule `end-with-dot`
- Remove final `.` from summaries

## 🧪 Test

I added a unit test for the rule.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -9,7 +9,7 @@ module.exports = {
     '**/target',
     '**/.yarn',
     'website/specs',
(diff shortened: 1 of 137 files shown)


shortcuts (approved):
nothing to add I like the idea!

Fluf22 (approved):
Perfect 🆗

PR #3033, opened 2024-04-22
fix(cts): use the oldest authentication to avoid failure on creation

millotp:
## 🧭 What and Why

There was a new auth created on `T8JK9S7I7X` that made the CTS fail, we can list by ascending order to prevent this issue.


shortcuts (approved):
great catch
great fix
great pierre

morganleroi (dismissed):

shortcuts (approved):

PR #3042, opened 2024-04-26
fix(specs): simplify filter types

millotp:
## 🧭 What and Why

Use recursive types to avoid creating in unnecessary oneOf and allow to use the full potential of the filters

### DIFF
--- a/templates/ruby/base_object.mustache
+++ b/templates/ruby/base_object.mustache
@@ -23,10 +23,8 @@
       end
       {{#getAdditionalPropertiesIsAnyType}}
 
-      # merge additional_properties into transformed_hash
-      @additional_properties.each_pair do |k, v|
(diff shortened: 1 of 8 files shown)


templates/ruby/base_object.mustache
```diff
@@ -23,10 +23,8 @@
       end
       {{#getAdditionalPropertiesIsAnyType}}
 
-      # merge additional_properties into transformed_hash
-      @additional_properties.each_pair do |k, v|
-        transformed_hash[k.to_sym] = v
-      end unless @additional_properties.nil?
+      # add extra attribute to transformed_hash
```
shortcuts, line 26: huge fix

shortcuts (approved):
insane feature gg

PR #3056, opened 2024-04-30
fix(python): properly remove existing codegen

millotp:
## 🧭 What and Why

Models were not properly removed


scripts/pre-gen/removeExistingCodegen.ts
```diff
@@ -66,8 +66,8 @@ export async function removeExistingCodegen({
     case 'python':
       clientModel = toSnakeCase(clientName);
       clientApi = toSnakeCase(clientName);
-      baseModelFolder = '';
-      baseApiFolder = '';
+      baseModelFolder = 'algoliasearch/models';
```
shortcuts, line 69: this folder doesn't exist it's `algoliasearch/*/models`

shortcuts (approved):

PR #3057, opened 2024-05-03
docs: update contribution guideline

kai687:
## 🧭 What and Why

This PR updates the guidelines for writing long descriptions.
Essentially, don't use Markdown in the first sentence/paragraph in the descriptions.

The Algolia CLI uses the Search API spec to generate flags from the API parameter/index settings. The help text is generated from the first sentence of the `description` key.

It looks better on the command line if the help text doesn't contain Markdown.

For the most part, I already updated the specs with that in mind. There are some Markdown links, but they will be automatically removed when generating the flags and help text.


millotp (approved):
good idea !

PR #3069, opened 2024-05-07
fix(python): chunked_batch helper

kai687:
## 🧭 What and Why

Fix a bug in the `chunked_batch` helper for the Python API client.
The bug shows when using the `replace_all_objects` helper—the new index won't have all the records:

- < 1,000 records: only the first record (`0 % batch_size == 0`) will be sent.
- \> 1,000 records: only the first `1 + n x 1,000` records will be sent. The records after the last integer multiple of `batch_size` won't be sent.

### Changes included:

- Send the request when the length of the `request` array reaches the `batch_size`, or when the end of the `objects` list is reached.
- Change the type of the `objects` "list" in `replace_all_objects` and `chunked_batch` to be an `Iterable` instead of an explicit list. This makes it possible to use also with things like list comprehensions.

## 🧪 Test

No additional test added (didn't quite know how we could test this, as the API is mocked in these tests... we would have to test whether the replaced index has the correct records in it).

### DIFF
--- a/clients/algoliasearch-client-csharp/algoliasearch/Models/Search/SearchForFacetValuesResponse.cs
+++ b/clients/algoliasearch-client-csharp/algoliasearch/Models/Search/SearchForFacetValuesResponse.cs
@@ -25,7 +25,7 @@ public partial class SearchForFacetValuesResponse
   /// Initializes a new instance of the SearchForFacetValuesResponse class.
   /// </summary>
   /// <param name="facetHits">Matching facet values. (required).</param>
-  /// <param name="exhaustiveFacetsCount">See the &#x60;facetsCount&#x60; field of the &#x60;exhaustive&#x60; object in the response. (required).</param>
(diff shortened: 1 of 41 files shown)


templates/python/search_helpers.mustache
```diff
@@ -223,7 +223,7 @@
         responses: List[BatchResponse] = []
         for i, obj in enumerate(objects):
             requests.append(BatchRequest(action=action, body=obj))
-            if i % batch_size == 0:
+            if len(requests) == batch_size or i == len(objects) - 1:
```
millotp, line 226: I don't think you can take the length of a `Iterable` type, because the length is determined after iterating over it

shortcuts:
I think we have that issue in other clients too, i'll add a ticket to our backlog just to be sure

alternatively, a universal solution, might be to first build a list of list of batch by iterating over all objects, then iterating on the list to do the requests

PR #3086, opened 2024-05-13
fix(cts): use a ever lasting abtest

millotp:
## 🧭 What and Why

Instead of creating a new abtest every 3 month because of abtest limit, @JonathanMontane created a abtest that last a 100 year ! (thanks a lot)

Also fix python init files


shortcuts (approved):

JonathanMontane:
Cc @cdhawke - just to let you know about that special Abtest 😁

PR #3089, opened 2024-05-16
fix(specs): exhaustiveFacetsCount is not deprecated for sffv

kai687:
## 🧭 What and Why

The `exhaustiveFacetsCount` property is considered deprecated in the Search response,
and users should rely on `exhaustive.facetsCount` instead.

The response from Search for facet values doesn't have an `exhaustive` object,
so `exhaustiveFacetsCount` can't be deprecated and we need to use different descriptions in both cases.

🎟 JIRA Ticket:

### Changes included:

- Don't reuse the definition of `exhaustiveFacetsCount`, but inline them for the Search and Search for facet values responses, not marking it deprecated in the Sffv response.


Fluf22 (approved):
🆗

shortcuts (approved):
ggreat catchhh

PR #3091, opened 2024-05-17
fix(kotlin): ci release workflow

Fluf22:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

- We were not automatically releasing new Kotlin versions


millotp (approved):
good catch !

PR #3139, opened 2024-06-05
Feat/generate guides

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

provide the snippet methods directly in the docs guides, so you can generate a guide for all clients, in one go

todo:
- [ ] update all mustache to include the import separator
- [ ] add a waitTask helper to generate a wait for the snippet
- [ ] avoid using e2e tests index names for snippets
- [ ] update all guides


website/docs/clients/guides/replace-all-rules-synonyms.mdx
```diff
@@ -2,123 +2,23 @@
...
-
-var client = new SearchClient("<YOUR_APP_ID>", "<YOUR_API_KEY>");
-await client.SaveRulesAsync("<YOUR_INDEX_NAME>", rules, forwardToReplicas: false, clearExistingRules: true);
-```
-
-</TabItem>
-
-
-</TabsLanguage>
+${waitForTaskSnippet()}
```
millotp, line 20: this should be `waitForAppTaskSnippet` but let's tackle this in another PR

millotp: because its only used for dict, should we call it `waitForDictionary` ?

millotp (approved):

website/docs/clients/guides/replace-all-rules-synonyms.mdx
shortcuts: I'm not sure maybe we missed some use cases, let's keep the legacy name for safety

PR #3140, opened 2024-06-06
fix(go): upgrade the linter

millotp:
## 🧭 What and Why

The go linter has finally added a option to lint generated file so we can remove the exception


shortcuts (approved):
Perfect

PR #3154, opened 2024-06-10
fix(clients): chunked batch helper

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-1970

### Changes included:

supersede https://github.com/algolia/api-clients-automation/pull/3069

ensure clients push the first (if only one) and last batch with the same logic in all clients


templates/javascript/clients/client/api/helpers.mustache
```diff
@@ -279,9 +279,10 @@ async chunkedBatch({ indexName, objects, action = 'addObject', waitForTasks, bat
   let requests: Array<BatchRequest> = [];
   const responses: Array<BatchResponse> = [];
 
-  for (const [i, obj] of objects.entries()) {
+  const objectEntries = objects.entries();
+  for (const [i, obj] of objectEntries) {
     requests.push({action, body: obj});
-    if (i % batchSize === 0) {
+    if (requests.length === batchSize || (i === objects.length-1 && requests.length > 0)) {
```
millotp, line 285: the lentgh must be > 0 because you push just above, same in the other languages, you don't need that check

millotp (commented):
nice fix !

templates/javascript/clients/client/api/helpers.mustache
shortcuts: ah bah yes!

millotp (approved):
gg !

PR #3157, opened 2024-06-11
fix(specs): rankingInfo required properties

millotp:
## 🧭 What and Why

According to the doc examples, `promoted`, `words` and `filters` are not required on the `rankingInfo` response, and it's blocking on the dart client: https://github.com/algolia/algoliasearch-client-dart/issues/6


morganleroi (approved):
GG

PR #3159, opened 2024-06-11
fix(clients): remove unused models from lite clients

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-2440

### Changes included:

js and dart ships unused models because the lite specs keep them


millotp:
so nice ! too bad it doesn't impact the bundle size for js 

Screenshot 2024-06-11 at 14 29 13

shortcuts:
> so nice ! too bad it doesn't impact the bundle size for js
> 
> Screenshot 2024-06-11 at 14 29 13

yes the umd doesn't include the models it's pure js, it might impact the final esm bundle size though

millotp (approved):
gg ! this could be used to find dead specs also, if we have any

PR #3163, opened 2024-06-11
chore(go): update enum casing

Fluf22:
## 🧭 What and Why

🎟 JIRA Ticket: [DI-1641]

### Changes included:

- Make enum have a upper snake case

[DI-1641]: https://algolia.atlassian.net/browse/DI-1641?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/generators/src/main/java/com/algolia/codegen/AlgoliaGoGenerator.java
+++ b/generators/src/main/java/com/algolia/codegen/AlgoliaGoGenerator.java
@@ -1,9 +1,12 @@
 package com.algolia.codegen;
 
 import com.algolia.codegen.exceptions.*;
+import com.algolia.codegen.lambda.ScreamingSnakeCaseLambda;
 import com.algolia.codegen.utils.*;
 import com.algolia.codegen.utils.OneOf;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Iterables;
+import com.samskivert.mustache.Mustache;
 import io.swagger.v3.oas.models.OpenAPI;
 import io.swagger.v3.oas.models.Operation;
 import io.swagger.v3.oas.models.servers.Server;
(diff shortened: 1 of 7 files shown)


generators/src/main/java/com/algolia/codegen/AlgoliaGoGenerator.java
```diff
@@ -29,6 +30,10 @@ public void processOpts() {
     additionalProperties.put("packageName", client.equals("query-suggestions") ? "suggestions" : client);
     additionalProperties.put("enumClassPrefix", true);
     additionalProperties.put("isSearchClient", client.equals("search"));
+    additionalProperties.put(
```
millotp, line 33: the preferred way is to use `addMustacheLambdas` like [here](https://github.com/algolia/api-clients-automation/blob/main/generators/src/main/java/com/algolia/codegen/cts/AlgoliaCTSGenerator.java#L92) but that works too

generators/src/main/java/com/algolia/codegen/AlgoliaGoGenerator.java
```diff
@@ -29,6 +30,10 @@ public void processOpts() {
...
+      "lambda.toUpperSnakeCase",
```
millotp, line 34: I think the official name is `SCREAMING_SNAKE_CASE` ahah

millotp (commented):
the DX is way better, thanks !

generators/src/main/java/com/algolia/codegen/AlgoliaGoGenerator.java
Fluf22: I'll update it, I was wondering what naming to use

Fluf22:
The rebase + conflict resolution was failing because JetBrains kept updating the go.mod+sum of the playground, which made me do the edit manually, and I must have missed something since some changes were reverted
I'll update again and push a fix

millotp (approved):
beautiful !

PR #3165, opened 2024-06-12
chore(java): Changed fromValue(String) to use the actual Enum instead

larskristianhaga:
## 🧭 What and Why

🎟 JIRA Ticket: No ticket

### Changes included:

- Changed fromValue(String) to use the actual Enum instead

## 🧪 Test

Not ran, should be fine if compiles

### DIFF
--- a/snippets/java/src/test/java/com/algolia/Ingestion.java
+++ b/snippets/java/src/test/java/com/algolia/Ingestion.java
@@ -19,7 +19,7 @@ class SnippetIngestionClient {
     // Call the API
     client.createAuthentication(
       new AuthenticationCreate()
-        .setType(AuthenticationType.fromValue("oauth"))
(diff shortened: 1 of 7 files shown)


millotp:
Hello @larskristianhaga, the files you edited are generated, and they will get overwritten when merged.
This project uses `openapi-generator` to generate those snippets, the file to update in https://github.com/algolia/api-clients-automation/blob/main/templates/java/tests/generateInnerParams.mustache#L20

larskristianhaga:
Damn, thanks for reply.
I'll see if I get the time to update the correct files instead

millotp:
changes have been merged in https://github.com/algolia/api-clients-automation/commit/4003918ee66ae3666e0abda9cc0c369d999cd98c, all snippets tests and docs should use the actual enum now 😄

PR #3168, opened 2024-06-12
chore(scripts): generate LTS files

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-1770

### Changes included:

gave a shot at generating a release schedule table for deprecated/active/maintained versions, like on https://nodejs.org/en/about/previous-releases

this script will be ran before a release PR is triggered, it follows the exact same process as the node one https://github.com/nodejs/release#release-schedule using their own tool https://github.com/nodejs/lts-schedule

### How does it work

Since we only support the last patch version technically, every time a new release is made, we can delete the current version

When a major is released, the current version enters in a `maintenance` state, every time we release, we will see if there's maintenance versions that are out of date

if you want to try it our, make `createReleasePR` stop after the update of the config files (L515), and run it multiple times to see maintenance version popping

### Output example

I've gave a try to using their tool to generate the png too, which seems pretty nice

### js
javascript-lts

### kotlin
kotlin-lts

### php
php-lts

### python
python-lts

### ruby
ruby-lts

### scala
scala-lts

### swift
swift-lts

### csharp
csharp-lts

### dart
dart-lts

### go
go-lts

### java
java-lts


millotp (approved):
perfect ! nice tool

morganleroi (approved):
Huge ! Indeed the generated png are awesome !

PR #3174, opened 2024-06-13
feat(scripts): allow subset release and major

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-2444

### Changes included:

updates the release script to be able to only release a subset of clients, and also to promote that subset (or every) to the next major version

this will allow us to release the GA without introducing a breaking change commit, and also allows us to prevent bypassing `dart` which is already stable.

also moves the release script to the CLI script, makes more sense and allows leveraging the current logic

### DIFF
--- a/scripts/cli/index.ts
+++ b/scripts/cli/index.ts
@@ -9,7 +9,9 @@ import { startTestServer } from '../cts/testServer';
 import { formatter } from '../formatter.js';
 import { generate } from '../generate.js';
 import { playground } from '../playground.js';
+import { createReleasePR } from '../release/createReleasePR.js';
 import { snippetsGenerateMany } from '../snippets/generate.js';
+import type { Language } from '../types.js';
 
 import type { LangArg } from './utils.js';
 import {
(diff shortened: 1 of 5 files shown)


morganleroi (approved):
Let's test this !! Awesome work here. 🥇

scripts/cli/index.ts
```diff
@@ -200,4 +207,23 @@ program
     await snippetsGenerateMany(generatorList({ language, client, clientList }));
   });
 
+program
```
millotp, line 210: could be really cool to have a dry run option for this one !

scripts/release/createReleasePR.ts
```diff
@@ -208,18 +208,37 @@ export function getNextVersion(current: string, releaseType: semver.ReleaseType
...
+  languages: Language[];
+  major?: boolean;
 }): Promise<Versions> {
   const versionsToPublish: Versions = {};
 
   for (const [lang, version] of Object.entries(versions)) {
+    const currentVersion = versions[lang].current;
+
+    if (!languages.includes(lang as Language)) {
+      console.log(`${lang} is not in the given language list, skipping release`);
```
millotp, line 225: I think this console log is interesting for debug mode, but in regular use it would be more interesting to log only the chosen languages

millotp (commented):
such a good idea !

scripts/release/createReleasePR.ts
shortcuts: the release script is pretty verbose which I believe is nice as it's a process that can fail in many ways :/ I'm fine with keeping the log, wdyt?

scripts/cli/index.ts
shortcuts: true!! I'll add it in a subsequent pr

PR #3176, opened 2024-06-13
fix(scripts): waitForAllRelease works with dart

millotp:
## 🧭 What and Why

The `waitForAllRelease` script doesn't work for dart for some reason when using the `workflowID` with a branch.
Now the api supports using the name of the yaml directly so we can simplify the script, and use better types.


shortcuts (approved):
much clear!!

PR #3177, opened 2024-06-13
chore: remove utils CI for JS

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

Utils are build anyway, so no need to do it twice, and we can run all tests it won't take more time than spinning up jobs


millotp (approved):
I'm loving it

PR #3196, opened 2024-06-17
feat(scripts): push to docs on release

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-2458

### Changes included:

bring back our beloved spread script, which publishes to the official documentation:
- the documentation specs
- the generated code snippets
- the SLA chart and PNGs

we only spread after a successful release, see example PR here: https://github.com/algolia/doc/pull/9098

### DIFF
--- a/scripts/buildSpecs.ts
+++ b/scripts/buildSpecs.ts
@@ -149,6 +149,10 @@ async function transformBundle({
       toAbsolutePath(`website/src/generated/${clientName}-snippets.js`),
       `export const snippets = ${transformCodeSamplesToGuideMethods(JSON.parse(JSON.stringify(snippetSamples)))}`,
     );
+    await fsp.writeFile(
+      toAbsolutePath(`website/src/generated/${clientName}-snippets.json`),
+      transformCodeSamplesToGuideMethods(JSON.parse(JSON.stringify(snippetSamples))),
+    );
   }
 
   for (const [pathKey, pathMethods] of Object.entries(bundledSpec.paths)) {
--- a/scripts/buildSpecs.ts
+++ b/scripts/buildSpecs.ts
@@ -145,13 +145,15 @@ async function transformBundle({
     : ({} as SnippetSamples);
 
   if (docs) {
+    const snippets = transformCodeSamplesToGuideMethods(JSON.parse(JSON.stringify(snippetSamples)));
+    // the JS file will be removed once algolia/doc leverages the JSON one
     await fsp.writeFile(
       toAbsolutePath(`website/src/generated/${clientName}-snippets.js`),
-      `export const snippets = ${transformCodeSamplesToGuideMethods(JSON.parse(JSON.stringify(snippetSamples)))}`,
+      `export const snippets = ${snippets}`,
     );
     await fsp.writeFile(
       toAbsolutePath(`website/src/generated/${clientName}-snippets.json`),
-      transformCodeSamplesToGuideMethods(JSON.parse(JSON.stringify(snippetSamples))),
+      snippets,
     );
   }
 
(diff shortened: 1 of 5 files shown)


scripts/buildSpecs.ts
```diff
@@ -145,9 +145,15 @@ async function transformBundle({
     : ({} as SnippetSamples);
 
   if (docs) {
+    const snippets = transformCodeSamplesToGuideMethods(JSON.parse(JSON.stringify(snippetSamples)));
+    // the JS file will be removed once algolia/doc leverages the JSON one
```
millotp, line 149: we should support the `JSON` one you mean, the `js` one is for our doc.

scripts/ci/codegen/pushToAlgoliaDoc.ts
```diff
@@ -0,0 +1,92 @@
+/* eslint-disable no-console */
```
millotp, line 1: which version of the file did you take for this ? there was a lot of bug fixes done to the previous one

scripts/ci/codegen/pushToAlgoliaDoc.ts
```diff
@@ -0,0 +1,92 @@
...
+  );
+  await fsp.rm(tempGitDir, { force: true, recursive: true });
+  await run(`git clone --depth 1 ${githubURL} ${tempGitDir}`);
+  await run(`git checkout -B ${targetBranch}`, { cwd: tempGitDir });
+
+  const dest = toAbsolutePath(`${tempGitDir}/app_data/api/specs`);
+  await emptyDirExceptForDotGit(dest);
+  await run(`cp ${toAbsolutePath('specs/bundled/*.doc.yml')} ${dest}`);
+  await run(`cp ${toAbsolutePath('config/release.config.json')} ${dest}`);
+  await run(`cp ${toAbsolutePath('website/src/generated/*.json')} ${dest}`);
```
millotp, line 51: the snippets are already contained in the bundled spec, is it necessary to also send the json file ?

millotp (commented):
nice

scripts/buildSpecs.ts
shortcuts: Yes but the JS one will be removed once the doc is live that's what i meant

scripts/ci/codegen/pushToAlgoliaDoc.ts
shortcuts: Yup it's so that they can leverage the JSONs for the guides like we did!

scripts/ci/codegen/pushToAlgoliaDoc.ts
shortcuts: _I think_ the last but you are right I'll double check!

scripts/buildSpecs.ts
millotp: so we plan on removing our doc entirely ? :(

scripts/ci/codegen/pushToAlgoliaDoc.ts
millotp: we had to do this to have both the snippets for redoc, and the snippets for custom guide, are they using the same techno ?

scripts/buildSpecs.ts
shortcuts: Yup at least for GA, @kai687 has great plans for it!

scripts/ci/codegen/pushToAlgoliaDoc.ts
shortcuts: Yup! Given the time constraint it seems like the better, wdyt?

millotp: I don't know how well this will work with ruby but why not !

shortcuts: I believe fine with it's simple json input but at least it gives @kai687 the opportunity to try the current solution

millotp (approved):

PR #3197, opened 2024-06-17
feat(scripts): move helpers to snippet files for algolia doc

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-2458

### Changes included:

duplicates the snippets for our netlify doc to the JSON of the initial snippet, with a template pattern, this allows us to ship a single file to the documentation, which doesn't use JS anyway so it's better if we let them do the templating/variable replace directly

also move the specs logic to its own folder, the file was pretty complicate to navigate in


scripts/specs/helper-snippets.ts
```diff
@@ -0,0 +1,140 @@
...
+          maxHitsPerQuery: 20,
+          maxQueriesPerIPPerHour: 100,
+          validity: 300
+      ))`,
+    delete: `try await client.waitForApiKey(with: response.key, operation: ApiKeyOperation.delete)`,
+  },
+};
+
+export const waitForTask = {
+  csharp: `await client.WaitForTaskAsync("<<indexName>>", response.TaskID);`,
```
shortcuts, line 126: I've used this format for templating for now but it will most likely evolve from @kai687 's PoC

scripts/specs/snippets.ts
```diff
@@ -0,0 +1,131 @@
...
+          snippetSamples[language].waitForApiKey = {
+            default: waitForApiKey[language],
+          };
+        }
+
+        if (!('waitForTask' in snippetSamples[language])) {
+          snippetSamples[language].waitForTask = {
+            default: waitForTask[language],
+          };
+        }
```
shortcuts, line 62: this is actually the only bits that changed, lmk if you want 2 PRs instead for the file move

scripts/specs/snippets.ts
```diff
@@ -0,0 +1,131 @@
+import fsp from 'fs/promises';
+
+import { GENERATORS, capitalize, createClientName, toAbsolutePath } from '../common.js';
+import type { CodeSamples, Language, SnippetForMethod, SnippetSamples } from '../types.js';
```
millotp, line 4: can you move some of those types to the new `types.ts` file pls ?

scripts/specs/snippets.ts
millotp, line 62: all of those can be out of the for loop and without the `if` I think

scripts/specs/helper-snippets.ts
millotp: maybe we can use the same format as for our CTS, we use `${{localhost}}` for example

millotp (commented):
look good ! nice split

scripts/specs/helper-snippets.ts
shortcuts: It will change for sure later anyway so... I guess we don't really mind for now

scripts/specs/snippets.ts
shortcuts: indeed!

scripts/specs/snippets.ts
shortcuts: indeed2!

millotp (approved):
perfect !

PR #3198, opened 2024-06-17
chore: skip website build/deploys on non main target

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

we build the doc for every kind of PRs which consumes a lot of resources for nothing, we can at least limit it to the main branch and PRs that targets it


millotp (approved):
why not !

If we want to save the planet we should rewrite our CLI in rust ahah

PR #3202, opened 2024-06-18
chore(deps): solve some dependabots warning

millotp:
## 🧭 What and Why

Dependabots is complaining a lot because of `lts` and all its vulnerable deps, we can fix some of them.

Also fix the release CI for swift, it was required to have it like that


shortcuts (commented):
great cleanup gg! can you confirm the patch logic still work with yarn upgrades? D:

shortcuts (dismissed):

shortcuts (approved):
nice!!

PR #3204, opened 2024-06-18
chore(clients): use the ruff python formatter

millotp:
## 🧭 What and Why

The [ruff](https://docs.astral.sh/ruff/) formatter is blazingly fast and reduces the number of deps for python.


shortcuts (commented):
Can we keep the previous ternary syntax? It was more readable

Other than that, looks really great!

millotp:
> Can we keep the previous ternary syntax? It was more readable

I tried too but it's not supported yet, there is a feature request https://github.com/astral-sh/ruff/issues/9722

millotp:
wdyt about the and syntax ? I could replace:
```python
Outliers.from_dict(obj.get("outliers"))  if obj.get("outliers") is not None  else None     
```
with
```python
obj.get("outliers") and Outliers.from_dict(obj.get("outliers"))   
```

shortcuts:
> wdyt about the `and` syntax ? I could replace:
> 
> ```python
> Outliers.from_dict(obj.get("outliers")) if obj.get("outliers") is not None else None     
> ```
> 
> with
> 
> ```python
> obj.get("outliers") and Outliers.from_dict(obj.get("outliers"))   
> ```

implicit None seems better though :/ I guess we can wait for them to support it

Have you tried adding parentheses around the ternaries in the template, just to see if it's correctly indented?

millotp:
> Have you tried adding parentheses around the ternaries in the template, just to see if it's correctly indented?

Indeed that works perfectly !

shortcuts (approved):
Niceee

PR #3209, opened 2024-06-18
fix(go): use generic for nullable types

millotp:
## 🧭 What and Why

Thanks to generics we don't need to generate any `Nullable{{dataType}}` thing and can just have one `Nullable[T]` util which does the exact same thing.


shortcuts (approved):
muuuch cleaner indeed

PR #3212, opened 2024-06-19
fix(specs): update discover endpoint

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-2474

### Changes included:

remove the GET discover endpoint and update the payload for the watch response


millotp (approved):
Great ! It's been a while since those specs have been updated

PR #3213, opened 2024-06-19
fix(scripts): js ci when algoliasearch is missing

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket:

### Changes included:

see https://github.com/algolia/api-clients-automation/pull/3212, js fails to run tests for algoliasearch because it's not supposed to run, we can ignore it when it's the case


millotp (approved):
funny edge case

PR #3214, opened 2024-06-20
fix(specs): ingestion cursor for task

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

leftover of the cursor field


millotp (approved):

Fluf22:
Seems like we are in a bad random streak and we only target failing Apple servers

PR #3216, opened 2024-06-20
fix(go): use createIterable method like in other clients

Fluf22:
## 🧭 What and Why

🎟 JIRA Ticket: [DI-1813]

### Changes included:

- fix typo in Swift client
- add createIterable method

[DI-1813]: https://algolia.atlassian.net/browse/DI-1813?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


templates/go/search_helpers.mustache
```diff
@@ -279,97 +288,116 @@ func (c *APIClient) WaitForApiKeyWithContext(
...
+			slices.Sort(apiKey.Referers)
+			slices.Sort(response.Referers)
+
+			return slices.Equal(apiKey.Referers, response.Referers)
+		}
+	} else {
+		validateFunc = func(response *GetApiKeyResponse, err error) bool {
+			switch operation {
+			case API_KEY_OPERATION_ADD:
+				return err == nil && response != nil && response.CreatedAt > 0
```
shortcuts, line 367: isn't anything other 404 enough? we want to stop if it's invalid credentials or something

shortcuts (commented):
looks super nice, would be great to have struct with Options to better handle default values in the wait helpers but that's for a next iteration!

templates/go/search_helpers.mustache
Fluf22: This hasn't changed from before, it's not in the scope of this PR
Anyway, I think it's covered by `err == nil`

shortcuts: yes but then it means that anything not a 200 wouldn't validate the func, which is true but we don't want to keep polling on that case

can be done in a later PR if you want but that's one line

Fluf22: Ah, I understand what you say
It should be `return err !== nil || response != nil && response.CreatedAt > 0`, then, right?

shortcuts: not sure you can oneline it because it requires the cast to apiError type but to me anything that is not a 404 can validate this

Fluf22: It pains me that we are not doing like in the other clients, but I guess it's Go specificities

Fluf22: Ah but I can't do that, haha
The status code is in the error

shortcuts (approved):
joli

PR #3219, opened 2024-06-21
chore: push to docs before validating release

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket:

### Changes included:

release failed, see https://github.com/algolia/api-clients-automation/actions/runs/9610824715/job/26508506724, but it blocks pushing to the doc, while it shouldn't


millotp (changes requested):
I don't think thats correct, that mean people will see the doc of a version thats not even released yet, I think its better to have the doc outdated that too much updated, and that we should just fix the release if something is broken

shortcuts:
It's not needed anymore anyway

PR #3227, opened 2024-06-21
fix(specs): clarify decompounding limitations

kai687:
## 🧭 What and Why

There was a recent clarification in the old docs to the `decompoundQuery` and `decompoundedAttributes` parameters, that they don't work with some Unicode characters.


millotp (approved):
looks good !

PR #3241, opened 2024-06-24
feat(javascript): legacy `getRecommendations` signature

aymeric-giraudet:
I tested the client with InstantSearch a little bit and some things don't work : 

- `algoliasearch/lite` does not have a default export
- `getRecommendations` does not default `threshold` to `0` like it does in v4, this makes our current implem fail
- Our client hydration logic in InstantSearch doesn't work as the transporter has `baseHeaders` and `baseQueryParamaters` rather than `headers` and `queryParameters`

### DIFF
--- a/specs/recommend/paths/getRecommendations.yml
+++ b/specs/recommend/paths/getRecommendations.yml
@@ -6,6 +6,7 @@ post:
   x-cacheable: true
   x-acl:
     - search
+  x-legacy-signature-recommend: true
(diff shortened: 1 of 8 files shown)


shortcuts (commented):
Wonderful, waiting for the gen to see the output but nothing to add! Thanks :)

shortcuts (approved):
awesome thanks!!

PR #3245, opened 2024-06-24
feat(go): add raw http response methods

Fluf22:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

- add a way to retrieve the raw http response


shortcuts (approved):
can you add tickets for every other clients plssss it's definitely nice to have

PR #3246, opened 2024-06-24
fix(scripts): delete leftover folders from spread scripts

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

it seems like we spread the codegen for every clients even when there's no changes, which led to:
- release commit on the target repo, when there's no releases
- nbGitDiff being always > 1 because of leftover files (deleted in our repo, but not on the target)

let's see with this if it's enough


Fluf22 (approved):
😮

PR #3249, opened 2024-06-25
chore(docs): add doc for the test server

millotp:
## 🧭 What and Why

Document briefly the tests servers, and refactor to split and reuse the express logic.


scripts/cli/index.ts
```diff
@@ -5,7 +5,7 @@ import { buildClients } from '../buildClients.js';
 import { LANGUAGES, setVerbose } from '../common.js';
 import { ctsGenerateMany } from '../cts/generate.js';
 import { runCts } from '../cts/runCts.js';
-import { startTestServer } from '../cts/testServer';
+import { startTestServer } from '../cts/testServer/index.js';
```
shortcuts, line 8: is it required with esm to define the index file? 🤔

scripts/cts/testServer/timeout.ts
```diff
@@ -0,0 +1,26 @@
...
+import { setupServer } from '.';
+
+let timeoutCounter = 0;
+
+export function getTimeoutCounter(): number {
+  return timeoutCounter;
+}
+
+function addRoutes(app: express.Express): void {
+  app.get('/1/test/retry', (req, res) => {
```
shortcuts, line 14: this one is defined in scripts/cts/testServer/gzip.ts also, is it expected?

scripts/cts/testServer/algoliaMock.ts
shortcuts, line 1: maybe rename it to replaceAllObjects explicitly so we will have 1 file per use case later on?

shortcuts (commented):
noice

scripts/cli/index.ts
millotp: Apparently not !

scripts/cts/testServer/timeout.ts
millotp: yes, so that the retry strat can call the same endpoint but with different behavior

millotp: I'll add a comment

shortcuts (approved):
awesome

PR #3250, opened 2024-06-25
fix(go): method in the doc

Fluf22:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

- document the return value of the `WithHTTPInfo` methods


shortcuts (commented):
it will lead to a code change, can you make it a fix commit so that the release and changelog picks it? thanks

shortcuts (approved):
great addition!

PR #3253, opened 2024-06-25
fix(clients): chunked batch helper size property consistency

Fluf22:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

- use batchSize value for slice memory allocation (go)
- add chunkedBatch derivatives with default values (java)


shortcuts (dismissed):
great CATCH

shortcuts (approved):
gg

PR #3255, opened 2024-06-25
fix(specs): virtual is a fetchedIndex props

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket:

### Changes included:

the property introduced in https://github.com/algolia/api-clients-automation/pull/3254 is an index property, not an index settings


Fluf22 (approved):
Small change, big diff

PR #3258, opened 2024-06-26
feat(clients): generate common LICENSE

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-2525

### Changes included:

most clients don't have a LICENSE, we now have one at the root of the template folder with a symlink in each client to generate it


Fluf22 (approved):
Nice

shortcuts:
ah bah the generator don't want symlink either :(

PR #3264, opened 2024-06-27
feat(cts): test the timeouts of the retry stragety

millotp:
## 🧭 What and Why

🎟 JIRA Ticket: [APIC-587](https://algolia.atlassian.net/browse/APIC-587)

Use the test server to verify that the client uses the correct timeouts and delays.

[APIC-587]: https://algolia.atlassian.net/browse/APIC-587?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


scripts/cts/testServer/timeout.ts
```diff
@@ -4,24 +4,89 @@ import type express from 'express';
...
+const timeoutState: Record<string, { timestamp: number[]; duration: number[] }> = {};
 
-export function getTimeoutCounter(): number {
-  return timeoutCounter;
+function aboutEqual(a: number, b: number, epsilon = 100): boolean {
+  return Math.abs(a - b) <= epsilon;
+}
+
+export function assertValidTimeouts(expectedCount: number): void {
+  // assert that the retry strategy uses the correct timings, by checking the time between each request, and how long each request took before being timeouted
```
shortcuts, line 14:
```suggestion
  // assert that the retry strategy uses the correct timings, by checking the time between each request, and how long each request took before being timed out
```

scripts/cts/testServer/timeout.ts
```diff
@@ -4,24 +4,89 @@ import type express from 'express';
...
+          throw new Error(`Expected increasing delay between requests for ${lang}`);
+        }
+        break;
+      case 'PHP':
+        if (!aboutEqual(state.duration[0] * 2, state.duration[1], 200)) {
+          throw new Error(`Expected increasing delay between requests for ${lang}`);
+        }
+        break;
+      default:
+        // the delay should be the same, because the `retryCount` is per host instead of global
```
shortcuts, line 48: shouldn't the default be exponential?

shortcuts: idk what's expected but it feels weird to me to retry instantly

shortcuts (commented):
this is huge

scripts/cts/testServer/timeout.ts
millotp: yeah almost all clients behave like this except for php and js, basically the retryCount is per host so the delay will increase on the next request, not on the next host, so the delay is the same.
If we run a second query in this test, we will see increased delay, but then the test will take too long.

shortcuts: ah so:
1. query search
2. fails, retried on all hosts with no exponential delay between hosts
3. query search again
4. fails, retried with increased delay, still not exponential between hosts

? this seems wrong to me because we assume the user retries but it's the client that should handle all that 🤔

millotp: yeah its weird, but it was like that in legacy clients, this PR aims at checking the logic, not fixing every client as it would be very complicated

shortcuts: yup my comment doens't block it, I'm trying to understand if we should fix that before GA

I think we should ask the search team for advices and harmonize everything

millotp: if we have time yes, it shouldn't be a breaking change to modify this logic

shortcuts (approved):
nothing to add, super clean!

PR #3266, opened 2024-06-27
chore(ruby): scope to last commit

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-2535

### Changes included:

seems like fixes were released on https://github.com/rubygems/release-gem but not tagged, we can try pointing to last commit to see if it works, also that's what they use https://github.com/segiddins/rubygems-await/blob/main/.github/workflows/push_gem.yml, recommended from https://guides.rubygems.org/trusted-publishing/adding-a-publisher/


Fluf22 (approved):
👌🏻

PR #3295, opened 2024-07-01
fix(scripts): do not run notifications on forks

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-2543

### Changes included:

gave an unsuccessful try on https://github.com/algolia/api-clients-automation/pull/3293 at running the notifications steps on forks, tldr it's not possible without leaking secrets, so for now we will just skip the message

also gave a shot at https://github.com/marocchino/sticky-pull-request-comment which removes a lot of custom logic


.github/workflows/check.yml
```diff
@@ -419,6 +440,17 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.ALGOLIA_BOT_TOKEN }}
           PR_NUMBER: ${{ github.event.number }}
 
+      - name: update generation commend
```
Fluf22, line 443:
```suggestion
      - name: update generation comment
```

Fluf22 (approved):

PR #3296, opened 2024-07-01
feat(specs): add redirect to renderingContent

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-2541

### Changes included:

Add the redirect property to the rendering content field, referenced from https://github.com/algolia/algoliasearch-client-go/blob/64232385c99915e8dc51daa1fa4c753c2ab7e97a/algolia/search/rendering_content.go#L5, solves a customer request


morganleroi:
WDYT about adding a custom comment for the `Redirect` property in `SearchResponse` that seems to be an internal property ?

shortcuts:
> WDYT about adding a custom comment for the `Redirect` property in `SearchResponse` that seems to be an internal property ?

good point, done!

morganleroi (approved):
Thanks !

PR #3297, opened 2024-07-01
fix(kotlin): macos compatible helpers and build on CI

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-2535

### Changes included:

build fail on macos https://github.com/algolia/algoliasearch-client-kotlin/actions/runs/9699224976/job/26767673731 but looks good on ubuntu

this is because we use non native features in some helpers, now our CI builds both version so we don't make the same mistake again


.github/workflows/check.yml
```diff
@@ -382,9 +412,9 @@ jobs:
       - client_gen
       - client_gen_javascript
       - swift_cts_macos
+      - kotlin_build_macos
     if: |
       always() &&
-      (needs.swift_cts_macos.result == 'success' || needs.swift_cts_macos.result == 'skipped') &&
```
Fluf22, line 387: Are you sure about this?

shortcuts: Yup it should already be validate by the next clause I think

PR #3303, opened 2024-07-02
fix(scripts): skip failure report for ruby

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-2494

### Changes included:

le ruby release succeed but the ruby release action then lookup on the registry to find the version and somehow fails (but it worked!)

since the GA release approach and we need wait release to succeed in order to spread new versions to the docs, we can skip ruby for now

### DIFF
--- a/scripts/ci/codegen/waitForAllReleases.ts
+++ b/scripts/ci/codegen/waitForAllReleases.ts
@@ -94,7 +94,9 @@ async function waitForAllReleases(languagesReleased: Language[]): Promise<void> 
         console.log(
           `${success ? '✅' : '❌'} ${ciRun.language} CI finished with conclusion: ${ciRun.run.conclusion}`,
         );
-        if (!success) {
+
+        // Ruby action is flaky for now but the release succeed
+        if (!success && ciRun.language !== 'ruby') {
           failures.push(ciRun.language);
         }
         // stop fetching this run.


scripts/ci/codegen/waitForAllReleases.ts
```diff
@@ -94,7 +94,9 @@ async function waitForAllReleases(languagesReleased: Language[]): Promise<void>
         console.log(
           `${success ? '✅' : '❌'} ${ciRun.language} CI finished with conclusion: ${ciRun.run.conclusion}`,
         );
-        if (!success) {
+
+        // Ruby action is flaky for now but the release succeed
+        if (!success && ciRun.language !== 'ruby') {
```
millotp, line 99: I think the fix should be in `clients/algoliasearch-client-ruby/.github/workflows/release.yml` directly, with a small `continue-on-error: true`, that way if something else goes wrong we can still see it

shortcuts: wouldn't it behave exactly like this? the difference here is that on the Ruby repo, it will still fail

millotp: yeah but its wrong on the ruby repo, the release is working, it should also be green there

shortcuts: yes it doesn't depend on us, the action really fails :/ IMO we should find an other GHA but I don't have the time right now and don't want to block the docs

millotp: the action doesn't fail since its released, its just the wait after that fails.
When a users looks at the ruby repo to see what is released it should be green IMO

shortcuts: it fails to lookup if it's correctly released, but this PR isn't meant to be a long term fix, just until we get some bandwidth to properly find an other release action without blocking downstream consumers

millotp: in that case the fix I proposed is even less bandwidth, its only 1 line !

shortcuts: but it's not correct, the action fails so it shouldn't be green, otherwise it might hide other errors that we won't look out for

millotp: anyway we don't agree on where the green should be, for me the action doesn't fail since it works so it should be green

millotp (dismissed):

scripts/ci/codegen/waitForAllReleases.ts
millotp: its just more complexity in this file

shortcuts: ahah I get your point but considering how fragile this action seems to be I'd not be confident enough to hide errors, I feel like this having the red check on that side forces us to have a look that it's **still** that error

millotp: this is why this script exists, to force us to look at errors in child repo, doing the bypass here is the exact same thing, we will never look at the ruby repo again

shortcuts: then in that case, this one shouldn't block subsequent steps and do simple reporting

millotp: I agree it should not block for ruby, by putting the CI green on the ruby repo directly

shortcuts: it's weird because many repos use the rubygems action without complaining I wonder what changes for us, maybe it doesn't support pre-releases

shortcuts: XD just realized it's a single simple file we can patch it maybe https://github.com/rubygems/release-gem/blob/v1/action.yml, would've been faster than this conversation

shortcuts: >   await-release:
    description: "Whether to poll for the release to be available on RubyGems.org"
    required: false
    default: "true"

millotp: ahah nice they know its flaky

shortcuts: do we shake hands now? https://github.com/algolia/api-clients-automation/pull/3303/commits/a879bd00fe61da745da9eb6b5ae47dc2b151caaf

shortcuts: 🤝

millotp (approved):
🤝

shortcuts:
🤝 nice

PR #3307, opened 2024-07-03
chore(ci): use algolia-bot for gen comments

millotp:
## 🧭 What and Why

We already use algolia-bot everywhere else in this repo, IMO it makes sense to also use it here

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -29,7 +29,7 @@ jobs:
 
       - uses: marocchino/sticky-pull-request-comment@v2
         with:
-          GIT_TOKEN: ${{ secrets.ALGOLIA_BOT_TOKEN }}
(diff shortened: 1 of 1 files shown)


shortcuts (dismissed):
good point of using the built-in one is that there's no token to manager nor permission to be careful with, but up to you

.github/workflows/check.yml
```diff
@@ -29,6 +29,7 @@ jobs:
 
       - uses: marocchino/sticky-pull-request-comment@v2
         with:
+          GIT_TOKEN: ${{ secrets.ALGOLIA_BOT_TOKEN }}
```
shortcuts, line 32:
```suggestion
          GITHUB_TOKEN: ${{ secrets.ALGOLIA_BOT_TOKEN }}
```

millotp: oops

Fluf22 (approved):
Why not, it's a valid point, but @shortcuts has a point too

PR #3328, opened 2024-07-08
chore(CTS): benchmark the search method

millotp:
## 🧭 What and Why

Benchmark some methods using the mock server

You can run the benchmarks with `apic cts run -vbrec`


specs/search/helpers/README.md
shortcuts, line 1: good idea

shortcuts: is it reflected on the docs?

tests/CTS/benchmark/search/benchmark.json
```diff
@@ -0,0 +1,35 @@
+[
+  {
+    "testName": "benchmark the search method",
```
shortcuts, line 3: thinking of it, this in fact benchmarks any method of the search client, no? do we expect different outcomes?

shortcuts (commented):
wonderful idea

specs/search/helpers/README.md
millotp: can I link a simple md file to the doc ?

shortcuts: To render it no :/ but you can point to the file indeed with a link

tests/CTS/benchmark/search/benchmark.json
millotp: yes we could test anything, I just picked the search endpoint because it does body serialization and has a lot of props in the body

shortcuts (dismissed):
nothing to add, it looks really nice!!

once we are confident with our testserver implems we should make sure our contributing guide is super complete, just for the sake of maintenance and other contributors, because it keeps growing

shortcuts (approved):
oops good catch

shortcuts (approved):

PR #3330, opened 2024-07-08
chore(ci): cache .build swift dir

millotp:
## 🧭 What and Why

The `.build` folder ca save a lot of time when cached properly


.github/actions/setup/action.yml
```diff
@@ -202,6 +202,14 @@ runs:
       with:
         swift-version: ${{ steps.versions.outputs.SWIFT_VERSION }}
 
+    - name: Cache the build folder
+      id: cache-swift-build
+      if: ${{ inputs.language == 'swift' }}
+      uses: actions/cache@v4
+      with:
+        path: tests/output/swift/.build
+        key: swift-cts-build-${{ steps.versions.outputs.SWIFT_VERSION }}-${{ runner.os }}
```
shortcuts, line 211: should we hash the source folder maybe to invalidate it?

millotp: the swift build system already does that, it only recompiles changed files

shortcuts: ah ok then cooool

shortcuts (approved):
w o n d e r f u l

PR #3331, opened 2024-07-08
chore(ci): remove gradle noise

millotp:
## 🧭 What and Why

When using the tooling properly, with `yarn cli` or `apic`, there is no need for all the extra `gradle` files everywhere, which are not needed.

Make better use of the `setup-gradle` action to reuse the cache from the main branch but avoid recreating one per branch, and do the validation directly instead of using another action.

Also remove the custom yarn cache to try because it seems to be the exact same as the default action

Screenshot 2024-07-08 at 16 24 57


shortcuts (commented):
this looks huge

shortcuts (approved):
so nice, playground still works right?

millotp:
> so nice, playground still works right?

for java yes, I don't think it ever work for kotlin

PR #3332, opened 2024-07-08
chore: cleanup before pushing to docs

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

delete files before pushing to algolia doc, see https://github.com/algolia/doc/pull/9156


millotp (commented):
this looks a big dangerous, nothing indicates that this folder has generated files, anyone can add a file there.
should we rename the folder to `app_data/api/specs/generated/` and then its okay to rimraf ?

millotp (approved):
non blocking for now

PR #3333, opened 2024-07-08
fix(specs): clarify browse parameters

kai687:
## 🧭 What and Why

I felt like [this information](https://www.algolia.com/doc/api-reference/api-methods/browse/#method-param-browseparameters) about the behavior of the `browse` operation is important to document.


shortcuts (approved):
makes total sense

PR #3334, opened 2024-07-08
fix(ruby): use rubyfmt instead of rubocop

millotp:
## 🧭 What and Why

Try [rubyfmt](https://github.com/fables-tales/rubyfmt) instead of rubocop, because it takes more than 2 minutes of the CI of wasted time, and the config is a pain to write, there are too many options.
And it doesn't format properly data structures.

With rubyfmt, its way faster, around 2s for the client, but there is 0 config so we must accept the output, which I find okay


shortcuts (approved):
such a nice find

PR #3336, opened 2024-07-09
fix(scripts): add bundled readme to docs

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

there's no precommit hook on the doc so I won't add overhead to their repo, however we can duplicate the README of the bundled folder there


millotp (approved):
nice

PR #3344, opened 2024-07-10
chore(ci): run mac ci in parrallel

millotp:
## 🧭 What and Why

mac CI for kotlin and swift doesn't generate code on PR so we can run them on parallel as the codegen, but they are required on the main branch because codegen will push to the child repo


shortcuts (approved):
gg! let's try that

PR #3354, opened 2024-07-12
chore: keep guides for docs close to snippets

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket:

### Changes included:

since the split of the CI, the generation of the snippets is done outside of the release step, but the website generated folder isn't pushed to the repo, see https://github.com/algolia/api-clients-automation/actions/runs/9908119385/job/27373697603

since this folder will disappear when the algolia doc leverages our snippets, we can already move our snippet guides in a dedicated folder that we push and keep code in, so the release step can also leverage it


millotp:
I think you need to update `snippet.js` in the website, or remove it entirely

shortcuts:
> I think you need to update `snippet.js` in the website, or remove it entirely

I've ignored it on purpose, it still works but will be removed in a week or so, so... it's fine for now I guess

millotp:
if the files are not generated nor commited it will fail right ? with a `import not found` error

shortcuts:
> if the files are not generated nor commited it will fail right ? with a `import not found` error

since the build docs steps is in the same as the website build I think it's fine, even if it's not committed it still has access to it directly

millotp (approved):
ah yes okay I thought you moved the generated folder too

PR #3376, opened 2024-07-15
chore(scripts): generate clients in parallel

millotp:
## 🧭 What and Why

just a test to see if its faster, and safe to do that.


millotp:
works fine locally but the CI hates it

shortcuts (approved):
that's nice, how come the ci don't like it?

PR #3383, opened 2024-07-17
fix(go): use options for more helpers

millotp:
## 🧭 What and Why

In order to use the functional options pattern for `WaitForApiKey`, I have to move the options to each clients, otherwise we get circular dependencies.
Also add options for `PartialUpdateObjects` and fix playground. 

heavily inspired by https://github.com/uptrace/uptrace-go/blob/master/uptrace/config.go


Fluf22 (approved):
This is great work! Thank you!

PR #3384, opened 2024-07-18
fix(ruby): add more retry error details

otomatik:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DASH-1649

### Changes included:

- adding more details when raising an exception after last retry end in error state

## 🧪 Test

Tested in  this similar PR: https://github.com/algolia/algoliasearch-client-ruby/pull/501


clients/algoliasearch-client-ruby/lib/algolia/transport/transport.rb
```diff
@@ -71,10 +73,14 @@ def request(call_type, method, path, body, opts = {})
...
-          return response unless outcome == RETRY
+          if outcome == RETRY
+            last_retry = {host: host.url, error: response.error}
+          else
+            return response
+          end
         end
 
-        raise Algolia::AlgoliaUnreachableHostError, "Unreachable hosts"
+        raise Algolia::AlgoliaUnreachableHostError, "Unreachable hosts. Last error for #{last_retry[:host]}: #{last_retry[:error]}"
```
kombucha, line 83: Is it not possible to collect all the errors in an array? 👀 
Different hosts might fail for different reasons

Updating the error message is ❤️, but could we also have the underlying errors as an attribute on `Algolia::AlgoliaUnreachableHostError` ? That way you can do more introspection when handling the errors.
Kind of like [cause](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause) in Javascript ^^

Finally, tests ? 🤓 🧪

otomatik: about tests, @millotp do you have recommendations on how to add some since I can see existing ones are generated?

millotp: we don't have a test for this error message, I'll try to add one, but you can test it with the playground if you want, like `yarn cli playground ruby search`

kombucha (dismissed):
LGTM 👍

millotp (approved):

clients/algoliasearch-client-ruby/lib/algolia/transport/transport.rb
millotp: A test has been added in #3405 for this !

PR #3385, opened 2024-07-18
chore(cts): reuse method template

millotp:
## 🧭 What and Why

We can reuse the `method` template in all languages, and for all tests suites, instead of copy pasting it 4 times.

Also simplify the logic and remove unused props, to make it more similar to the requests test suite


scripts/cts/testServer/timeout.ts
```diff
@@ -29,7 +29,7 @@ export function assertValidTimeouts(expectedCount: number): void {
         expect(state.duration[0] * 2).to.be.closeTo(state.duration[1], 200);
         break;
       case 'Swift':
-        expect(state.duration[0]).to.be.closeTo(state.duration[1], 500);
+        expect(state.duration[0]).to.be.closeTo(state.duration[1], 800);
```
shortcuts, line 32:
```suggestion
        expect(state.duration[0]).to.be.closeTo(state.duration[1], 9000);
```

let's preshot

website/docs/contributing/testing/common-test-suite.md
```diff
@@ -140,7 +140,7 @@ When writing your template, here is a list of variables accessible from `mustach
           "method": "the method to call on the API Client",
           "testName": "the descriptive name test (default to `method`)",
           "testIndex": "the index of the test to avoid duplicate function name",
-          "hasParameters": "true if the method has parameters, useful for `GET` requests",
+          "hasParams": "true if the method has parameters, useful for `GET` requests",
```
shortcuts, line 143: makes me think we should use a json schema here instead of json

shortcuts (approved):
soooooooooo clean!!!

website/docs/contributing/testing/common-test-suite.md
millotp: yeah I can give it a try in another PR

PR #3404, opened 2024-07-21
feat(go): expose ExposeIntermediateNetworkErrors to the config

millotp:
## 🧭 What and Why

Fixes an old todo.


Fluf22 (approved):
Should we offer the same option in the other clients?
This could be done later of course

morganleroi (approved):
👍🏼

millotp:
> Should we offer the same option in the other clients? This could be done later of course

maybe, there is already work done for ruby https://github.com/algolia/api-clients-automation/pull/3384

PR #3405, opened 2024-07-21
chore(cts): assert the retry strategy error

millotp:
## 🧭 What and Why

Suggested in #3334, we can add tests for the retry strategy error message, even if its different for every language.


morganleroi (approved):
Nice ! Thanks (a lot) for doing it.

tests/CTS/client/search/api.json
```diff
@@ -87,6 +87,47 @@
...
+        }
+      },
+      {
+        "type": "method",
+        "method": "customGet",
+        "parameters": {
+          "path": "1/test/hang/${{language}}"
+        },
+        "expected": {
+          "error": {
```
shortcuts, line 114: if we add a new language, will this throw? I did not saw this specific case

millotp: yes the CTS will add a default message and the test will fail, so its necessary to add it here

shortcuts (approved):
gg!!!!!!!!

PR #3406, opened 2024-07-21
chore(cts): report benchmark results

millotp:
## 🧭 What and Why

Report benchmark results in the generation comment directly

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -254,6 +254,9 @@ jobs:
         if: ${{ !github.event.pull_request.head.repo.fork && steps.cts-e2e.outcome == 'failure' }}
         run: yarn cli cts run javascript ${{ fromJSON(needs.setup.outputs.JAVASCRIPT_DATA).toRun }} --no-client --no-requests
 
+      - name: Run benchmarks
+        run: yarn cli cts run javascript ${{ fromJSON(needs.setup.outputs.JAVASCRIPT_DATA).toRun }} --benchmark --no-client --no-requests --no-e2e
+
       - name: Generate code snippets for documentation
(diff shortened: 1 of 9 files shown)


.github/workflows/check.yml
```diff
@@ -474,6 +480,25 @@ jobs:
...
 
+      - name: Read benchmark results
+        id: benchmark
+        run: |
+          # merge all benchmark results into a single json, and remove the files
+          results=$(jq -s 'add' -c tests/output/**/benchmarkResult.json)
+          {
+            echo 'BENCHMARK_SECTION<<EOF'
+            echo "<details>"
+            echo "<summary>📊 Benchmark results</summary>"
```
shortcuts, line 491: can we add a small disclaimer like

`take those results with a grain of salt, best effort benchmarking solution, in order to highlight general performances of our Search API`

or something like that? just so people don't expect those to be a source of truth or things like that

shortcuts: also we should mention that it's only tested on search?

shortcuts (commented):
wonderful

.github/workflows/check.yml
millotp: added a small disclaimer, I would like to add more benchmark in the future so I'm not very specific.

.github/workflows/check.yml
```diff
@@ -474,6 +480,26 @@ jobs:
...
+            echo "" # empty line is required to make the table work
```
shortcuts, line 492: do we still need it now that we have a comment?

millotp: need what ?

millotp: Screenshot 2024-07-22 at 14 43 11

yes otherwise it looks like this

shortcuts: ahhh ok make sense

shortcuts (approved):
image

PR #3410, opened 2024-07-23
feat(specs): add v2 endpoints for ingestion

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-2612

### Changes included:

add v2 endpoints for tasks on the ingestion API, mark all v1 as deprecated


specs/ingestion/common/schemas/task.yml
```diff
@@ -1,6 +1,48 @@
...
+    - sourceID
+    - destinationID
+    - enabled
+    - action
+    - createdAt
+
+TaskV1:
+  type: object
+  additionalProperties: false
+  deprecated: true
```
millotp, line 44: maybe we can add all the deprecated in a second time, lets give some time for people to upgrade without having warnings all over the place 😉

shortcuts: it only warns in python, our usage is in JS so I guess that's fine to properly tag the fields

millotp: there are depreciation warning in scala and swift also

shortcuts: I don't think we have to worry about those, the client is still in beta and our consumers rely on typescript so we should be fine

shortcuts:
everything seems good, we can wait for the new endpoints to be on prod to merge this one

millotp (approved):
love it

PR #3411, opened 2024-07-24
fix(specs): rename ingestion list methods

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

There's inconsistencies between [our specs principle](https://api-clients-automation.netlify.app/docs/contributing/docs#operation-summaries), [our return types](https://github.com/algolia/api-clients-automation/blob/main/specs/ingestion/paths/sources/sources.yml#L24) and [our operation names](https://github.com/algolia/api-clients-automation/blob/main/specs/ingestion/paths/sources/sources.yml#L6).

For the better, and before GA, we should align on those names.


morganleroi (approved):
Yes !

PR #3413, opened 2024-07-24
chore(specs): add eslint rule to avoid cross-references

millotp:
## 🧭 What and Why

Suggested by @shortcuts, cross-reference are forbidden because they break our caching and generation strategy, we can add a custom rule to assert that.

The rule needs the list of clients because we allow everything inside each spec, just not with another spec.


eslint/src/rules/refCommon.ts
```diff
@@ -0,0 +1,72 @@
...
+  'query-suggestions',
+  'recommend',
+  'search',
+  'usage',
+];
+
+export const refCommon: Rule.RuleModule = {
+  meta: {
+    docs: {
+      description: 'the $ref must target the current spec, or the common spec',
```
shortcuts, line 22: Maybe make it more informative? E.g. "if you intended to use a model from an other spec, move it to the common folder" or something

shortcuts (commented):
Sooooo niceeeee gg!!

shortcuts (approved):
gggggggg

shortcuts (approved):
gggggg

PR #3414, opened 2024-07-24
fix(specs): remove cross-references of analytics

millotp:
## 🧭 What and Why

Remove the last cross references in the specs


shortcuts (approved):
gg!

PR #3418, opened 2024-07-25
fix(cts): assert the error of chunkedBatch helpers

millotp:
## 🧭 What and Why

This makes sure that the error are thrown correctly and not silenced.
Maybe we should align on the error message.


tests/CTS/client/search/helpers.json
```diff
@@ -220,6 +220,56 @@
...
+          "error": {
+            "csharp": "{\\\"message\\\":\\\"Invalid Application-ID or API key\\\",\\\"status\\\":403}",
+            "go": "API error [403] {\\\"message\\\":\\\"Invalid Application-ID or API key\\\",\\\"status\\\":403}",
+            "java": "Status Code: 403 - {\\\"message\\\":\\\"Invalid Application-ID or API key\\\",\\\"status\\\":403}",
+            "javascript": "Invalid Application-ID or API key",
+            "kotlin": "Client request(POST http://${{localhost}}:6680/1/indexes/cts_e2e_saveObjects_kotlin/batch) invalid: 403 Forbidden. Text: \\\"{\\\"message\\\":\\\"Invalid Application-ID or API key\\\",\\\"status\\\":403}\\\"",
+            "php": "Invalid Application-ID or API key",
+            "python": "Invalid Application-ID or API key",
+            "ruby": "Invalid Application-ID or API key",
+            "scala": "Invalid Application-ID or API key",
```
shortcuts, line 266: Consistency!! almost there

shortcuts (approved):
so nice gg

tests/CTS/client/search/helpers.json
millotp: one of them is a lie, we don't generate helpers tests for scala

PR #3421, opened 2024-07-25
fix(scripts): mention generated is spread msg commit

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

it makes it clearer if we mention that the code is generated


millotp (approved):
beautiful

PR #3424, opened 2024-07-26
chore(deps): remove deprecated packages

millotp:
## 🧭 What and Why

remove the `zlib` deps which has been deprecated for 9 years lol
also fix the peer dependencies to have 0 warnings on yarn finally


shortcuts (approved):
9 years w t f

PR #3425, opened 2024-07-26
chore(scripts): replace tsc by esbuild

millotp:
## 🧭 What and Why

Replace tsc and ncc by esbuild which is 100 times faster, so fast that we can directly put it in the `apic` command and not worry about forgetting to build the scripts.

I also discovered `node --enable-source-maps` which is a game changer, and gives us true typescript stack trace instead of the ugly js


scripts/package.json
```diff
@@ -4,19 +4,19 @@
...
+    "createMatrix": "NODE_NO_WARNINGS=1 node --enable-source-maps dist/ci/githubActions/createMatrix.js",
     "lint": "eslint --ext=ts,js,mjs,cjs .",
     "lint:deadcode": "knip",
     "pre-commit": "node ./ci/husky/pre-commit.mjs",
-    "pushGeneratedCode": "NODE_NO_WARNINGS=1 node dist/scripts/ci/codegen/pushGeneratedCode.js",
-    "pushToAlgoliaDoc": "NODE_NO_WARNINGS=1 node dist/scripts/ci/codegen/pushToAlgoliaDoc.js",
-    "setRunVariables": "NODE_NO_WARNINGS=1 node dist/scripts/ci/githubActions/setRunVariables.js",
-    "spreadGeneration": "NODE_NO_WARNINGS=1 node dist/scripts/ci/codegen/spreadGeneration.js",
-    "start": "NODE_NO_WARNINGS=1 node dist/scripts/cli/index.js",
+    "pushGeneratedCode": "NODE_NO_WARNINGS=1 node --enable-source-maps dist/ci/codegen/pushGeneratedCode.js",
```
shortcuts, line 13: maybe have a command that does `NODE_NO_WARNINGS=1 node --enable-source-maps $*` and call it with the path?

shortcuts (commented):
this is hugeee

shortcuts (dismissed):
huge improvements!!

shortcuts (approved):

PR #3444, opened 2024-07-29
fix(specs): proper title with linter

millotp:
## 🧭 What and Why

The `title` are use by openapi generator to create the name of the model, they need to be set when the model is inlined in the spec, and with some consistency to make them usable in the scripts.


morganleroi (approved):
Who ! Indeed it's better now !

PR #3452, opened 2024-08-01
chore(clients): remove mention of GA

millotp:
## 🧭 What and Why

This statement is not true and is causing confusing internally, we can add it back later when we have more info on the true release state


Fluf22 (dismissed):

shortcuts (commented):
Others have it too no? IIRC i made it the same for every clients

millotp:
I looked in the other readme and those are the only 3 to have the mention of July

Fluf22 (approved):

PR #3456, opened 2024-08-02
fix(specs): dictionary entry for stopwords has type property

kai687:
## 🧭 What and Why

The `Search for dictionary entries` response includes a "type" field for each dictionary entry, that can be "standard" or "custom".


millotp (approved):
looks good !

PR #3457, opened 2024-08-03
fix(clients): move the license to the root

millotp:
## 🧭 What and Why

For most languages, the LICENSE should be at the root, this is preventing the [go doc](https://pkg.go.dev/github.com/algolia/algoliasearch-client-go/v4@v4.0.0-beta.26/algolia/abtesting) from being generated for example: 

Screenshot 2024-08-03 at 08 59 45


shortcuts (approved):
good catch!

PR #3480, opened 2024-08-05
feat(scripts): provide SLA for every existing versions

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-2702

### Changes included:

update the release PR script to generate the SLA of every existing versions that fit in the current 2 year range, which allows us to provide an exhaustive list to the customer and also one source of truth (this repo), that keeps track of legacy clients, and current ones


millotp:
what

Fluf22 (approved):
👌

PR #3485, opened 2024-08-06
fix(scripts): prevent errors on push to docs

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

the script errors if the pull request is already opened, we can delete the branch so that it never fails

also updates the title and add a comment preview link


Fluf22 (approved):
🐛  🔫

PR #3490, opened 2024-08-07
fix(scripts): 2 years sla skips

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

we were missing some versions because of how the 2 year skip was computed, now we should have everything in that 2 year window


Fluf22 (approved):
Nice

PR #3492, opened 2024-08-08
fix(scripts): sla update date

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

we shouldn't update the date if it's the same version being released https://github.com/algolia/api-clients-automation/pull/3491#discussion_r1708837333 (which shouldn't be possible actually, but that's an other problem)


Fluf22 (approved):
🛡️

PR #3497, opened 2024-08-09
feat(specs): recursive snippets and highlights result

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-2723

### Changes included:

this is a regression introduced in the generated clients, it's not as good as what js v4 was proposing (a bit broad) but should still be better

contributes to https://github.com/algolia/api-clients-automation/issues/853#issuecomment-2277263394
closes https://github.com/algolia/algoliasearch-client-javascript/issues/1534


Fluf22 (approved):
Impressive work

PR #3523, opened 2024-08-14
feat(scripts): add script to generate github releases

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-2740

### Changes included:

because not every release processes push github releases, we add a script that does it. it will run once every releases are confirmed working, but can be run manually idempotently


Fluf22 (approved):
Nice addition

PR #3552, opened 2024-08-19
chore: prevent ad block spec files

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-2752

### Changes included:

ad blocker extensions prevent the analytics spec from rendering, we need to provide a different name when pushing the doc to prevent that


millotp (approved):
that's funny

PR #3560, opened 2024-08-20
chore: prevent cancel on release commits

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

we can prevent cancelling CI runs of release commit as recommended here https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs#example-only-cancel-in-progress-jobs-on-specific-branches


.github/workflows/check.yml
```diff
@@ -11,7 +11,7 @@ on:
 
 concurrency:
   group: ${{ github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: ${{ !contains(github.event.head_commit.message, 'prepare release')}}
```
shortcuts, line 14: looking at https://github.com/algolia/api-clients-automation/actions/runs/10471874630/job/29000072705#step:2:315 this should work, lmk if you have an other idea

PR #3568, opened 2024-08-21
chore(specs): breaking change method names mapping table

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-2782

### Changes included:

this pr provides a json file to be used as a source of truth for the documentation to generate a common mapping table of method renames with its associated documentation code snippet link


specs/major-breaking-changes-rename.json
```diff
@@ -0,0 +1,88 @@
+{
+  "abtesting": {},
+  "analytics": {},
+  "ingestion": {},
+  "insights": {},
```
millotp, line 5: maybe add it for insights too ? they should all have `pushEvents`

millotp (commented):
good idea !

specs/major-breaking-changes-rename.json
shortcuts: We did not had an insights client before so no bc, no?

millotp: ah maybe I'm confusing with another lib

millotp (approved):

PR #3572, opened 2024-08-22
chore(scripts): use markdown in jira

millotp:
## 🧭 What and Why

Jira tickets are very ugly because the rest api v3 only support ADF, which is not compatible with markdown.
We can use version 2 of the Jira API that supports md.


shortcuts (approved):
Might be much cleaner indeed!

Fluf22 (approved):
Great idea, thanks!

PR #3573, opened 2024-08-22
fix(scripts): dont fail on empty benchamrk

millotp:
## 🧭 What and Why

Ignore the benckmark report when the matrix is empty.
Also fix the common templates.


Fluf22 (approved):
:ok:

shortcuts (approved):
Joli

PR #3596, opened 2024-08-25
fix(javascript): dependency imports and low-level init

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-2828

### Changes included:

closes https://github.com/algolia/algoliasearch-client-javascript/issues/1538

we import from the src of the dependency which is fine in the monorepo context, but the path isn't correct when importing from the built package

we actually don't need the low level init method, we can just use the regular instance and spread it, it should behave the same


clients/algoliasearch-client-javascript/base.rollup.config.js
```diff
@@ -276,6 +276,7 @@ export function buildConfigs(pkg) {
           }),
           nodeResolve({
             preferBuiltins: true,
+            browser: isUmdBuild || isEsmBrowserBuild,
```
shortcuts, line 279: see https://www.npmjs.com/package/@rollup/plugin-node-resolve#browser, I should've set that way before D:

templates/javascript/clients/algoliasearch/builds/initClients.mustache
```diff
@@ -1,61 +1,15 @@
 function initRecommend(initOptions: InitClientOptions = {}): RecommendClient {
-  return createRecommendClient({
-    ...commonOptions,
-    ...initOptions.options,
-    ...initOptions,
-  });
+  return recommendClient(initOptions.appId || appId, initOptions.apiKey ||apiKey, initOptions.options);
```
shortcuts, line 2: this is redundant and I think was legacy, now that standalone clients checks it, we don't need it here

clients/algoliasearch-client-javascript/base.rollup.config.js
shortcuts: ok it's not fully working

millotp (approved):
looks good

PR #3598, opened 2024-08-26
chore(scripts): add exec command

millotp:
## 🧭 What and Why

Add the `exec` command to the CLI, to execute a command inside of docker with the correct image.

Also fix the completion script to be more correct on all commands.


shortcuts (approved):
long time no see my friend

PR #3603, opened 2024-08-28
Fix/java cmp

millotp:
## 🧭 What and Why

The java package was renamed last year but not in our code, it's time to fix that and to make the japicmp mandatory to avoid BC.


shortcuts (approved):

shortcuts (approved):
japi japi japo

PR #3606, opened 2024-08-28
chore: add a bug report template to github

millotp:
## 🧭 What and Why

Add a common bug report to all child repo, specifying that the code is generated because its not obvious and people don't know where to contribute.


templates/Bug_report.yml
```diff
@@ -0,0 +1,58 @@
...
+labels: ['bug', 'triage']
+body:
+  - type: markdown
+    attributes:
+      value: |
+        ## Please help us help you!
+
+        > [!IMPORTANT]  
+        > This repository is fully generated and all pull request will be rejected.
+        > If you wish to contribute, please refer to the [contribution guidelines](https://github.com/algolia/api-clients-automation/blob/main/CONTRIBUTING.md) on the `api-clients-automation` repository.
```
shortcuts, line 13: can you add a statement to eventually ask people to re-open issue here? at some point, we will close opening issues on those repos

shortcuts (commented):
nice

Fluf22 (approved):
Great idea

templates/Bug_report.yml
Fluf22: I'm not sure it will even become obvious why we prevent opening issues on those repo
I think it's on us to redirect from each repo to the mother one

shortcuts: as long as it's in the readme it's fine, but that's a discussion for later

we should still recommend people to open issue instead as that's kind of the goal of a monorepo

millotp: I'm not sure about that, I find it nice to have the issues on each repo, since we have good tooling to report and track them, it's easier for people

millotp: maybe this repo issues should only concern the generation and tooling, not the API clients code directly.

Fluf22: I agree with @millotp , sorry @shortcuts 
I find it clearer to have issues on the separate repo. Put yourselves in the shoes of a user: if you want to check wether a trouble you encounter already exists or not, you go to the client repo. If you are cool, you will see in the README (why would you even bother to do that?) that it's generated, and that you have to go see another repo. And there, there would be hundreds of non-related issues, or worse, some that would match your search query, but not your language.
It would be like going on a big GitHub repo to look across all existing issues for one specific thing

millotp: we can always iterate on this, I think this is a good start and if we see that people should contribute more to this repo we can change the message

shortcuts: issue labels are especially useful for that, also it helps deduping issues, e.g. if something is reported on API x on client y, people using other client z would still find that issue if it occurs on theirs

shortcuts: > we can always iterate on this, I think this is a good start and if we think that people should contribute more to this repo we can change the message

yes that's what I suggest here https://github.com/algolia/api-clients-automation/pull/3606#discussion_r1734739708, having a good issue template is still a good short term solution

millotp: > at some point, we will close opening issues on those repos

I don't agree with this statement, people are not supposed to care for other languages, they just want a simple way to report the bug where they use it

shortcuts: they don't have to care about other language, but an api/typing issue is different than a client issue, still they are both useful for a user and could be reported for an other client than theirs

millotp: In that case I already mention this repo, I don't think we need more

millotp: if you feel otherwise you can always add a commit on this PR !

millotp: no ahah thats not a good idea

shortcuts: it's just informative

shortcuts (approved):
SHORT TERM OK???????????????????

gg

PR #3608, opened 2024-08-28
fix(javascript): prefix with `node:` for node packages

atinux:
## 🧭 What and Why

It is a best practice to prefix node packages with `node:`

Introduced in Node 18: https://stateful.com/blog/node-18-prefix-only-modules

This also ease support for edge-runtime like Cloudflare Workers (https://developers.cloudflare.com/workers/runtime-apis/nodejs/crypto/)

### Changes included:

- Update import from `crypto` to `node:crypto`


shortcuts:
thanks a lot for the contribution and the insight ❤️

shortcuts (approved):
wonderful

PR #3612, opened 2024-08-29
chore(ci): add ability to skip e2e when something is down [skip-e2e]

millotp:
## 🧭 What and Why

Sometimes server are down, it shouldn't be blocking


shortcuts:
huge!

.github/workflows/check.yml
```diff
@@ -253,7 +253,7 @@ jobs:
       - name: Run e2e CTS
         id: cts-e2e
         continue-on-error: true
-        if: ${{ !github.event.pull_request.head.repo.fork }}
+        if: ${{ !github.event.pull_request.head.repo.fork && !contains(github.event.pull_request.title, '[skip-e2e]') }}
```
shortcuts, line 256: i wonder if  we need to check if we are on a pr or not?

shortcuts: looking at https://github.com/algolia/api-clients-automation/actions/runs/10605323049/job/29393867325#step:2:1 pull request doesn't exist on a main context, so we need to fallback otherwise it will not skip on main either

millotp: done in https://github.com/algolia/api-clients-automation/pull/3612/commits/6952b1c6ad6972ce8bcc7aef841f2a2e61b9a021

.github/workflows/check.yml
```diff
@@ -330,7 +330,7 @@ jobs:
         run: yarn cli build playground ${{ matrix.client.language }}
 
       - name: Run Java 'algoliasearch' public API validation
-        if: ${{ matrix.client.language == 'java' }}
+        if: ${{ matrix.client.language == 'java' && !contains(github.event.pull_request.title, '[skip-bc]') }}
```
shortcuts, line 333: this also need the main cond, no?

millotp: ah true

shortcuts (approved):
bravo

PR #3620, opened 2024-08-30
fix(specs): ignorePlurals can be a boolean string

millotp:
## 🧭 What and Why

Closes https://github.com/algolia/algoliasearch-client-go/issues/610, turns out boolean string are supported here.
Also add e2e test for `getRule` and `searchRules`, and also for unicode at the same time.

Delete all the `analytics` e2e because the CI won't do it for us.


specs/common/schemas/IndexSettings.yml
```diff
@@ -1168,3 +1169,9 @@ supportedLanguage:
     - ur
     - uz
     - zh
+
+booleanString:
+  type: string
+  enum:
+    - true
+    - false
```
shortcuts, line 1177: does it need to be quoted?

tests/CTS/requests/search/searchRules.json
```diff
@@ -1,16 +1,52 @@
...
+            ],
+            "consequence": {
+              "params": {
+                "ignorePlurals": "true"
+              },
+              "filterPromotes": true,
+              "promote": [
+                {
+                  "objectIDs": [
+                    "\u00C6on Flux"
```
shortcuts, line 36: ahhhhhh

shortcuts (commented):
looks great! waiting for the gen

shortcuts (approved):
GG!

PR #3619, opened 2024-08-30
docs(insights): Example event name should change with eventType

JonathanMontane:
## 🧭 What and Why

Looking at the spec for `View Filters` and seeing `example name: Product Clicked` is confusing.

image

🎟 JIRA Ticket: None

### Changes included:

- Added examples to the eventName attribute of each concretization of the event concept
- Removed based example on eventName attribute (to keep the spec json schema compliant)

## 🧪 Test

?? <- Let me know how I can do that! I ran `yarn website` to generate the bundled version and it seemed okay, but 🤷‍♂️ and I don't know how to view the website locally 😅

PR #3623, opened 2024-08-30
fix(specs): don't extend $ref objects

kai687:
## 🧭 What and Why

PR #3619 added `eventName` examples for different event objects in the Insights API, instead of using "Product clicked" for all events.

But, extending/overriding the `example` property isn't supported by the OpenAPI spec, and is ignored when rendering the docs. (tested locally with the redocly CLI). 

This PR removes the examples. We provide full examples for each event in the right column, so an extra example property of the event name isn't really necessary anyway.

If I understand the specs correctly:

- in OpenAPI 3.0.3, _"This object cannot be extended with additional properties and any properties added SHALL be ignored."_ 
- in OpenAPI 3.1.0, we can override `summary` and `description` properties of the referenced object, but still not `example`.

[OpenAPI 3.1.0](https://spec.openapis.org/oas/v3.1.0.html#reference-object)
[OpenAPI 3.0.3](https://spec.openapis.org/oas/v3.0.3.html#reference-object)

🎟 JIRA Ticket:

### Changes included:

- Remove `eventName` examples


shortcuts (commented):
Should we remove the product clicked example of the eventName ref then? This should avoid confusion

shortcuts:
> @shortcuts PR 3619 removed the `example` from the `eventName` prop, so it's not part of the spec anymore.

Ah indeed bah nice, cc @JonathanMontane just fyi

shortcuts (approved):

JonathanMontane:
Even better then!

PR #3640, opened 2024-08-31
chore(javascript): use tsup bundler

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket:

### Changes included:

closes https://github.com/algolia/algoliasearch-client-javascript/issues/1544

use tsup instead of rollup, just because I wanted to try it out and it seems to exactly do what we need, a bit faster and easier to configure.

also add attw and publint, in order to debug the exported types, package.json#export field, and the resolved imports.

- [x] use tsup (chore)
- [x] setup attw and publint (helps with the fix of https://github.com/algolia/algoliasearch-client-javascript/issues/1544)
- [x] update commons (was outdated)
- [x] provide node and browser bundle for clients (same as with rollup, but now it works again)
- [x] provide node bundle for fetch requester (feat)
- [x] use esm mini output from tsup into rollup for umd

## how to verify it works?

- generate and builds the js clients
- go to clients/algoliasearch-client-javascript
- run `test:size`, should be all green without being too far from the threshold
- run `test:bundle` on many clients, they are all green
- try the `5.2.4-beta.2` directly


.github/workflows/check.yml
```diff
@@ -226,6 +226,10 @@ jobs:
         if: ${{ startsWith(env.head_ref, 'chore/prepare-release-') }}
         run: cd clients/algoliasearch-client-javascript && yarn test:size
 
+      - name: Test JavaScript bundle and types
+        if: ${{ startsWith(env.head_ref, 'chore/prepare-release-') }}
+        run: cd clients/algoliasearch-client-javascript && yarn test:bundle
```
millotp, line 231: lovely

clients/algoliasearch-client-javascript/base.tsup.config.ts
```diff
@@ -0,0 +1,40 @@
...
+    external: [...getDependencies(pkg), 'node:crypto'],
+  };
+}
+
+export function getBaseBrowserOptions(pkg: PKG, cwd: string): Options {
+  return {
+    ...getBaseConfig(cwd),
+    platform: 'browser',
+    format: ['esm'],
+    target: ['chrome109', 'safari15.6', 'firefox115', 'edge126'],
```
millotp, line 37: ie is also a valid target, don't you want to include it ?

clients/algoliasearch-client-javascript/package.json
```diff
@@ -12,7 +12,8 @@
     "release:bump": "lerna version ${0:-patch} --no-changelog --no-git-tag-version --no-push --exact --force-publish --yes",
     "release:publish": "tsc --project scripts/tsconfig.json && node scripts/dist/scripts/publish.js",
     "test": "lerna run test $*",
-    "test:size": "bundlesize"
+    "test:size": "bundlesize",
+    "test:bundle": "lerna run test:bundle --verbose --skip-nx-cache --include-dependencies"
```
millotp, line 16: why the skip cache everywhere ? on the build step too ?

templates/javascript/clients/algoliasearch/lite.d.mustache
```diff
@@ -1,2 +1,2 @@
 // eslint-disable-next-line import/no-unresolved
-export * from './dist/lite/builds/node';
+export * from './dist/lite/node';
```
millotp, line 2: can you delete this template and just create the d.ts file directly pls ?

templates/javascript/clients/algoliasearch/lite.mustache
```diff
@@ -1,2 +1,2 @@
 // eslint-disable-next-line import/no-commonjs,import/extensions
-module.exports = require('./dist/lite/lite.cjs');
\ No newline at end of file
+module.exports = require('./dist/lite/builds/node.cjs');
```
millotp, line 2: same

templates/javascript/clients/index.d.mustache
```diff
@@ -1,2 +1,2 @@
 // eslint-disable-next-line import/no-unresolved
-export * from './dist/{{#isAlgoliasearchClient}}algoliasearch/{{/isAlgoliasearchClient}}builds/node';
+export * from './dist/node';
```
millotp, line 2: same

millotp (commented):
this is huuuge, congrats !

clients/algoliasearch-client-javascript/base.tsup.config.ts
shortcuts: The enum for target are prefixes actually, if you don't give a precise version for browsers it asks for it at runtime

clients/algoliasearch-client-javascript/package.json
shortcuts: nx and lerna was still in an early phase when I set it up and there wad cache issues, I guess it's legacy but we should be able to remove it as on the ci there's always no cache

templates/javascript/clients/algoliasearch/lite.d.mustache
shortcuts: This allows adding a new client without touching the js repo normally, if we remove those it means documenting to add them etc... it's easier that way

clients/algoliasearch-client-javascript/package.json
millotp: we could cache the .nx folder on the CI

shortcuts: idk how flaky it is but yup maybe

templates/javascript/clients/algoliasearch/lite.d.mustache
millotp: some files already need to be copied manually, I think we will just copy the package if we add a new spec

shortcuts: > some files already need to be copied manually

true but ideally you'd just add your spec and that's it

millotp (approved):
nothing to add, gg !

xPaw:
After this release, the umd build lite client (`dist/lite/builds/browser.umd.js`) is being exported as `window.lite.liteClient`, but before it was being exported as ` `window[ 'algoliasearch/lite' ].liteClient`, this does not seem ideal.

shortcuts:
indeed @xPaw and this shouldn't have changed (created an issue https://github.com/algolia/api-clients-automation/issues/3700), I'll try to see if we can provide both

PR #3641, opened 2024-09-01
fix(csharp): correct version in user agent [skip-e2e]

millotp:
## 🧭 What and Why

The version in the Csharp user agent is always equals to `0.0.0.0`, which makes it hard to see if the client is being used or not.


Fluf22 (approved):
Nice!
Curious to see if other clients have this defect 😬 (Swift 👀)

morganleroi (dismissed):

shortcuts (approved):
ggggg

PR #3643, opened 2024-09-02
fix(specs): no links across API specs

kai687:
## 🧭 What and Why

The "Delete QS config" endpoint had a link to the Search API's "Delete index" endpoint.
Those links aren't portable (e.g. they worked for the Docusauraus site, but don't work in the Algolia docs and likely won't work somewhere else). Therefore, it's better not to have links in the specs to other Algolia APIs.

### Changes included:

- Remove link in QS API description
- Update contribution guide to avoid links across APIs


Fluf22 (approved):
✅

millotp (approved):

PR #3649, opened 2024-09-04
fix(specs): note on insights API server

kai687:
## 🧭 What and Why

Correct a false statement about the Insights API's server `insights.algolia.io`. It uses geo routing to forward to either `insights.de.algolia.io` or `insights.us.algolia.io`.

### DIFF
--- a/specs/insights/spec.yml
+++ b/specs/insights/spec.yml
@@ -16,8 +16,9 @@ info:
 
     The base URLs for making requests to the Insights API are:
 
-    - `https://insights.us.algolia.io` (`https://insights.algolia.io` is an alias)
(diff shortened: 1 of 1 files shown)


shortcuts:
Ah yes indeed, it's the same for other servers like https://github.com/algolia/api-clients-automation/blob/main/specs/analytics/spec.yml#L72 and https://github.com/algolia/api-clients-automation/blob/main/specs/abtesting/spec.yml#L71

shortcuts:
ok @millotp confirmed it's just an alias

Fluf22 (approved):
Thanks for confirming this!

PR #3650, opened 2024-09-04
fix(specs): note on Analytics API server

kai687:
## 🧭 What and Why

You are too quick. The Analytics API also uses geo-routing. Follow up of #3649.


shortcuts (approved):
cc @millotp you liar

PR #3654, opened 2024-09-04
chore(ci): automaticaly create jira issues on gh issue

millotp:
## 🧭 What and Why

Since it's working well on all repo, we can add the action here too.


Fluf22 (approved):
👌

PR #3679, opened 2024-09-09
chore(ci): fallback step name when they are skipped

millotp:
## 🧭 What and Why

because pretty name are better


shortcuts (approved):
cute

PR #3680, opened 2024-09-09
feat(javascript): add workerd exports

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-2890 https://algolia.atlassian.net/browse/DI-2895

### Changes included:

closes https://github.com/algolia/algoliasearch-client-javascript/issues/1548

this pr contains a few changes in few lines, sorry for the mess, but it's all related.

## `fetch` build

in order to improve our dx, we can provide a `fetch` build output, as well as a `worker` exports (which [should also work for the cloudflare workers](https://developers.cloudflare.com/workers/wrangler/bundling/#conditional-exports)). this allows the user to directly load the client rather than having to manually setup fetch.

## manual testing

the manual testing is just a harness since the dependency clients are already tested, but in order to remove redundancy, common cases are moved to the `algoliasearch.common.test.ts` file.

## worker testing

cloudflare workers provide a [miniflare](https://miniflare.dev/) instance for testing, pretty easy to setup, that at least allows us to ensure it's properly starting when loading the fetch client.


clients/algoliasearch-client-javascript/base.tsup.config.ts
```diff
@@ -16,35 +19,44 @@ export function getBaseConfig(cwd: string): Options {
...
     return deps;
   }
 
-  if (env === 'node') {
-    return deps.filter((dep) => dep !== '@algolia/requester-browser-xhr');
+  switch (requester) {
+    case 'http':
+      return deps.filter((dep) => dep !== '@algolia/requester-browser-xhr');
+    case 'xhr':
+      return deps.filter((dep) => dep !== '@algolia/requester-node-http');
```
millotp, line 33: this is swapped

millotp: ah no its a !==, so it returns 2 requester ?

shortcuts: fetch is never a dependency, so it will return one or the other

clients/algoliasearch-client-javascript/packages/client-common/src/createEchoRequester.ts
```diff
@@ -1,15 +1,21 @@
 import type { EchoResponse, EndRequest, Requester, Response } from './types';
 
+type BasicURL = {
```
shortcuts, line 3: propose names if you want, it's mostly to avoid importing URL from anywhere

templates/javascript/clients/algoliasearch/builds/definition.mustache
```diff
@@ -21,9 +21,9 @@ import type {
 export * from './models';
 
 export type Algoliasearch = SearchClient & {
-    initRecommend: (initOptions: InitClientOptions)=> RecommendClient;
-    initAnalytics: (initOptions: InitClientOptions & InitClientRegion<AnalyticsRegion>)=> AnalyticsClient;
-    initAbtesting: (initOptions: InitClientOptions & InitClientRegion<AbtestingRegion>)=> AbtestingClient;
+    initRecommend: (initOptions?: InitClientOptions)=> RecommendClient;
```
shortcuts, line 24: it forced us to pass an empty object, everything in here is partial

templates/javascript/clients/client/builds/fetch.mustache
```diff
@@ -0,0 +1,26 @@
...
+    ...create{{#lambda.titlecase}}{{apiName}}{{/lambda.titlecase}}({
+      appId,
+      apiKey,{{#hasRegionalHost}}region,{{/hasRegionalHost}}
+      timeouts: {
+        connect: DEFAULT_CONNECT_TIMEOUT_NODE,
+        read: DEFAULT_READ_TIMEOUT_NODE,
+        write: DEFAULT_WRITE_TIMEOUT_NODE,
+      },
+      algoliaAgents: [{ segment: 'Fetch' }],
+      requester: createFetchRequester(),
```
shortcuts, line 16: only two lines that differ from the node implem

templates/javascript/clients/client/builds/fetch.mustache
```diff
@@ -0,0 +1,26 @@
...
+      responsesCache: createNullCache(),
+      requestsCache: createNullCache(),
+      hostsCache: createMemoryCache(),
+      ...options,
+    }),
+    {{#nodeSearchHelpers}}
```
shortcuts, line 22: crypto is pollyfilled on cloudflare workers, so all good for now

millotp (approved):
truth is, I don't understand any of this, gg !

Haroenv (approved):
the default of cjs may cause issues but it's probably ok

shortcuts:
> the default of cjs may cause issues but it's probably ok

for the worker export? I've defaulted it to esm

shortcuts:
I removed it in the end, it's never used when import and require is defined (thanks tsup for highlighting that D:)

PR #3682, opened 2024-09-09
fix(javascript): use transporter directly

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket:

### Changes included:

using `this` forces is to bind the client, we can avoid it by directly using the transporter


Fluf22 (approved):
Validated ✅

morganleroi (approved):
🙌🙌🙌

PR #3686, opened 2024-09-10
chore(scripts): improve release script with git diff

millotp:
## 🧭 What and Why

For the longest time, we based changelogs and release on the commit title only, but its often misleading or wrong.
Instead, we can use `git diff` to know exactly which commits impact which client, and based the release and changelog based of that.

Also take the time to refactor and simplify the release script and lint the tests.
The changelog will automatically use the correct language scope also, even if the commit was wrong.


scripts/release/createReleasePR.ts
```diff
@@ -401,32 +330,30 @@ export async function createReleasePR({
...
     const changelogCommits: string[] = [];
     for (const validCommit of validCommits) {
-      if (validCommit.scope !== lang && !COMMON_SCOPES.includes(validCommit.scope)) {
+      if (!validCommit.languages.includes(lang)) {
         continue;
       }
 
+      // sometimes the scope of the commits is not set correctly and concerns another language, we can fix it.
+      if (LANGUAGES.includes(validCommit.scope as Language) && validCommit.scope !== lang) {
+        validCommit.message = validCommit.message.replace(`(${validCommit.scope}):`, `(${lang}):`);
```
shortcuts, line 354: is this reflected in the changelog? I'd rather keep it matching the git history but still include it in the new version computing if possible

scripts/release/createReleasePR.ts
```diff
@@ -196,78 +177,33 @@ export function getNextVersion(current: string, releaseType: semver.ReleaseType
...
-  languages,
   releaseType,
-  dryRun,
 }: {
-  versions: VersionsBeforeBump;
-  commits: PassedCommit[];
-  languages: Language[];
+  commits: ParsedCommit[];
   releaseType?: semver.ReleaseType;
   dryRun?: boolean;
```
shortcuts, line 186: this is unused no? it would be great to stop before the PR/branch related stuff

shortcuts (commented):
nice clean!!

scripts/release/__tests__/createReleasePR.test.ts
```diff
@@ -858,15 +441,6 @@ describe('createReleasePR', () => {
...
         - abcdefg fix: somethin2
-        </details>
-
-        <details>
-          <summary>
-            <i>Commits with unknown language scope:</i>
-          </summary>
-
-          - abcdef2 fix(pascal): what
-        - abcdef2 fix(pascal): what is that
```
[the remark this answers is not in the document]
millotp: yes its now impossible to not know the language, since it's determined by git diff, either the change is in a client folder or its not and in that case the commit is not included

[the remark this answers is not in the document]
shortcuts: ok ok, will we still have a list of commits that are not included in the computing of the release? if so, all good!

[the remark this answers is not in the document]
millotp: yeah they all go in the `missing language scope` thing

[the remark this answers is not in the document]
shortcuts: cooool

scripts/release/createReleasePR.ts
shortcuts: ah no I was wondering about the dryRun because I did not saw the line where it was used, great then!

scripts/release/createReleasePR.ts
millotp: I find it weird to see a `feat(python): add linter` in the java changelog

shortcuts: can we replace it with `feat(clients)` maybe then?

millotp: sure

shortcuts: just make the scope wider idk D: it's weird to differ from the git history imo but you are right

shortcuts (approved):

PR #3688, opened 2024-09-11
feat(specs): add (optional) _automaticInsights to search result

raed667:
## 🧭 What and Why

Return the parameter `_automaticInsights` (when possible) when calling search.

# Why 

I need this value in the Recommend mono-repository so that the Recommend API can forward this parameter to the Recommend client.


millotp (approved):
looks good !

PR #3691, opened 2024-09-11
fix(specs): remove usage [skip-bc]

kai687:
## 🧭 What and Why

Adding a tag group to the Usage API spec, so that the custom POST/PUT/DEL requests won't show up in the API reference.


shortcuts (dismissed):
Nice thanks

PR #3692, opened 2024-09-11
fix(specs): recommend remove unsupported query parameters [RECO-2345] [skip-bc]

raed667:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/RECO-2345

### Changes included:

- Remove unsupported parameters: `page, hitsPerPage, offset, length`
- Comment out unknown parameters until they're supported: `customRanking, semanticSearch, mode, keepDiacriticsOnCharacters`


millotp:
hey @raed667, do you need help to make the CI pass ?

raed667:
@millotp thanks! i'm currently gathering a bit more information about the API, i'll let you once i start updating the PR again

playground/dart/lib/recommend.dart
```diff
@@ -15,7 +15,7 @@ void main() async {
     RelatedProducts(
       model: RelatedModel.relatedProducts,
       objectID: '6445156',
-      fallbackParameters: FallbackParams(
+      fallbackParameters: RecommendSearchParams(
```
raed667, line 18: 🤔 Would this be a breaking change ?

Was the rename from `fallbackParams` to `recommendSearchParams` due to some sort of clash in naming or a due to something else   ?

shortcuts: no actually I'm trying out why it fails on the CI but not locally, I'll revert that to the old name for this specific object, but I think there would be a name conflict on the search params still, if we keep it to searchParams, when it lands on `algoliasearch`

shortcuts: since algoliasearch encapsulates many clients, it imports their type, if we have searchParams (on search) and searchParams (on recommend) with different signature, it won't be able to re-export them :/

shortcuts: fallbackParams we can keep it since it only exists there

millotp (approved):
looks good !

PR #3695, opened 2024-09-11
chore(scripts): take generated commits into account for the release

millotp:
## 🧭 What and Why

I forgot about generated commits in my last PR, they should also contribute to the changelog, on behalf of the base commit.


Fluf22 (dismissed):
✅

scripts/release/createReleasePR.ts
shortcuts, line 1: can you add the default scope to `client` when it doesn't match? you know the comment from the previous pr

shortcuts (approved):

PR #3696, opened 2024-09-12
fix(specs): update analytics and abtesting max pagination query limit

febeck:
## 🧭 What and Why

Analytics team is enforcing the limit where combination of `limit` and `offset` can only target the top 1000 rows.
Documentation updated to reflect those updates

🎟 JIRA Ticket: [OPTIM-1255]

### Changes included:

`Limit` and `Offset` were previously defined in a global `common` folder. Usages were made in analytics and one single usage in AB test.
To avoid any confusion
- move usage with improved specificity about limit and offset into analytics
- define single usage in AB test API locally
- remove no-longer definition from common folder

## 🧪 Test

[OPTIM-1255]: https://algolia.atlassian.net/browse/OPTIM-1255?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


shortcuts (approved):
looks good! thanks for the fix :D

shortcuts:
should we wait for someone from your team to review as well, or is it good to go?

cdhawke:
@shortcuts we need to wait - we had to revert the change and i'll approve when we get it redeployed

febeck:
Thanks for the review @shortcuts . As @cdhawke  said, this change has been reverted and therefore I put the PR in draft

millotp (approved):

PR #3697, opened 2024-09-12
chore(ci): retry flaky tasks

millotp:
## 🧭 What and Why

Some jobs on the CI fails a lot, for different reasons, we can add a bit of retry for the 2 macOS jobs, and for the release because maven is often down, or kotlin has a weird failure.


.github/workflows/check.yml
```diff
@@ -430,13 +430,14 @@ jobs:
           type: minimal
           language: kotlin
 
-      - name: remove previous build from linux
```
millotp, line 433: this didn't help

.github/workflows/check.yml
```diff
@@ -622,7 +630,7 @@ jobs:
 
   push_and_release:
     runs-on: ubuntu-22.04
-    timeout-minutes: 15
+    timeout-minutes: 30
```
shortcuts, line 633: omgwtf

millotp: kotlin is taking 12 min to deploy and often fails, so weed need 30 min when it retries

shortcuts (approved):
gg nothing to add

Issue #3698, opened 2024-09-12
[bug]: v4+ Filters don't work using SearchClient::searchSingleIndex()


aykutersoy:
### Description

We're using v4.4.0 and filters don't seem to work when using `SearchClient::searchSingleIndex()`

Either I'm doing something wrong or filters in query parameters are not taken into account when searching (Please see below reproducible scenario)

I've also tried this but because `searchSingleIndex` create bodyless requestOptions below body attribute just gets overridden. 

```
...
$searchResponse = $client->searchSingleIndex(
    indexName: $indexName,
    searchParams: ['query' => ''],
    requestOptions: [
        'body' => [
            'filters' => 'value=1',
        ],
    ],
);
...
```

### Language

PHP

### Client

Search

### Steps to reproduce

```
<?php
// File: hello_algolia.php
require_once realpath(__DIR__ . "/vendor/autoload.php");

use Algolia\AlgoliaSearch\Api\SearchClient;
$appID = "ALGOLIA_APPLICATION_ID";
// API key with `addObject` and `search` ACL
$apiKey = "ALGOLIA_API_KEY";
$indexName = "test-index";

$client = SearchClient::create($appID, $apiKey);

// Create a new record
$record = [
    "objectID" => "object-1",
    "name" => "test record",
    "value" => "1",
];
$record2 = [
    "objectID" => "object-2",
    "name" => "test record 2",
    "value" => "2",
];

// Add the record to an index
$client->saveObject($indexName, $record);
$client->saveObject($indexName, $record2);

// Search for ''
$searchResponse = $client->searchSingleIndex(
    indexName: $indexName,
    searchParams: ['query' => ''],
    requestOptions: [
        'queryParameters' => [
            'filters' => 'value=1',
        ],
    ],
);

echo json_encode($searchResponse);

// Return both of the records

```

### Self-service

- [X] I'd be willing to fix this bug myself.

millotp:
Hello, this endpoints uses the read transporter to allow browser and the transporter to cache the request, can you try passing the filter like this ?
```php
$response = $client->searchSingleIndex(
    '<YOUR_INDEX_NAME>',
    ['query' => 'myQuery',
      'filters' => 'value=1',
    ],
 );
```

I will add this to the snippets if it works properly

aykutersoy:
@millotp It worked perfectly, thank your very much for. I've also noticed the `hitsPerPage` doesn't do anything when it's in `requestOptions` but works when it's in search params

```
[...]

PR #3699, opened 2024-09-12
Changed x-use-read-transporter for searchSingleIndex

aykutersoy:
Changed x-use-read-transporter for searchSingleIndex, the reason behind that is to be able to pass body when querying single index with filters

Closes https://github.com/algolia/api-clients-automation/issues/3698

## 🧭 What and Why

TBH I'm not 100% this the correct approach to the issue I've created (https://github.com/algolia/api-clients-automation/issues/3698) but it does the job when I set the filters in body attribute which was the case before v4

```
$searchResponse = $client->searchSingleIndex(
    indexName: $indexName,
    searchParams: ['query' => ''],
    requestOptions: [
        'body' => [
            'filters' => 'value=1',
        ],
    ],
);
```

🎟 JIRA Ticket:

### Changes included:

- List changes

### DIFF
--- a/specs/search/paths/search/searchSingleIndex.yml
+++ b/specs/search/paths/search/searchSingleIndex.yml
@@ -2,7 +2,7 @@ post:
   tags:
     - Search
   operationId: searchSingleIndex
-  x-use-read-transporter: true
+  x-use-read-transporter: false
   x-cacheable: true
   x-acl:
     - search


aykutersoy:
https://github.com/algolia/api-clients-automation/issues/3698#issuecomment-2346653529

Above comment solves the problem I had

Issue #3700, opened 2024-09-12
[bug]: window export changed after tsup bundler


shortcuts:
### Description

see https://github.com/algolia/api-clients-automation/pull/3640#issuecomment-2346825007

### Language

JavaScript

### Client

All

### Steps to reproduce

- load umd bundle
- see export name changed

### Self-service

- [X] I'd be willing to fix this bug myself.

PR #3701, opened 2024-09-12
fix(javascript): umd export name

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-2909

### Changes included:

close https://github.com/algolia/api-clients-automation/issues/3700

see context here https://github.com/algolia/api-clients-automation/issues/3700

the tsup build changed the exported global name on `window`, previously exposed `window['algoliasearch/lite']` changed to `window.lite`, which is incorrect, and a breaking change.

we now provide the previous names


Fluf22 (approved):
Toi need a PhD to handle js bundling

PR #3758, opened 2024-09-16
fix(python): pydantic v2 discriminator

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket:

### Changes included:

try pydantic v2 discriminator oneOf style

### DIFF
--- a/specs/ingestion/common/schemas/destination.yml
+++ b/specs/ingestion/common/schemas/destination.yml
@@ -10,7 +10,7 @@ Destination:
     name:
       $ref: './common.yml#/name'
     input:
-      $ref: '#/DestinationIndexName'
+      $ref: '#/DestinationInput'
     createdAt:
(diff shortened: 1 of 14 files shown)


specs/ingestion/common/schemas/destination.yml
```diff
@@ -10,7 +10,7 @@ Destination:
     name:
       $ref: './common.yml#/name'
     input:
-      $ref: '#/DestinationInput'
+      $ref: '#/DestinationIndexName'
```
millotp, line 13: is it possible to avoid the bc ?

shortcuts: I guess yes, it's just a name so I can change it

shortcuts: ah it still thinks it's a bc :/

PR #3761, opened 2024-09-17
fix(cts): use vitest instead of jest

millotp:
## 🧭 What and Why

Node has been adding a lot of cool feature that renders other packages useless, in this PR I migrate dotenv and jest for the CTS.


playground/javascript/node/package.json
```diff
@@ -29,11 +21,10 @@
     "algoliasearch": "link:../../../clients/algoliasearch-client-javascript/packages/algoliasearch"
   },
   "devDependencies": {
-    "dotenv": "16.4.5",
     "typescript": "5.6.2"
   },
   "engines": {
-    "node": ">= 14.0.0",
```
shortcuts, line 36: This should stay I think, since we allow node 14 on the client it makes it easier to try older versions

templates/javascript/tests/package.mustache
```diff
@@ -13,9 +13,7 @@
   },
   "devDependencies": {
     "@types/jest": "29.5.13",
-    "@types/node": "22.5.1",
-    "jest": "29.7.0",
```
shortcuts, line 17: Any reason not to use vitest instead like the client and scripts? This makes the migration a bit harder

shortcuts (commented):
Noice

playground/javascript/node/package.json
millotp: yeah but the env-file option is only available in node >= 20, it will not work properly

templates/javascript/tests/package.mustache
millotp: I figured the less librairies the better, but I agree that vitest is nicer to use that node runner

millotp: converted to vitest

shortcuts: i love you

config/base.tsconfig.json
```diff
@@ -6,6 +6,7 @@
     "allowSyntheticDefaultImports": true,
     "esModuleInterop": true,
     "strict": true,
+    "skipLibCheck": true,
```
shortcuts, line 9: I think we spread that config to the js client no? if it's the case, can we set the previous values there?

playground/javascript/node/package.json
shortcuts: I guess the multi version testing should take care of that, so not a huge deal

scripts/package.json
```diff
@@ -17,7 +17,7 @@
     "setRunVariables": "yarn runScript dist/ci/githubActions/setRunVariables.js",
     "spreadGeneration": "yarn runScript dist/ci/codegen/spreadGeneration.js",
     "start": "yarn runScript dist/cli/index.js",
-    "test": "vitest",
+    "test": "tsc --noEmit && vitest",
```
shortcuts, line 20: I think it's not required with vitest anymore but not sure

shortcuts, line 20: also, vitest watches by default, don't we need --run here?

shortcuts: maybe the default changes on CI actually

shortcuts (commented):
soooo nice

config/base.tsconfig.json
millotp: no the js client is fully independent, it cannot ref file outside of its folder because it's not accessible on the child repo

scripts/package.json
millotp: indeed its not required, I added it because nor the linter nor vitest check the types, I wanted at least 1 thing to validate our types.

millotp: apparently it's smart enough to not watch on the CI

shortcuts: ok that's great!

shortcuts: does it still work if we run the cts locally though? or will it hang?

shortcuts (approved):
so nice!!! it will also help me when adding the browser cts test, thanks!

PR #3764, opened 2024-09-17
chore(scripts): try experimental-strip-types [skip-bc]

millotp:
## 🧭 What and Why

Node 22 has `--experimental-strip-types` that uses `swc` under the hood to directly execute typescript, rendering all the bundler useless.


shortcuts:
Could be great to wait for active development before using this, but at the same time our scripts are pretty basic so I guess it's not a huge deal

It makes everything so much easier!

PR #3765, opened 2024-09-17
fix(csharp): remove global enum converter

morganleroi:
## 🧭 What and Why

This PR removes the global Enum to string converter from the Serializer configuration and move this to every Algolia POCO enum. It means that only Algolia POCO enum will be converted as string. Every other enum will be converter as int (default behaviour from System.Text.Json or NewtonSoft).

With this PR, the generated client behaves the same way the legacy one.

Please note that we do not use `JsonStringEnumConverter` from `System.Text.Json` because our enum values are flagged with custom name.

### Changes included:

- List changes
- Remove global enum converter from the Serializer.


millotp:
what is the usage of those enum ? currently in the snippets its like `Enum.Parse<Action>("AddObject"),`, could it be prettier ?

millotp (approved):

PR #3767, opened 2024-09-17
feat(javascript): add react-native field in package.json

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-2939

closes https://github.com/algolia/algoliasearch-client-javascript/issues/1554

### Changes included:

seems like a regression when adding the tsup bundler, I assumed the browser was covered with the `exports` field but [metro only provides an experimental support for it](https://metrobundler.dev/docs/package-exports/#configuration-options), we can [leverage `browser` in the meantime](https://metrobundler.dev/docs/resolution#resolve_module)


millotp (approved):
Will that break something with cloudflare or react native ?

shortcuts:
> Will that break something with cloudflare or react native ?

For cloudflare it should ignore it, for react native it should fix it

templates/javascript/clients/package.mustache
```diff
@@ -45,6 +45,7 @@
   },
   "jsdelivr": "./dist/builds/browser.umd.js",
   "unpkg": "./dist/builds/browser.umd.js",
+  "browser": "./dist/builds/browser.umd.js",
```
Haroenv, line 48: I don't think this is right, it will break cjs and esm most likely, I think a `react-native` key that goes to the cjs browser build is the way to go to fix that issue

shortcuts: yup, let's hope https://github.com/facebook/metro/issues/670#issuecomment-1099926849 is still a valid example, worst case the impact is smaller with this specific field

templates/javascript/clients/package.mustache
```diff
@@ -45,6 +45,7 @@
   },
   "jsdelivr": "./dist/builds/browser.umd.js",
   "unpkg": "./dist/builds/browser.umd.js",
+  "react-native": "./dist/builds/browser.umd.js",
```
Haroenv, line 48: do we not want the cjs build?

shortcuts: I think it was previously resolving to `browser`, which points to umd 🤔 does it support umd?

shortcuts: let's try with esm, if it doesn't work I'll introduce an xhr-cjs build D:

Haroenv (approved):

Fluf22 (approved):
🤞

PR #3771, opened 2024-09-18
 fix(javascript): react-native lite resolution

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-2948

### Changes included:

fixes https://github.com/algolia/algoliasearch-client-javascript/issues/1554#issuecomment-2358671330

the lite package also needs to have an entry point in the `react-native` field.

this object syntax is supported for every fields normally, so I hope it also works for react native

also adds the `react-native` field for the browser-xhr and fetch requester


millotp (approved):
the never ending fix

PR #3777, opened 2024-09-19
fix(python): exclude unset API values on model dump

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-2951

### Changes included:

see https://algolia.slack.com/archives/C336MP4UA/p1726487274846109 for context

Pydantic requires default values on optional fields (which is an other problem I should tackle), but those default values are also added to the dict when dumping the model, which we don't want because we only want to rely on what the API returns, in case there's a change there before it is reflected on the client.

We now tell Pydantic to only rely on values that are set, and ignore the defaults.

You can try it by running the playground as-is, and see that the default searchParams are not provided, only the one explicit defined


playground/python/requirements.txt
```diff
@@ -2,5 +2,5 @@ argcomplete==3.5.0
 click==8.1.7
 packaging==24.1
 pipx==1.7.1
-platformdirs==4.3.3
+platformdirs==4.2.2
```
millotp, line 5: why the downgrade ? I think this file should be excluded from PR

millotp (commented):
nice

playground/python/requirements.txt
shortcuts: idk I've let the tooling do its thing, it was after [this change](https://github.com/algolia/api-clients-automation/pull/3777/files#diff-db88605483b5202ed2a12783c3e8bf104451406dab6f47bf4d47bbef6c2b87f0L12)

millotp (approved):
gg

PR #3781, opened 2024-09-19
fix(clients): use the dsn with the read transporter

millotp:
## 🧭 What and Why

The java search client is not using the read transporter at all.


Fluf22 (approved):
Should work indeed

PR #3782, opened 2024-09-19
feat(specs): add recommend batch rules endpoint

raed667:
## 🧭 What and Why

Add the endpoint `/1/indexes/{indexName}/{model}/recommend/rules/batch` to the spec. 

Definition can be found here: https://github.com/algolia/doc/blob/master/app_data/api_methods/recommend/recommend_batch_rules.yml

🎟 JIRA Ticket:

### Changes included:

- List changes


raed667:
trying to understand what is the issue with this `RecommendRule-2` ?

specs/recommend/paths/batchRecommendRules.yml
```diff
@@ -0,0 +1,40 @@
...
+          title: rules
+          type: array
+          description: Recommend rules.
+          additionalProperties: false
+          items:
+            $ref: '../common/schemas/RecommendRule.yml#/RecommendRule'
+
+  responses:
+    '200':
+      $ref: '../../common/responses/UpdatedAt.yml'
```
millotp, line 38: Hey, because of the swift client using Cocoapod, we can't have the same model name in multiple clients, can you create a copy of `UpdatedAtResponse` called `RecommendUpdatedAtResponse` for example, so that it doesn't impact the search client ?

raed667: i think i broke something ^^' ` ISO C99 requires whitespace after the macro name`

shortcuts:
that's it, you master the repo now!

specs/recommend/paths/batchRecommendRules.yml
```diff
@@ -0,0 +1,40 @@
+post:
+  tags:
+    - rules
+  operationId: batchRecommendRules
+  x-use-read-transporter: false
+  x-cacheable: false
```
shortcuts, line 6:
```suggestion
```

shortcuts (commented):
looks great!

shortcuts (approved):
waiting for the gen before merging

PR #3786, opened 2024-09-20
fix(go): true speed of benchmark

millotp:
## 🧭 What and Why

Don't run benchmark in `race` mode because it kills the perfs


shortcuts:
ooooh that will be FAST

shortcuts:
(btw this is a chore we shouldn't put it in the changelog)

shortcuts (dismissed):
nothing to add! waiting for the gen, it looks great

millotp:
> (btw this is a chore we shouldn't put it in the changelog)

The name of the commit doesn't determine the changelog anymore, only if there are changes in the clients folder.
I guess this should be a fix(benchmark) maybe

shortcuts (approved):
> > (btw this is a chore we shouldn't put it in the changelog)
> 
> The name of the commit doesn't determine the changelog anymore, only if there are changes in the clients folder. I guess this should be a fix(benchmark) maybe

true but still a chore IMO, non blocking anyway, great fix!

PR #3787, opened 2024-09-20
fix(clients): use the correct host for search

millotp:
## 🧭 What and Why

The host for the retry stategy should end in .com


Fluf22 (approved):
✅

PR #3806, opened 2024-09-22
chore(cts): replace localhost at runtime

millotp:
## 🧭 What and Why

Because `network=host` doesn't work properly on mac, we need to use `host.docker.internal` locally and `localhost` on the CI.
Instead of generating the tests with the correct host, we can guess it at runtime based on the `CI` var env, this means we don't need to generate the tests anymore to run locally, the same code can run anywhere now !

Kotlin and ruby are a bit of a pain because the error message also contains the address, but we can replace it too.
I add to use a weird `%localhost%` otherwise mustache interprets it as a variable, or kotlin interprets it as a regex reference.


Fluf22 (approved):
That's so kind of you to have implemented this!
It was so much painful to have to edit it manually

millotp:
> It was so much painful to have to edit it manually

it would have been better to have a proxy or something but I didn't manage to make it work, this will do for now

shortcuts (approved):
never had an issue with the previous implem but this one looks good too, nice :)

PR #3822, opened 2024-09-23
fix(kotlin): deserialization

Fluf22:
## 🧭 What and Why

🎟 JIRA Ticket: [CR-6916]

### Changes included:

- update recommend spec to add oneOf discriminators
- update template to generate all possible oneOf cases
- update template to fix array deserialization

## 🧪 Test

[CR-6916]: https://algolia.atlassian.net/browse/CR-6916?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/templates/kotlin/oneof_interface.mustache
+++ b/templates/kotlin/oneof_interface.mustache
@@ -26,13 +26,9 @@ import kotlin.jvm.JvmInline
 @Serializable({{classname}}Serializer::class)
 public sealed interface {{classname}} {
     {{#composedSchemas.oneOf}}
-    {{^isModel}}
-    {{^isEnumRef}}
     @Serializable
     @JvmInline
     public value class {{#lambda.type-to-name}}{{{datatypeWithEnum}}}{{/lambda.type-to-name}}Value(public val value: {{{datatypeWithEnum}}}): {{classname}}
-    {{/isEnumRef}}
-    {{/isModel}}
     {{/composedSchemas.oneOf}}
 
     public companion object {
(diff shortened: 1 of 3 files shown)


templates/kotlin/oneof_interface.mustache
```diff
@@ -78,14 +70,33 @@ internal class {{classname}}Serializer : JsonContentPolymorphicSerializer<{{clas
...
+                              element is JsonObject{{#vendorExtensions.x-discriminator-fields}} && element.containsKey("{{{.}}}"){{/vendorExtensions.x-discriminator-fields}} -> {{{datatypeWithEnum}}}.serializer()
+                            {{/isPrimitiveType}}
+                          {{/isBoolean}}
+                        {{/isDouble}}
+                      {{/isLong}}
+                    {{/isInteger}}
+                  {{/isEnumRef}}
+                {{/isArray}}
+              {{/isMap}}
+            {{/isModel}}
```
shortcuts, line 91: image

templates/kotlin/oneof_interface.mustache
```diff
@@ -78,14 +70,33 @@ internal class {{classname}}Serializer : JsonContentPolymorphicSerializer<{{clas
...
             {{#isModel}}
             element is JsonObject{{#vendorExtensions.x-discriminator-fields}} && element.containsKey("{{{.}}}"){{/vendorExtensions.x-discriminator-fields}} -> {{{datatypeWithEnum}}}.serializer()
             {{/isModel}}
             {{#isMap}}
             element is JsonObject -> {{classname}}.{{#lambda.type-to-name}}{{{datatypeWithEnum}}}{{/lambda.type-to-name}}Value.serializer()
             {{/isMap}}
             {{#isArray}}
-            element{{#items}}{{#isModel}}.isJsonArrayOfObjects{{/isModel}}{{#isPrimitiveType}}.isJsonArrayOfPrimitives{{/isPrimitiveType}}{{^isModel}}{{^isPrimitiveType}} is JsonArray{{/isPrimitiveType}}{{/isModel}}{{/items}} -> {{{classname}}}.{{#lambda.type-to-name}}{{{datatypeWithEnum}}}{{/lambda.type-to-name}}Value.serializer()
```
shortcuts, line 88: we don't have types that mix those? this seems complicated so I wonder if there was not an edge case

Fluf22: @shortcuts in the end, it uses the same serializer for both cases, which are covered by `is JsonArray` since it's in both conditions, so it seems to me it is the same

templates/kotlin/oneof_interface.mustache
Fluf22: Felt like I was in a Context tree in React for a while

shortcuts (approved):
ah ok it was broken broken indeed 🤔 we definitely lack e2e then

shortcuts:
awesome btw, gg!!

PR #3823, opened 2024-09-23
feat(javascript): add logger-console package from v4

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-2957

### Changes included:

import the existing logger-console package, with minor tweaks to make it match our new v5 client.

the usage is similar and assert in the manual algoliasearch tests


millotp (approved):
soooo nice, I can't find it but we had a very old jira ticket for this

PR #3824, opened 2024-09-23
feat(clients): cleanup after replaceAllObjects failure [skip-bc]

millotp:
## 🧭 What and Why

[DI-2928](https://algolia.atlassian.net/browse/DI-2928)

closes https://github.com/algolia/algoliasearch-client-php/issues/739

 Whatever happens, at the end of the `replaceAllObjects` helper, the temporary index should not exists anymore.

- [x] java
- [x] javascript
- [x] c#
- [x] swift
- [x] ruby
- [x] php
- [x] python
- [x] kotlin
- [x] scala
- [x] go
- [ ] dart

[DI-2928]: https://algolia.atlassian.net/browse/DI-2928?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


templates/go/search_helpers.mustache
```diff
@@ -657,35 +657,47 @@ func (c *APIClient) ReplaceAllObjects(indexName string, objects []map[string]any
 		return nil, err
 	}
 
-  opts = append(opts, WithWaitForTasks(true))
+	opts = append(opts, WithWaitForTasks(true))
 
 	batchResp, err := c.ChunkedBatch(tmpIndexName, objects, ACTION_ADD_OBJECT, opts...)
 	if err != nil {
+		_, _ = c.DeleteIndex(c.NewApiDeleteIndexRequest(tmpIndexName))
```
shortcuts, line 664: if we have a defer and it succeed, it doesn't noop? that's why you've added it everywhere?

shortcuts (approved):
nailed it

templates/go/search_helpers.mustache
millotp: indeed, we don't want to make an extra call in case the final move succeeds

millotp:
thanks a lot @Fluf22 for the help !

PR #3835, opened 2024-09-25
fix(specs): do not expose internal URL in the public spec

kai687:
## 🧭 What and Why

Don't expose the Crawler API's dev URL in the public spec.

cc @clemfromspace


millotp (approved):
nice catch

PR #3839, opened 2024-09-26
chore(python): proper type hinting

Fluf22:
## 🧭 What and Why

🎟 JIRA Ticket: [DI-3000] & [DI-3005]

### Changes included:

- Fix type hinting in python
- Add pyright in the toolbox and trigger it at format run time

[DI-3000]: https://algolia.atlassian.net/browse/DI-3000?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DI-3005]: https://algolia.atlassian.net/browse/DI-3005?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/clients/algoliasearch-client-python/algoliasearch/http/request_options.py
+++ b/clients/algoliasearch-client-python/algoliasearch/http/request_options.py
@@ -1,6 +1,6 @@
 from copy import deepcopy
 from sys import version_info
-from typing import Any, Dict, List, Optional, Tuple, Union
+from typing import Any, Dict, Optional, Union
 from urllib.parse import quote
 
 from algoliasearch.http.base_config import BaseConfig
@@ -13,20 +13,20 @@ else:
 
 
 class RequestOptions:
-    _config: BaseConfig
-    headers: Dict[str, str]
-    query_parameters: Dict[str, Any]
-    timeouts: Dict[str, int]
-    data: Dict[str, Any]
-
     def __init__(
         self,
         config: BaseConfig,
-        headers: Dict[str, str] = {},
-        query_parameters: Dict[str, Any] = {},
-        timeouts: Dict[str, int] = {},
-        data: Dict[str, Any] = {},
+        headers: Optional[Dict[str, str]] = None,
+        query_parameters: Optional[Dict[str, Any]] = None,
+        timeouts: Optional[Dict[str, int]] = None,
+        data: Optional[Dict[str, Any]] = None,
     ) -> None:
+        if headers is None:
+            headers = {}
+        if query_parameters is None:
+            query_parameters = {}
+        if timeouts is None:
+            timeouts = {}
         self._config = config
         self.headers = headers
         self.query_parameters = {
@@ -51,13 +51,12 @@ class RequestOptions:
             query_parameters=data.get("query_parameters", {}),
             timeouts=data.get("timeouts", {}),
             data=data.get("data", {}),
-        )
+        )  # pyright: ignore
 
     def merge(
         self,
-        query_parameters: List[Tuple[str, str]] = [],
-        headers: Dict[str, Optional[str]] = {},
-        _: Dict[str, int] = {},
+        query_parameters: Optional[Dict[str, Any]] = None,
+        headers: Optional[Dict[str, str]] = None,
         data: Optional[str] = None,
         user_request_options: Optional[Union[Self, Dict[str, Any]]] = None,
     ) -> Self:
@@ -65,11 +64,15 @@ class RequestOptions:
         Merges the default config values with the user given request options if it exists.
         """
 
-        headers.update(self._config.headers)
+        if query_parameters is None:
+            query_parameters = {}
+        if headers is None:
+            headers = {}
+        headers.update(self._config.headers or {})
 
         request_options = {
             "headers": headers,
-            "query_parameters": {k: v for k, v in query_parameters},
+            "query_parameters": query_parameters,
             "timeouts": {
                 "read": self._config.read_timeout,
                 "write": self._config.write_timeout,
--- a/clients/algoliasearch-client-python/algoliasearch/http/serializer.py
+++ b/clients/algoliasearch-client-python/algoliasearch/http/serializer.py
@@ -10,7 +10,16 @@ class QueryParametersSerializer:
     Parses the given 'query_parameters' values of each keys into their string value.
     """
 
-    query_parameters: Dict[str, Any] = {}
+    def __init__(self, query_parameters: Optional[Dict[str, Any]]) -> None:
+        self.query_parameters: Dict[str, Any] = {}
+        if query_parameters is None:
+            return
+        for key, value in query_parameters.items():
+            if isinstance(value, dict):
+                for dkey, dvalue in value.items():
+                    self.query_parameters[dkey] = self.parse(dvalue)
+            else:
+                self.query_parameters[key] = self.parse(value)
 
     def parse(self, value) -> Any:
         if isinstance(value, list):
@@ -27,19 +36,8 @@ class QueryParametersSerializer:
             dict(sorted(self.query_parameters.items(), key=lambda val: val[0]))
         ).replace("+", "%20")
 
-    def __init__(self, query_parameters: Optional[Dict[str, Any]]) -> None:
-        self.query_parameters = {}
-        if query_parameters is None:
-            return
-        for key, value in query_parameters.items():
-            if isinstance(value, dict):
-                for dkey, dvalue in value.items():
-                    self.query_parameters[dkey] = self.parse(dvalue)
-            else:
-                self.query_parameters[key] = self.parse(value)
 
-
-def bodySerializer(obj: Any) -> Any:
+def body_serializer(obj: Any) -> Any:
     """Builds a JSON POST object.
 
     If obj is None, return None.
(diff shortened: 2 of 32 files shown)


clients/algoliasearch-client-python/algoliasearch/http/request_options.py
```diff
@@ -13,28 +13,30 @@
...
+        data: Dict[str, Any] = None,
     ) -> None:
+        if headers is None:
+            headers = {}
+        if query_parameters is None:
+            query_parameters = []
+        if timeouts is None:
+            timeouts = {}
+        if data is None:
+            data = {}
```
shortcuts, line 31: you can init them here instead of None, no? Also, do wee need optional?

clients/algoliasearch-client-python/algoliasearch/http/request_options.py
```diff
@@ -48,23 +50,34 @@ def from_dict(self, data: Dict[str, Dict[str, Any]]) -> Self:
...
         )
 
     def merge(
         self,
-        query_parameters: List[Tuple[str, str]] = [],
-        headers: Dict[str, Optional[str]] = {},
-        _: Dict[str, int] = {},
+        query_parameters: List[Tuple[str, Any]] = None,
+        headers: Dict[str, str] = None,
+        timeouts: Dict[str, int] = None,
```
shortcuts, line 67: TBC but you can remove timeouts

clients/algoliasearch-client-python/algoliasearch/http/serializer.py
```diff
@@ -27,19 +36,8 @@ def encoded(self) -> str:
...
-        for key, value in query_parameters.items():
-            if isinstance(value, dict):
-                for dkey, dvalue in value.items():
-                    self.query_parameters[dkey] = self.parse(dvalue)
-            else:
-                self.query_parameters[key] = self.parse(value)
-
 
-def bodySerializer(obj: Any) -> Any:
+def body_serializer(obj: Any) -> Any:
```
shortcuts, line 40: tyyy

templates/python/config.mustache
```diff
@@ -40,7 +40,7 @@ class {{#lambda.pascalcase}}{{client}}{{/lambda.pascalcase}}Config(BaseConfig):
         if {{^fallbackToAliasHost}}not region or {{/fallbackToAliasHost}}(region is not None and region not in _regions):
             raise ValueError(f"`region` {{^fallbackToAliasHost}}is required and {{/fallbackToAliasHost}}must be one of the following: {', '.join(_regions)}")
 
-        self.hosts = HostsCollection([Host({{#fallbackToAliasHost}}"{{{hostWithFallback}}}" if region is None else {{/fallbackToAliasHost}} "{{{regionalHost}}}".replace("{region}", region))])
+        self.hosts = HostsCollection([Host({{#fallbackToAliasHost}}"{{{hostWithFallback}}}" if region is None else {{/fallbackToAliasHost}} "{{{regionalHost}}}".replace("{region}", region or ""))])
```
shortcuts, line 43: do we want that default string?

shortcuts (commented):
noice

clients/algoliasearch-client-python/algoliasearch/http/request_options.py
Fluf22: If we init them in the method params, it creates mutable types, which python doesn't like, because it can creates side effect. Simple example:
```py
def func(key, value, query_params={}):
    query_params[key] = value
    return query_params
```
image

clients/algoliasearch-client-python/algoliasearch/http/request_options.py
Fluf22: Shouldn't we offer to overwrite the timeouts here?

templates/python/config.mustache
Fluf22: It's a little cumbersome, but it's to remove the warning :/
If I don't do that, python complains because, as we are in a method call (`.replace()`), the `region is not None` is not asserted anymore

Fluf22: I can remove it, but it will leave the related type hint errors.
Wdyt?

clients/algoliasearch-client-python/algoliasearch/http/request_options.py
Fluf22: A maybe more clear example about the danger in our specific case:
image

shortcuts: yup it's expected if you store it in self, or update the parameter, but here we don't do it IIRC we just read the parameter, put it in a dict, and return it, no?

clients/algoliasearch-client-python/algoliasearch/http/request_options.py
shortcuts: I think this is handled by `user_request_options`, this _timeout should be the one set at the init time of the client, but it's already stored in self

templates/python/config.mustache
shortcuts: that's a miss assertion above then, because if it doesn't have a `fallbackToAliasHost`, we should just have thrown, no?

clients/algoliasearch-client-python/algoliasearch/http/request_options.py
Fluf22: We do almost the same thing:
```py
self.query_parameters = {
            quote(k): quote(v)
            for k, v in QueryParametersSerializer(
                query_parameters
            ).query_parameters.items()
        }
```

templates/python/config.mustache
Fluf22: No, I'm not clear: there is an assertion. It's just that, here, we are in the `.replace()` so python is not sure we really asserted the value is not None
```py
if value is not None:
    "hello{world}".replace("{world}", value)
```

clients/algoliasearch-client-python/algoliasearch/http/request_options.py
shortcuts: ah I thought we were in the merge method for some reason

we create a new RequestOptions instance every time no?

shortcuts: I mean still better to avoid updating the reference anyway

Fluf22: Yes, anyway, PyCharm is really strict about this to avoid people making the mistake anywhere it's possible 😛

scripts/formatter.ts
```diff
@@ -51,7 +51,7 @@ export async function formatter(language: string, cwd: string): Promise<void> {
       break;
     case 'python':
       await run(
-        'poetry lock --no-update && poetry install --sync && pip freeze > requirements.txt && poetry run ruff check --fix --unsafe-fixes && poetry run ruff format',
+        `poetry lock --no-update && poetry install --sync && pip freeze > requirements.txt && poetry run ruff check --fix --unsafe-fixes && poetry run ruff format${cwd.includes('clients') ? ' && poetry run pyright' : ''}`,
```
millotp, line 54: can you leave it for the playground, snippets and guides pls ?

templates/python/snippets/method.mustache
```diff
@@ -19,20 +19,10 @@ def snippet_for_{{#lambda.snakecase}}{{method}}{{/lambda.snakecase}}{{testIndex}
     # Call the API 
     {{#hasResponse}}response = {{/hasResponse}}client.{{#lambda.snakecase}}{{method}}{{/lambda.snakecase}}({{#parametersWithDataType}}{{> tests/generateParams}}{{/parametersWithDataType}}{{#hasRequestOptions}} request_options={ {{#requestOptions.headers.parameters}}"headers":loads("""{{{.}}}"""),{{/requestOptions.headers.parameters}}{{#requestOptions.queryParameters.parameters}}"query_parameters":loads("""{{{.}}}"""),{{/requestOptions.queryParameters.parameters}} }{{/hasRequestOptions}})
 
-    {{#hasResponse}}print(response){{/hasResponse}}
-
-    # Skip deserialization
```
millotp, line 24: I think this was requested by customers, we need both async and sync in the snippets

tests/output/python/requirements.txt
millotp, line 1: this should not be pushed, I see it in all PR ahah

yarn.lock
millotp, line 1: same, but it should be fixed now

millotp (commented):
the more I look at this the more I hate python.

GG for doing this !

tests/output/python/requirements.txt
Fluf22: Shouldn't it be in the generation.config.mjs then?
I usually try to avoid pushing it manually, but it's not filtered by the pre-commit hook

millotp: yes maybe it should be, I thought it was already in it

templates/python/snippets/method.mustache
Fluf22: There is still sync/async snippets, it's above your comment
I removed the call to the _with_http_info variant of the method

millotp (approved):
all good !

PR #3840, opened 2024-09-27
fix(specs): API key description will be obfuscated

kai687:
## 🧭 What and Why

Update the description for the 'Get API key' operation.
API key description will be obfuscated if using a non-admin API key.
This is a port from a change made in the docs repo. More information in https://github.com/algolia/doc/pull/9412


shortcuts (approved):
good catch! gg

PR #3841, opened 2024-09-27
feat(specs): add cURL snippets

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-3026

### Changes included:

some customer requester cURL example in the API reference page https://www.algolia.com/doc/rest-api/search/, I assume it is to quickly try some stuff before setting up a real codebase

the openapi-generator is a bit overkill and doesn't properly handle oneOf cases, this simple solution seems to be enough


shortcuts:
use https://www.npmjs.com/package/@har-sdk/oas with https://github.com/Kong/httpsnippet instead

scripts/specs/format.ts
```diff
@@ -1,9 +1,11 @@
 import fsp from 'fs/promises';
 
+import oas2har from '@har-sdk/oas';
```
millotp, line 3:
```suggestion
import { oas2har } from '@har-sdk/oas';
```
since you only use 1 function

scripts/specs/format.ts
```diff
@@ -100,6 +103,33 @@ export async function transformBundle({
...
+
+          if (harRequestHeader.name === bundledSpec.components.securitySchemes.apiKey?.name) {
+            harRequestHeader.value = 'ALGOLIA_API_KEY';
+          }
+        }
+
+        specMethod['x-codeSamples'].push({
+          lang: 'cURL',
+          label: 'curl',
+          source: new HTTPSnippet(harRequest as HarRequest).convert('shell', 'curl', { indent: '\t' }),
```
millotp, line 129: why isn't it outputed like `source: >` like the other snippets ? maybe you need to convert it to a string with `.toString()` and replace the new lines

millotp (commented):
looks great !

scripts/specs/format.ts
shortcuts: https://github.com/algolia/api-clients-automation/pull/3841/commits/a7aad853b26e69ed87bcf7a26ee240cc8107bfce seems enough

millotp: much more readable !

millotp (approved):
looks f great !

PR #3868, opened 2024-10-01
feat(templates): add request_options to python helpers consistently

kai687:
## 🧭 What and Why

Make the Python helpers more consistent by supporting `request_options` for all helpers
that are based on `chunked_batch`.

### Changes included:

Add `request_options` to the following helpers:

- `save_objects`
- `delete_objects`
- `partial_update_objects`

The rest of them already work with `request_options`.

Also provide `dict` as accepted type for API key restrictions in `generate_secured_api_key`, since the function already handles that case.

### DIFF
--- a/templates/python/search_helpers.mustache
+++ b/templates/python/search_helpers.mustache
@@ -244,32 +244,35 @@
         self,
         index_name: str,
         objects: List[Dict[str, Any]],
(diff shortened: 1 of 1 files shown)


Fluf22 (approved):
Thanks for that!
Btw, I'm working on Python typing in general in another pr
I found other issues

PR #3872, opened 2024-10-03
chore(scripts): ignore guides supporting files

millotp:
## 🧭 What and Why

Ignore all generated files in the guides, and sometimes the `yarn.lock` file


Fluf22 (approved):
🆗

PR #3873, opened 2024-10-03
chore(ci): allow fork to access secrets

millotp:
## 🧭 What and Why

To allow external contributors, we need to allow them to access token on the CI, this is safe because we need to approve the CI before it can run.


Fluf22 (approved):
🆗

PR #3874, opened 2024-10-03
fix(python): pass query parameters to request

kai687:
## 🧭 What and Why

Apply the query parameters in `_session.request`.

The Python transporter didn't apply query parameters.
Methods with options passed as query parameters, such as `partial_update_object` with `create_if_not_exists` didn't work correctly.

Fixes: https://github.com/algolia/algoliasearch-client-python/issues/572

### DIFF
--- a/tests/CTS/requests/search/partialUpdateObject.json
+++ b/tests/CTS/requests/search/partialUpdateObject.json
@@ -104,7 +104,7 @@
       "attributesToUpdate": {
         "attributeId": "value"
       },
-      "createIfNotExists": false
+      "createIfNotExists": "false"
     },
     "request": {
       "path": "/1/indexes/theIndexName/uniqueID/partial",


millotp (commented):
I think the python client is working properly, query params are already part of the path, and they are working properly, we have e2e tests for that, maybe the client is running an older version ?

I tried it myself in the playground and it works as expected, with False or True

millotp:
[This test](https://github.com/algolia/api-clients-automation/blob/main/tests/output/python/tests/e2e/ingestion_test.py#L129) asserts that query params are correctly passed thanks to [this](https://github.com/algolia/api-clients-automation/blob/main/clients/algoliasearch-client-python/algoliasearch/http/transporter.py#L52), we encode the params ourself instead of relying on the `aiohttp` because we do custom stuff for arrays and whitespaces

millotp:
maybe there is something broken for the search client.
Did you try with the sync or async client ?

PR #3875, opened 2024-10-04
feat(swift): guide gen

Fluf22:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

- one gen per guide template
- in its own folder
- with the guide name as a folder name
- the file name is always main.swift
- with a gen Package.swift with all the guides
- each has its own taget
- so it can build on macos and linux and be copy/pasted by users


millotp:
for now we stick to a more common approach for swift guides, if we to run them in this repo we can come back to this PR

PR #3897, opened 2024-10-07
fix(javascript): node 14 compatibility

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-3024

### Changes included:

closes https://github.com/algolia/algoliasearch-client-javascript/issues/1559

`replaceAll` is only available in node 15, we support node 14 so this should do it


millotp (commented):
good catch

millotp (approved):

PR #3927, opened 2024-10-08
fix(ruby): prefix models by the module

millotp:
## 🧭 What and Why

The cts, guides, and snippets, should use this syntax to work out of the box, without importing anything.


Fluf22 (approved):
Nice!

PR #3933, opened 2024-10-09
chore(scripts): allow variadic compilation for javascript

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket:

### Changes included:

this pr updates the build script to allow variadic client compilation on the javascript client, since this is the only fully variadic client.

not sure if it breaks other things, will see with the CI

example: https://github.com/algolia/api-clients-automation/actions/runs/11253862443/job/31290141181?pr=3932 -- the ci runs only for a subset of clients, but we ask for every playground/snippets/guides to be compiled, since it requires the client to be compiled too, this would fail

### DIFF
--- a/scripts/buildLanguages.ts
+++ b/scripts/buildLanguages.ts
@@ -48,11 +48,13 @@ async function buildLanguage(language: Language, gens: Generator[], buildType: B
           packageName === 'algoliasearch' ? packageName : `@algolia/${packageName}`,
         );
         await run(`yarn build:many '{${packageNames.join(',')},}'`, { cwd, language });
-      } else if (buildType === 'playground') {
-        await run('yarn build', { cwd: `${cwd}/node`, language });
-      } else {
-        await run('yarn tsc --noEmit', { cwd, language });
+        break;
       }
+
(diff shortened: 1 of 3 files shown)


scripts/buildLanguages.ts
```diff
@@ -89,10 +101,20 @@ async function buildLanguage(language: Language, gens: Generator[], buildType: B
...
+export async function buildLanguages(generators: Generator[], scope: BuildType): Promise<void> {
   const langs = [...new Set(generators.map((gen) => gen.language))];
   const generatorsMap = generators.reduce(
     (map, gen) => {
+      // TODO: remove this when guides are mandatory and implemented in every clients
+      if (scope === 'guides' && !existsSync(toAbsolutePath(`docs/guides/${gen.language}`))) {
+        return map;
+      }
+
+      // there is no monitoring client for now
```
millotp, line 113: bah yes there is

millotp (commented):
good idea !

scripts/buildLanguages.ts
shortcuts: ah yes it's not needed anymore it was just missing the dart guides case I think

millotp:
gg

millotp (approved):
gg fr

shortcuts:
> gg fr

cimr frr

PR #3935, opened 2024-10-09
chore: apply codacy fixes

millotp:
## 🧭 What and Why

We use Codacy to statically analyse the code and find vulnerabilities.


clients/algoliasearch-client-go/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/algolia/algoliasearch-client-go/v4
 
-go 1.21
+go 1.22
```
shortcuts, line 3: 23 even, no?

shortcuts (commented):
Linter as a service

clients/algoliasearch-client-go/go.mod
millotp: Can people using go1.22 still use this lib if we use 1.23 ?

shortcuts: I think so yes, as long as we don't depend on std introduced in 1.21+

I guess we should add multi version testing to Go too D:

millotp: Just tried it with the CTS: `go: module ../../../clients/algoliasearch-client-go requires go >= 1.23 (running go 1.22.8)`

millotp: we should stick to a low version

shortcuts: oki nice then let's keep 1.21 because that's what we settled on at the major release

shortcuts (commented):
looks cleaaaan, not sure about Go

shortcuts (approved):
gégé

PR #3942, opened 2024-10-10
feat(javascript): add every APIs

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-3032 https://algolia.atlassian.net/browse/DI-3051

### Changes included:

for legacy purposes, algoliasearch only provided a few clients, recommend has been added recently but for consistency with other clients, and easier usage from our customer that leverage multiple products, we can have a single big bundle with every APIs.

the standalone clients are still usable, and dependencies of the algoliasearch client, the lite browser client remains unchanged.

also fixes:
- bundle the logger-console package, since it's not a dependency, it was never built
- ci head_ref variable


millotp (approved):
generated code looks good !

PR #3943, opened 2024-10-10
fix(go): better network error

millotp:
## 🧭 What and Why

The message `You can use 'ExposeIntermediateNetworkErrors: true'` should not appear when it is true, we can show the real errors instead.


Fluf22 (approved):
👌🏻

PR #3972, opened 2024-10-15
fix(ruby): add helper to add segment to user agent

millotp:
## 🧭 What and Why

🎟 JIRA Ticket: [DI-3070](https://algolia.atlassian.net/browse/DI-3070)

Add helpers to customize ruby user agent, either with: 
`client = Algolia::SearchClient.create(appID, apiKey, {user_agent_segments: ['Rails (3.0.0)']})`

or with `client.add_user_agent_segment('Rails', '3.0.0)`

### DIFF
--- a/clients/algoliasearch-client-ruby/lib/algolia/configuration.rb
+++ b/clients/algoliasearch-client-ruby/lib/algolia/configuration.rb
@@ -33,11 +33,17 @@ module Algolia
 
       @user_agent = UserAgent.new.add(client_name, VERSION)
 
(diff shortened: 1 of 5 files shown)


shortcuts (approved):
nice, did you find other clients where it was missing?

DevinCodes (approved):

millotp:
> nice, did you find other clients where it was missing?

I only looked at ruby for now because it's needed for https://github.com/algolia/algoliasearch-rails/pull/449, I don't know if I can add a test for that

PR #3973, opened 2024-10-15
fix(clients): escape char in comments

millotp:
## 🧭 What and Why

Remove all the ugly `&#x60;` or `&quot;` that are actually breaking the scala client.


shortcuts (approved):
wonderful

PR #3991, opened 2024-10-18
feat(javascript): assert algoliasearch bundle

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-3086

### Changes included:

use algoliasearch in the cts, since it's not in the workspace, and we run the tests from the dist, it will assert the end bundle as well and catch errors like https://github.com/algolia/algoliasearch-client-javascript/issues/1562

also fix a bug that required the recommend a monitoring client to provide an empty `{}` due to missing optional condition when no region are required


millotp (approved):
codegen looks flawless ! it might be worth to keep the single client init somewhere

PR #4015, opened 2024-10-22
fix(javascript): common package types

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-3100

### Changes included:

there was a regression in `5.9.0` that removed the type definition for the handwritten packages (requester, logger and common) because the `index.ts` gets confused for some reason when bundling types, the `.d.ts` (see published type https://www.npmjs.com/package/@algolia/client-common?activeTab=code)

also the library check was skipped when bundling types, so... our test suite wouldn't catch it


millotp (approved):
gg

PR #4016, opened 2024-10-22
fix(python): required aggregator parameter in browse_*

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-3101

### Changes included:

closes https://github.com/algolia/algoliasearch-client-python/issues/573

the aggregator should be required in the browse context, there's no point of calling the helper for just the last response


shortcuts:
> can you also update the doc pls ?

done here https://github.com/algolia/doc/pull/9478

PR #4024, opened 2024-10-23
fix(specs): results in getObjects is required

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

reported by @kombucha when migrating the dashboard api client, see https://algolia.slack.com/archives/C336MP4UA/p1729586897712759?thread_ts=1728917031.913569&cid=C336MP4UA

the results field is always there, but the objects are nullable


Fluf22 (approved):
🆗

PR #4028, opened 2024-10-24
chore(swift): fix configuration protection level

Fluf22:
## 🧭 What and Why

🎟 JIRA Ticket: [DI-3103]& fixes algolia/algoliasearch-client-swift#872

### Changes included:

- make the client configuration init public


shortcuts (approved):
oops

PR #4031, opened 2024-10-25
fix(specs): support synonyms type in camel case [skip-bc]

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-3108

### Changes included:

see https://github.com/algolia/AlgoliaWeb/pull/22442#discussion_r1816448163 and https://github.com/algolia/AlgoliaWeb/pull/22442#discussion_r1816451248 for context

the API accepts both casing, and the search is case insensitive, the camel case is the more obvious one and was supported in the previous major


shortcuts:
I guess this will fail for some clients since it generates the same name

millotp (approved):
much better !

PR #4060, opened 2024-10-31
fix(javascript): allow import from dist

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-3141

### Changes included:

It's impossible to import from the dist right now because it's not defined in the exports


Fluf22 (approved):
✅

PR #4061, opened 2024-10-31
fix(specs): add secrets payload for updates

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

the input payload needs to include the secrets shape too


Fluf22 (approved):
This approval was a difficult decision

PR #4094, opened 2024-11-12
chore(eslint): add linter to forbid bigInt [skip-bc]

millotp:
## 🧭 What and Why

I saw this type in one of the external contribution, we should never allow `BitInteger` or `BigDecimal` types


shortcuts (approved):
incredible

PR #4096, opened 2024-11-13
fix(go): handle HTML errors [skip-bc]

millotp:
## 🧭 What and Why

GCP Cloud Armor can return raw HTML error, we can catch them and handle them gracefully instead of failing the json deserialization.


shortcuts (commented):
looks much cleaner! let's not forget to add a test when we do other clients

shortcuts (approved):
meant to approve gg

PR #4097, opened 2024-11-13
fix(cts): add tests for HTML error

millotp:
## 🧭 What and Why

Same as go, we need to properly handle html error, currently it's less than ideal:
```json
 "error": {
    "csharp": "<html><body>429 too many requests</body></html>",
    "go": "API error [429] Too Many Requests",
    "java": "Status Code: 429 - <html><body>429 Too Many Requests</body></html>",
    "javascript": "<html><body>429 Too Many Requests</body></html>",
    "kotlin": "Client request(GET http://%localhost%:6676/1/html-error) invalid: 429 Too Many Requests. Text: \"<html><body>429 Too Many Requests</body></html>\"",
    "php": "json_decode error: Syntax error",
    "python": "Too Many Requests",
    "ruby": "unexpected token at '<html><body>429 Too Many Requests</body></html>'",
    "scala": "<html><body>429 Too Many Requests</body></html>",
    "swift": "HTTP error: Status code: 429 Message: No message"
  }
```


clients/algoliasearch-client-java/algoliasearch/src/main/java/com/algolia/internal/interceptors/RetryStrategy.java
```diff
@@ -84,6 +84,9 @@ private Response handleResponse(StatefulHost host, @Nonnull Response response) t
 
     try {
       String message = response.body() != null ? response.body().string() : response.message();
+      if (response.header("Content-Type", "application/json").contains("text/html")) {
```
shortcuts, line 87: can it have both? or do I misunderstand the header method?

shortcuts: ahhh it's a default?

shortcuts (approved):
efficace

clients/algoliasearch-client-java/algoliasearch/src/main/java/com/algolia/internal/interceptors/RetryStrategy.java
shortcuts: might speed up things my .2ns

millotp: it's only in the rate limiting case, it will rate limit them even more

PR #4104, opened 2024-11-14
fix(scripts): dart and js releases [skip-bc]

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-3212

### Changes included:

- add a check to avoid releasing `undefined` versions of dart
- drop `lerna` for version bumping as it doesn't support mixed package stage (stable and pre-release) bumps nor filtering on package bumps (e.g. bumping stable then bumping pre-releases with different release type)
  - we now list every packages with `lerna` then use `yarn` directly to bump the version based on the `releaseType`

also move dart and js custom logic to their own file because it became a mess in that release script file

### Test:

run `yarn cli release -b -d` and see dart properly bumped, javascript as well


scripts/release/dart.ts
shortcuts, line 1: nothing changed here, just moved the code

scripts/release/updateAPIVersions.ts
```diff
@@ -58,11 +58,9 @@ export async function updateAPIVersions(versions: Versions, changelog: Changelog
...
     }
 
-    if (lang === 'javascript') {
+    if (lang === 'javascript' && releaseType) {
       setVerbose(CI);
-      await run(`yarn install && yarn release:bump ${releaseType}`, {
-        cwd: getLanguageFolder(lang),
-      });
+      await updateJavaScriptPackages(releaseType);
     }
```
shortcuts, line 64: except the code move, this was the only change

millotp (approved):
gg

PR #4106, opened 2024-11-14
fix(javascript): add support for private package publish

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

the release script for js failed on the composition package, I assume it is because we went from a private to a public package with manual interventions

this pr updates the public script to support (in a non programmatically way) the private and alpha releases


millotp (approved):
looks good, too bad there is no way to know based on the package.json

PR #4108, opened 2024-11-15
fix(specs): make the searchParams compatible with v4 [skip-bc]

millotp:
## 🧭 What and Why

The properties `optionalWords` and `insideBoundingBox` can be string on the engine, we need to reflect that on the client. This was not fixed on purpose in #772 but this is needed now to upgrade docusaurus in https://github.com/facebook/docusaurus/pull/10672

Also fix a type error that breaks the `search` method because of type incompatibility.


shortcuts (approved):
CLEAN

PR #4113, opened 2024-11-15
chore(scripts): change dist folder and bump

millotp:
## 🧭 What and Why

The js release failed because tsc changed the output dir for some reason.
I tried releasing manually and failed so I had to bump all version a patch up and release on top of the buggy version.
This is very dirty and I have learned a lesson not to release on a friday night


shortcuts (approved):
ah ok so because there's only one file now it's not worth a dir anymore?

millotp:
I didn't find the deleted file but yes that might be the reason

shortcuts:
> Sec32fun32

Thanks @Sec32fun32 for your input! Could you please elaborate?

PR #4127, opened 2024-11-18
fix(specs): do not cache searchCompositionRules

e-krebs:
## 🧭 What and Why

🎟 JIRA Ticket:

### Changes included:

- remove `x-use-read-transporter` & `x-cacheable` that where set by default


millotp (approved):
👍

PR #4131, opened 2024-11-18
chore(deps): upgrade openapi-generator to v7.10.0

millotp:
## 🧭 What and Why

With 7.10.0, there are a lot fewer bug, and we can actually fix them in the spec directly thanks to a bypass.


shortcuts (approved):
nicety

PR #4136, opened 2024-11-19
feat(specs): add info and link about indexing rate limits

kai687:
## 🧭 What and Why

The old API client method pages for indexing operations had a note about rate limits.
This PR adds a minimal version to the API specs with a link to the  to avoid duplicating information.

🎟 JIRA Ticket:

### Changes included:

- Add a sentence with a link to [support article](https://support.algolia.com/hc/en-us/articles/4406975251089-Is-there-a-rate-limit-for-indexing-on-Algolia) to indexing operations in the Search API and to the Records tag description
- Change the summary for `deleteBy` to "Delete records matching a filter" as you can't pass a search query to this method, so it's misleading.


Fluf22 (approved):
✅

PR #4138, opened 2024-11-20
fix(javascript): set latest version on prereleases

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

the published alpha version isn't set as the latest available, meaning that npm shows the .1 everytime (cf screenshot)

we now set the latest tag explicitly, which seems to be enough https://www.npmjs.com/package/@algolia/client-composition?activeTab=versions

image


millotp (approved):
we don't need to do it for the normal packages ?

PR #4141, opened 2024-11-21
fix(python): expose `user_agent` on the config

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

there's no easy way to just get the user agent or just add segments to it, rather than keeping it local to the config we could just expose it so it's easier to play with

### DIFF
--- a/clients/algoliasearch-client-python/algoliasearch/http/user_agent.py
+++ b/clients/algoliasearch-client-python/algoliasearch/http/user_agent.py
@@ -19,6 +19,10 @@ class UserAgent:
     def get(self) -> str:
         return self.value
 
-    def add(self, segment: str, version: Optional[str] = __version__) -> Self:
-        self.value += "; {} ({})".format(segment, __version__ if version is None else version)
+    def add(self, segment: str, version: Optional[str] = None) -> Self:
+        self.value += "; {}".format(segment)
+
+        if version is not None:
+            self.value += " ({})".format(version)
+
         return self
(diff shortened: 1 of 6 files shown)


playground/python/app/search.py
```diff
@@ -13,6 +13,9 @@ def main():
     client = SearchClientSync(
         environ.get("ALGOLIA_APPLICATION_ID"), environ.get("ALGOLIA_ADMIN_KEY")
     )
+    client._config.user_agent.add("playground")
+
+    print("user_agent", client._config.user_agent.get())
```
shortcuts, line 18: this adds
```
user_agent Algolia for Python (4.9.2); Python (3.11.6); Search (4.9.2); playground (4.9.2)
```

clients/algoliasearch-client-python/algoliasearch/http/user_agent.py
```diff
@@ -20,5 +20,5 @@ def get(self) -> str:
         return self.value
 
     def add(self, segment: str, version: Optional[str] = __version__) -> Self:
-        self.value += "; {} ({})".format(segment, version)
+        self.value += "; {} ({})".format(segment, __version__ if version is None else version)
```
millotp, line 23: version is already `__version__` by default no ?

shortcuts: it defaults to None above so without the ternary it will stay None

But actually we might want to keep None if not provider 🤔

millotp: as you want, but this is some weird behavior

millotp (approved):
gggg

PR #4144, opened 2024-11-22
fix(javascript): allow async param on createIterablePromise

millotp:
## 🧭 What and Why

Allow all params of `createIterablePromise` to return `PromiseLike` and call them with await, if people want to use `await` inside their fonctions.

This is not a breaking change and stays compatible with before.

### DIFF
--- a/clients/algoliasearch-client-javascript/packages/client-common/tsconfig.json
+++ b/clients/algoliasearch-client-javascript/packages/client-common/tsconfig.json
@@ -2,8 +2,9 @@
   "extends": "../../tsconfig.json",
   "compilerOptions": {
     "types": ["node", "vitest/globals"],
-    "outDir": "dist"
+    "outDir": "dist",
+    "skipLibCheck": true
   },
   "include": ["src"],
-  "exclude": ["dist", "node_modules", "src/__tests__"]
+  "exclude": ["dist", "node_modules"]
 }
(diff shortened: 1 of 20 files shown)


shortcuts (commented):
great find! can we leave one of the playground with an async and sync implem in it? just so the build secures it

shortcuts:
> great find! can we leave one of the playground with an async and sync implem in it? just so the build secures it

actually just a test in client-common would be good?

clients/algoliasearch-client-javascript/packages/client-common/tsconfig.json
```diff
@@ -2,8 +2,9 @@
   "extends": "../../tsconfig.json",
   "compilerOptions": {
     "types": ["node", "vitest/globals"],
-    "outDir": "dist"
+    "outDir": "dist",
+    "skipLibCheck": true
```
shortcuts, line 6:
```suggestion
    "outDir": "dist"
```

does it still work without it?

shortcuts (commented):
:noice:

clients/algoliasearch-client-javascript/packages/client-common/tsconfig.json
millotp: nope, still complains about rollup

shortcuts: (╯°□°)╯︵ dnʅʅoɹ

shortcuts (approved):
looks bon to moi gg!

PR #4169, opened 2024-11-26
guides: dart -- add dynamic snippet

levimichael:
Small updates -- I intended to use the dynamic snippet like we do for the other clients, bit since `saveObjects()` is not available on the Dart client, this PR just turned into a little cleanup exercise.

### DIFF
--- a/templates/dart/guides/search/saveObjectsMovies.mustache
+++ b/templates/dart/guides/search/saveObjectsMovies.mustache
@@ -20,7 +20,7 @@ void main() async {
       );
     }
 
-  try {
+    try {
       // push data to algolia
       await client.batch(
         indexName: 'movies_index',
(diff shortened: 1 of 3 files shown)


templates/dart/guides/search/saveObjectsMovies.mustache
```diff
@@ -21,7 +20,7 @@ void main() async {
       );
     }
 
-    try {
+  try {
       // push data to algolia
       await client.batch(
```
millotp, line 25: did you mean to call `{{> tests/method}}` here ?

levimichael: No, i'll actually just close this! We ended up not really doing what the PR was intended to achieve.

PR #4170, opened 2024-11-26
fix(specs): `saveRule` response type

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-3246

### Changes included:

seems like a wrong type leftover from 2 years ago but no objectID fields are returned from the response, it actually returns `{"updatedAt":"2024-11-26T16:07:56.523Z","taskID":588516396000}`


Fluf22 (approved):
✅

millotp (approved):
good catch !

millotp (approved):

PR #4171, opened 2024-11-26
chore(ci): run the macos jobs in parallel [skip-bc]

millotp:
## 🧭 What and Why

Run the macos jobs in parallel of the other codegen because they are slow as hell


shortcuts (approved):
thankssssss

PR #4010, opened 2024-10-21
fix(specs): method names in breaking changes table

kai687:
## 🧭 What and Why

Update the `major-breaking-changes-rename.json` file to fit the table introduced in https://github.com/algolia/doc/pull/9415

- I removed `index.exists` because this helper isn't documented yet. I want to add the docs for it in https://github.com/algolia/doc/pull/9420

🎟 JIRA Ticket:

### Changes included:

- List changes

PR #4188, opened 2024-12-02
chore: update major bc changes references

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

observed in https://github.com/algolia/AlgoliaWeb/pull/22965

the ruby client had a different naming so we can add it to the table as well


specs/major-breaking-changes-rename.json
```diff
@@ -56,6 +57,7 @@
     "index.deleteObjects": "client.deleteObjects",
     "index.deleteRule": "client.deleteRule",
     "index.deleteSynonym": "client.deleteSynonym",
+    "index.exists": "client.indexExists",
```
millotp, line 60: https://github.com/algolia/api-clients-automation/pull/4010

shortcuts: cc @kai687 can we just not generate the link for it? it's still great to let users that can use an LSP/IDE benefit from the inline documentation

shortcuts: cc @kai687

shortcuts, line 60:
```suggestion
```

Fluf22 (approved):
✅

PR #4191, opened 2024-12-02
fix(ruby): expose static helper [skip-bc]

millotp:
## 🧭 What and Why

Suggested by @kombucha in https://github.com/algolia/AlgoliaWeb/pull/22965#discussion_r1865709618, we can expose static helpers where needed, and forward the call of instance method to the static function to make it non-breaking.


shortcuts (approved):
gg

PR #4193, opened 2024-12-02
fix(specs): add sourceType to listTasks [skip-bc]

millotp:
## 🧭 What and Why

Add the `sourceType` query param to `listTasks`.

Also fix the type of the `type` query param in the `listRuns` endpoint.

### DIFF
--- a/specs/ingestion/common/observabilityParameters.yml
+++ b/specs/ingestion/common/observabilityParameters.yml
@@ -30,7 +30,9 @@ runType:
   in: query
   description: Run type for filtering the list of task runs.
   schema:
-    $ref: './schemas/run.yml#/RunType'
+    type: array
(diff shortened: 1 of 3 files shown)


specs/ingestion/common/observabilityParameters.yml
```diff
@@ -30,9 +30,7 @@ runType:
   in: query
   description: Run type for filtering the list of task runs.
   schema:
-    type: array
-    items:
-      $ref: './schemas/run.yml#/RunType'
+    $ref: './schemas/run.yml#/RunType'
```
shortcuts, line 33: can you leave it as an array actually? it will be the case really soon

shortcuts (approved):
great!

PR #4194, opened 2024-12-03
chore: use vite for browser playground [skip-bc]

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

use vite for browser playground, because it's easier to use and it seems to work with the workspaces, it also allows a build from the CI


millotp (dismissed):
nice

millotp (approved):

PR #4196, opened 2024-12-04
guides: push records.json location [skip-bc]

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

we shouldn't recommend to read a file at the root of the disk, we can keep it at the script level


morganleroi (approved):
Thanks !

PR #4228, opened 2024-12-11
fix(clients): lock version

millotp:
## 🧭 What and Why

The dart release is failing because the lock file is outdated compared to the ruby and bundler version.


shortcuts (approved):
ooook let's go

PR #4230, opened 2024-12-11
fix(specs): add `ignoreConjugations` to `AlternativesAsExact` [skip-bc]

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

reported by @kombucha, this property is missing on the allowed enum values


shortcuts:
> While this will help in the short term, ideally we shouldn't see the client crashing anytime the engine adds new valid values / extra properties IMO. Something like #4143, but also applied to enum values and such would be desirable 👀

yup @millotp is taking a look at it!

PR #4231, opened 2024-12-11
fix(ruby): handle unknown attributes in index_exists

millotp:
## 🧭 What and Why

This should fix the workaround in https://github.com/algolia/AlgoliaWeb/pull/23088, if we encounter a parsing error we know the index exist and can return true.


Fluf22 (approved):
✅

PR #4234, opened 2024-12-12
feat(specs): update composition specs according to upcoming API changes

e-krebs:
## 🧭 What and Why

Update composition specs according to upcoming API changes
https://algolia.atlassian.net/wiki/spaces/CMP/pages/5556273184/DD+-+Composition+API+updates

🎟 JIRA Ticket: [EMERCH-1797](https://algolia.atlassian.net/browse/EMERCH-1797)

[EMERCH-1797]: https://algolia.atlassian.net/browse/EMERCH-1797?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


specs/composition/common/schemas/SearchResponse.yml
```diff
@@ -33,3 +34,227 @@ searchHits:
...
+  type: object
+  additionalProperties: true
+  required:
+    - processingTimeMS
+  properties:
+    compositions:
+      $ref: '#/compositionsSearchResponse'
+    abTestID:
+      type: integer
+      description: A/B test ID. This is only included in the response for indices that are part of an A/B test.
```
shortcuts, line 88:
```suggestion
    abTestID:
    abTestID:
      $ref: '../../../common/schemas/SearchResponse.yml#/baseSearchResponse/properties/abTestID'
```

Can you use references here please instead of copying? this way we still have to maintain a single object

e-krebs: for the whole baseSearchReponse ?
I couldn't make it work as it broke both the javascript & swift builds unfortunately

shortcuts: for this composition searchResponse yes you can use property level references and it should work the same way 🤔

e-krebs: ha, I'll try that, I wasn't aware this is possible

shortcuts: yup exactly like in my example should work

e-krebs:
@shortcuts I previously mis-implemented the search response, it is now fixed (cf. doc linked in description for the correct shape)

tests/CTS/requests/composition/listCompositions.json
shortcuts, line 1: would be great to have more concrete e2e so we can assert those changes, can be done later though

shortcuts (approved):
thanks!

PR #4253, opened 2024-12-17
fix(ruby): Set default hits_per_page for SearchClient#browse_objects using accessor

dtyoung:
## 🧭 What and Why
SearchClient#browse_objects is attempting to set a default `hits_per_page` value on `browse_params` using `[]` syntax, but Search::BrowseParamsObject doesn't implement the `[]` method so any call to browse_objects using either the default or a custom BrowseParamsObject param throws an error.

```
irb(main):001> client = Algolia::SearchClient.create("application_id", "api_key"))
=> #<Algolia::SearchClient:0x00000001674d0f40 ...

irb(main):002> client.browse_objects("index_name")
(irb):2:in `<main>': undefined method `[]' for an instance of Algolia::Search::BrowseParamsObject (NoMethodError)
browse_params[:hits_per_page] = browse_params[:hits_per_page] || 1000
```

### Changes included:
- Use the setter defined by the hits_per_page`attr_accessor` in BrowseParamsObject to set the default value on the provided BrowseParamsObject param.


shortcuts (approved):
thanks a lot!!!!!!!!

PR #4254, opened 2024-12-17
fix(clients): support dict in helpers

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-3335

### Changes included:

we have runtime errors if we use dict/hash instead of models, but the clients support both, we should align in the helpers as well


playground/ruby/search.rb
```diff
@@ -6,7 +6,7 @@
 client = Algolia::SearchClient.create(ENV['ALGOLIA_APPLICATION_ID'], ENV['ALGOLIA_ADMIN_KEY'])
 # set a custom user agent
 client.add_user_agent_segment('Algolia for rails', "test")
-res = client.search_single_index('contacts', Algolia::Search::SearchParamsObject.new(query: 'Jimmie'))
+res = client.browse_objects('qigbuery-RECORDS')
```
morganleroi, line 9: Do we really want this name ? `qigbuery-RECORDS`

shortcuts: ahahha this is one of our e2e index plus me that does silly things

we don't really _validate_ playgrounds in this repo, as long as it compiles it's fine

morganleroi (approved):
Seems good but I don't read Ruby :-D

shortcuts:
> Seems good but I don't read Ruby :-D

me neither dw

PR #4285, opened 2024-12-24
feat(javascript): expose current apiKey on the client instance [skip-bc]

Haroenv:
## 🧭 What and Why

Expose the currently used api key. This is good for instantiating other information (like insights client) and is consistent with the appId being exposed.

### Changes included:

- add apiKey to the client template

## 🧪 Test

```js
const appId = "appId";
const apiKey = "apiKey";

const client = algoliasearch(appId, apiKey) // and other clients

expect(client. appId).toBe(appId)
expect(client.apiKey).toBe(apiKey)
```


shortcuts (approved):
nice ty!!

shortcuts (approved):
20/20

PR #4297, opened 2025-01-02
feat(specs): add notification settings to tasks [skip-bc]

millotp:
## 🧭 What and Why

🎟 JIRA Ticket: [DI-3239](https://algolia.atlassian.net/browse/DI-3239)

Add `notifications` to the task API

[DI-3239]: https://algolia.atlassian.net/browse/DI-3239?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


specs/ingestion/common/schemas/task.yml
```diff
@@ -41,7 +43,7 @@ TaskV1:
   type: object
   additionalProperties: false
   deprecated: true
-  description: The V1 task object, please use methods and types that don't contain the V1 suffix.
+  description: The V1 task object, please use methods and types that doesn't contain the V1 suffix.
```
Fluf22, line 46: `don't` is the right form, it's a plural 😁

Fluf22 (commented):
Just a typo

specs/ingestion/common/schemas/task.yml
```diff
@@ -553,3 +569,30 @@ ShopifyMarket:
...
+      type: object
+      additionalProperties: false
+      properties:
+        enabled:
+          description: Whether to send email notifications, note that this doesn't prevent the task from being blocked.
+          type: boolean
+  required:
+    - email
+
+Policies:
```
shortcuts, line 589: is `policies` a relevant name in that case? Do we already have other rules in mind than the error thresholds?

shortcuts: also could be great to move the failure threshold there

specs/ingestion/common/taskParameters.yml
```diff
@@ -70,6 +70,14 @@ triggerType:
       $ref: './schemas/task.yml#/TriggerType'
     example: onDemand,schedule,subscription,streaming
 
+withNotifications:
+  name: withNotifications
```
shortcuts, line 74:
```suggestion
withEmailNotifications:
  name: withEmailNotifications
```

better be descriptive than introduce bc when we add more notifications

specs/ingestion/common/schemas/task.yml
millotp: yes we have other policies in mind, for example the `safetyCheck` as you talked about on slack, to prevent a index being emptied if the `action` is replace.
Yes it would be nice to move `failureThreshold`, but it's a big breaking change

shortcuts: I was mostly wondering if the name made sense but I don't have any other in mind anyway

millotp: we had other naming ideas with @Fluf22 but this one made the most sense, maybe `advancedOptions` ?

shortcuts: to me those seems to mostly be error settings but maybe we can generalize it indeed, policies just feel like guidelines or things like that rather than personalization

PR #4310, opened 2025-01-06
fix(ruby): use discriminators for oneOf [skip-bc]

millotp:
## 🧭 What and Why

🎟 JIRA Ticket:

fixes https://github.com/algolia/algoliasearch-client-ruby/issues/507

Using discriminator to determine which class of oneOf to deserialize into.

Also cleanup some unused part of other templates.


shortcuts (approved):
looks PERFECTO

PR #4327, opened 2025-01-11
fix(specs): deprecate task v1 [skip-bc]

millotp:
## 🧭 What and Why

The tasks v1 api was deprecated 6 month ago


shortcuts (dismissed):
merci

shortcuts (approved):
nice catch

PR #4356, opened 2025-01-13
fix(specs): ingestion expected and received events type [skip-bc]

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

those fields are now always present in the response


morganleroi (approved):
Nice !

PR #4358, opened 2025-01-14
fix(specs): recommendHit is generic

millotp:
## 🧭 What and Why

🎟 JIRA Ticket:

Make the `RecommendHit` generic.

### DIFF
--- a/generators/src/main/java/com/algolia/codegen/cts/manager/JavaCTSManager.java
+++ b/generators/src/main/java/com/algolia/codegen/cts/manager/JavaCTSManager.java
@@ -46,5 +46,9 @@ public class JavaCTSManager implements CTSManager {
   public void addDataToBundle(Map<String, Object> bundle) throws GeneratorException {
     bundle.put("packageVersion", getVersion());
     bundle.put("import", Helpers.camelize(this.client).toLowerCase());
+    bundle.put("defaultGeneric", "Hit.class");
+    if (this.client.equals("recommend")) {
+      bundle.put("defaultGeneric", "RecommendHit.class");
+    }
   }
 }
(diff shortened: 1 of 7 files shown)


generators/src/main/java/com/algolia/codegen/cts/manager/JavaCTSManager.java
```diff
@@ -46,5 +46,9 @@ public void addSnippetsSupportingFiles(List<SupportingFile> supportingFiles, Str
   public void addDataToBundle(Map<String, Object> bundle) throws GeneratorException {
     bundle.put("packageVersion", getVersion());
     bundle.put("import", Helpers.camelize(this.client).toLowerCase());
+    bundle.put("defaultGeneric", "Hit.class");
+    if (this.client.equals("recommend")) {
+      bundle.put("defaultGeneric", "RecommendHit.class");
+    }
```
shortcuts, line 52: should it be controlled at the spec level with something like `x-defaultGeneric` maybe? I guess that's rare enough to not optimize it

generators/src/main/java/com/algolia/codegen/utils/GenericPropagator.java
```diff
@@ -172,18 +176,22 @@ public static void propagateGenericsToModels(Map<String, ModelsMap> modelsMap, b
 
     Map<String, CodegenModel> models = convertToMap(modelsMap);
 
-    for (CodegenModel model : models.values()) {
-      markPropagatedGeneric(model, m -> m.getVars(), skipOneOf);
-      markPropagatedGeneric(model, m -> m.getRequiredVars(), skipOneOf);
-    }
+    // we don't know in which order the model will come, so we iterate multiple times to make sure
+    // models at any depth are propagated
+    for (int i = 0; i < 5; i++) {
```
shortcuts, line 181: 🧙🏻

PR #4359, opened 2025-01-15
fix(specs): remove the processing outcome

millotp:
## 🧭 What and Why

Following https://github.com/algolia/data-ingestion/pull/4200, the `processing` run outcome can no longer be returned by the API.


Fluf22 (approved):
✅

PR #4360, opened 2025-01-15
 fix(specs): extend Analytics descriptions

gazconroy:
## 🧭 What and Why

Extending the Analytics OAS to cover what was previously in the old (non-OAS) spec.

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DOC-1578

### Changes included:

- Updating and extending some of the descriptions
- Adding an operation to the `revenue` tag


millotp (approved):
thanks !

PR #4361, opened 2025-01-15
chore(scripts): only warn about breaking changes

millotp:
## 🧭 What and Why

It's very annoying to add `[skip-bc]` to all PR because we slowed down our release cycle.
Now it will only be a warning in the sticky comment.


Fluf22 (approved):
thankyou.gif

Fluf22 (approved):
Let's go

PR #4363, opened 2025-01-16
fix(specs): subscription specific action

Fluf22:
## 🧭 What and Why

🎟 JIRA Ticket: relates to [DI-3278]

### Changes included:

- add new indexing operation for realtime updates linked to an e-comm task

### Warning

Wait for the definitive name before merging

[DI-3278]: https://algolia.atlassian.net/browse/DI-3278?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


specs/ingestion/common/schemas/task.yml
```diff
@@ -24,6 +24,8 @@ Task:
       $ref: '#/failureThreshold'
     action:
       $ref: '#/ActionType'
+    subscriptionAction:
+      $ref: '#/ActionType'
```
shortcuts, line 28: it only accepts a subset no?

Fluf22: I don't see any reason to prevent users from using all the action types from the API

PR #4389, opened 2025-01-22
fix(dart): add readme for composition

millotp:
## 🧭 What and Why

The release is failing because a Dart package needs a README.


Fluf22 (approved):
✅

PR #4390, opened 2025-01-22
chore(scripts): add the contrib in the github release

millotp:
## 🧭 What and Why

So that the contributor can appear in the github release and be tagged.
Also add a test to this and found a bug !


Fluf22 (approved):
✅

Fluf22 (approved):
OK

PR #4391, opened 2025-01-22
guides: add missing snippets to docs

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-3429

### Changes included:

add some missing snippets found in https://github.com/algolia/doc/pull/9682, more to come I guess but this unblocks me for https://github.com/algolia/doc/pull/9682


generators/src/main/java/com/algolia/codegen/cts/tests/SnippetsGenerator.java
```diff
@@ -101,7 +101,7 @@ public void run(Map<String, CodegenModel> models, Map<String, CodegenOperation>
         continue;
       }
 
-      List<Snippet> ops = Arrays.stream(snippets.get(operationId)).filter(r -> r.isSnippet).toList();
+      List<Snippet> ops = Arrays.stream(snippets.get(operationId)).toList();
```
millotp, line 104: this is not needed anymore, we are guarented to have at least 1 no ?

scripts/formatter.ts
```diff
@@ -57,7 +57,7 @@ export async function formatter(language: string, cwd: string): Promise<void> {
         'poetry lock && poetry sync --no-root && pip freeze > requirements.txt && poetry run ruff check --fix --unsafe-fixes && poetry run ruff format',
         { cwd, language },
       );
-      if (!cwd.includes('tests')) {
+      if (!cwd.includes('tests') && !cwd.includes('snippets')) {
```
millotp, line 60: why ? this is the whole purpose of having snippets

millotp (commented):
maybe we should reorder the CTS to make sure the one we want to display in the doc is at the top

generators/src/main/java/com/algolia/codegen/cts/tests/SnippetsGenerator.java
shortcuts: true

scripts/formatter.ts
shortcuts: ah yes it was to check something I can remove it now

Fluf22 (dismissed):
Ok

millotp (approved):

PR #4393, opened 2025-01-23
fix(specs): add a linter to assert that type is present

millotp:
## 🧭 What and Why

Following #4392, we can use a custom linter to verify that type is present and matches what we expect.


Fluf22 (approved):
Let's go!

Fluf22 (approved):

PR #4396, opened 2025-01-24
guides: add snippets for parameters

millotp:
## 🧭 What and Why

This add all the snippets that we can generate for `searchSingleIndex` and `setSettings` under https://github.com/algolia/doc/tree/master/app_data/parameters


Fluf22 (approved):
That's just amazing

PR #4454, opened 2025-02-11
fix(scripts): do not remove files before push

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

those scripts are pretty stable and we don't want to remove other people content if any


Fluf22 (approved):
🆗

PR #4455, opened 2025-02-11
fix(scripts): cleanup guides gen before generation

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

if we rename of remove existing guides/snippets they won't be deleted, this should take care of it


Fluf22 (dismissed):
Thanks!

Fluf22 (approved):
👌🏻

PR #4460, opened 2025-02-15
fix(java): saveObjects always passing waitForTasks as false

Giglium:
## 🧭 What and Why

The `com.algolia.api.SearchClient#saveObjects(String, Iterable, boolean, RequestOptions)` method takes the `waitForTasks` parameter, but it is ignored and always set to `false`.

🎟 JIRA Ticket:

### Changes included:

- Fixed the method by passing the `waitForTasks` parameter instead of hardcoding `false`

## 🧪 Test


shortcuts (approved):
Ohhhh great catch thanks!!

PR #4479, opened 2025-02-18
fix(csharp): remove specific minor version for System deps

morganleroi:
## 🧭 What and Why

[DI-3477](https://algolia.atlassian.net/browse/DI-3477)

Remove minor version for System dependencies. 

Fixes algolia/algoliasearch-client-csharp#866

[DI-3477]: https://algolia.atlassian.net/browse/DI-3477?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


templates/csharp/netcore_project.mustache
```diff
@@ -39,9 +39,9 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.Logging" Version="[8.0.0, 10.0.0)" />
-    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="[8.0.0, 10.0.0)" />
-    <PackageReference Include="System.Text.Json" Version="[8.0.0, 10.0.0)" />
+    <PackageReference Include="Microsoft.Extensions.Logging" Version="[8.0, 10.0)" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="[8.0, 10.0)" />
+    <PackageReference Include="System.Text.Json" Version="[8.0, 10.0)" />
```
millotp, line 44: should it be `8.0.1` to exclude the vulnerable version ? I think this is what the issue is complaining about

millotp (commented):
nice !

templates/csharp/netcore_project.mustache
morganleroi: If I put a specific minor (`8.0.5` to avoid the security alert) it would be a breaking change because If someone updates the client then he will need to upgrade the System.Text.Json if their version if below `8.0.5`. 

With this change, they can decide wether or not they want to update their `System.Text.Json` deps. (But Algolia lib will alway try to pull the latest minor now.)

millotp: okay but this doesn't change anything to the package resolution, charp chooses the min version, not the max.
I tried with `8.0.0` and `8.0` in the CTS, in both case the resolved version is `8.0.0` in `obj/project.assets.json`, so the vulnerability is still there

morganleroi: You're right.

millotp (approved):

PR #4563, opened 2025-03-05
chore(ci): always run all clients

millotp:
## 🧭 What and Why

The improvements thanks to only running client that has changes is dwarfed by swift being so slow, we don't need this anymore.
And it's causing issues with guides because they are not in 1 folder by client


shortcuts (approved):
YES

shortcuts (approved):
all thanks swift for being slow

PR #4588, opened 2025-03-11
fix(clients): implement Is on custom errors [skip-e2e]

millotp:
## 🧭 What and Why

This allow users to use `errors.Is` on our custom errors


shortcuts (approved):
🙆🏻‍♂️

PR #4612, opened 2025-03-17
fix(python): npe in request options merge method

Fluf22:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

- `data` can be `None` in `request_options`

### DIFF
--- a/clients/algoliasearch-client-python/algoliasearch/http/request_options.py
+++ b/clients/algoliasearch-client-python/algoliasearch/http/request_options.py
@@ -89,6 +89,9 @@ class RequestOptions:
                 _user_request_options = user_request_options
 
             for key, value in _user_request_options.items():
-                request_options[key].update(value)
+                if request_options[key] is None:
+                    request_options[key] = value
+                else:
+                    request_options[key].update(value)
 
         return self.from_dict(request_options)


clients/algoliasearch-client-python/algoliasearch/http/request_options.py
```diff
@@ -89,6 +89,9 @@ def merge(
                 _user_request_options = user_request_options
 
             for key, value in _user_request_options.items():
-                request_options[key].update(value)
+                if request_options[key] is None:
```
shortcuts, line 92: L68 there's a default for the None parameters, maybe do the same for data?

Fluf22: I was hesitant because in other languages, if there are data set in the request options, we overwrite the original object.
It's not the case in python (yet), but I thought it was more safe to handle it carefully that way, to prevent any misbehaviour if we make the code more consistent with the other languages in the future

shortcuts: I don't think we will go to that level of consistency honestly it seems too much but your solution works as well so i'm good with it!

shortcuts (approved):
ty and good catch!

PR #4613, opened 2025-03-17
feat(clients): add new realtime-personalization api

benamib:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/PRED-3652

### Changes included:

- Add new API specs for Realtime Perso

### TODO
- [ ] Update Doc link


specs/realtime-personalization/spec.yml
```diff
@@ -0,0 +1,44 @@
...
+  - url: https://ai-personalization.{region}.algolia.com
+    variables:
+      region:
+        enum:
+          - us
+          - eu
+        default: us
+security:
+  - appId: []
+    apiKey: []
```
bengreenbank, line 27: We can use the common schemas:

https://github.com/algolia/api-clients-automation/blob/12511129f79635815074086dbb63a93542222f3b/specs/advanced-personalization/spec.yml#L60-L65

specs/realtime-personalization/spec.yml
```diff
@@ -0,0 +1,44 @@
...
+  description: |
+    Related guide: Algolia Realtime Personalization.
+components:
+  securitySchemes:
+    appId:
+      $ref: '../common/securitySchemes.yml#/appId'
+    apiKey:
+      $ref: '../common/securitySchemes.yml#/apiKey'
+servers:
+  - url: https://ai-personalization.{region}.algolia.com
```
bengreenbank, line 18: We can add a description:

https://github.com/algolia/api-clients-automation/blob/12511129f79635815074086dbb63a93542222f3b/specs/advanced-personalization/spec.yml#L70-L71

specs/realtime-personalization/common/schemas/personalizationFilters.yml
```diff
@@ -0,0 +1,20 @@
+personalizationFilters:
```
bengreenbank, line 1: I don't think we should call them personalization filters because we expect that the response will grow in the future to include more than that, for example, the raw affinities, like we do in Advanced Perso:

https://github.com/algolia/api-clients-automation/blob/12511129f79635815074086dbb63a93542222f3b/specs/advanced-personalization/common/schemas/User.yml#L1-L30

Perhaps just something like user profile response.

specs/realtime-personalization/spec.yml
benamib: It's already available a little bit above : https://github.com/algolia/api-clients-automation/pull/4613/files#diff-d74ad6c5ebf69c953183cce9f102977cf72830dc00f217da413872dc9ea9e041R11

I reproduced what was already done in the previous api

specs/realtime-personalization/common/schemas/personalizationFilters.yml
benamib: Would `user` be a better name in our case ?

specs/realtime-personalization/common/schemas/user.yml
```diff
@@ -0,0 +1,44 @@
...
+      description: User ID of the user.
+    search:
+      type: object
+      description: |
+        Index personalization filters by index name.
+      additionalProperties:
+        $ref: '#/searchFilters'
+
+searchFilters:
+  type: object
```
bengreenbank, line 23: Does this allow us to store search configurations for multiple groups of indices, and have a dynamic name?

Example extracted from Francois' presentation:
Screenshot 2025-03-19 at 17 19 59

Not sure how we represent that in the spec though.

benamib: The attribute `additionalProperties` allows us to tell the generator that the key of the object can be dynamic but that the format of the value should be of type `searchFilters`. 

I'm not fond of this method but we don't have too much choice to respect the response we want. 
We could still challenge the response format to replace the objects by arrays which would make the specs clearer imo like: 
```
...
"search": [
	{ "aliases": "abc", "strategy": "def", "filters": {...} }
]
...
```

bengreenbank: I believe the reason for this structure was so that the FE could retrieve the required attributes as quick as possible (it's more time consuming to filter an array based on a key than it is to get a specific key from a map).

This looks good for now, but if it causes issues we can rethink it

shortcuts (approved):

specs/realtime-personalization/common/parameters.yml
```diff
@@ -0,0 +1,8 @@
+# path
+UserToken:
+  name: userToken
```
francoischalifour, line 3: We call this `userID` in all the personalization APIs.

specs/realtime-personalization/common/parameters.yml
```diff
@@ -0,0 +1,8 @@
...
+  in: path
+  required: true
+  description: Unique identifier representing a user for which to fetch the personalization profile.
```
francoischalifour, line 6:
```suggestion
  description: Unique identifier used to retrieve the personalization profile of a specific user.
```

specs/realtime-personalization/spec.yml
```diff
@@ -0,0 +1,51 @@
...
+        enum: [eu, us]
+        default: eu
+security:
+  - appId: []
+    apiKey: []
+tags:
+  - name: users
+    x-displayName: Users
+    description: |
+      User profiles contains the search filters each user has for the different facets in your index.
```
francoischalifour, line 32:
```suggestion
      User profiles contain search filters that personalize search results in real time.
```

PR #4618, opened 2025-03-21
fix(python): missing Decimal to float conversion for body serializer

Enase:
## 🧭 What and Why
Added additional verification that converts Decimal to float in request body serializer.

Fixes ability to serialize and save Django models with DecimalField, resolves https://github.com/algolia/algoliasearch-django/issues/340


shortcuts (approved):
thanks a lot!!!!!

PR #4659, opened 2025-03-26
chore: reduce manual intervention on javascript clients

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket:

### Changes included:

it requires manual intervention from an apic maintainer when someone wants to add a new client, because there's a bunch of condition for standalone clients to not be part of the `algoliasearch` package during the pre-release phase.

this pr adds a boolean to determine if the package should be included in `algoliasearch` or not (see composition), in order to eventually reduce the manual work required on that front.

### DIFF
--- a/config/clients.config.json
+++ b/config/clients.config.json
@@ -144,7 +144,8 @@
       },
       {
         "name": "composition",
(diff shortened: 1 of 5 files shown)


Fluf22 (approved):
✅

PR #4696, opened 2025-04-01
chore(scripts): reuse java-version

millotp:
## 🧭 What and Why

Use the `.java-version` from the client to make sure it's valid, otherwise we only test it during a release at the very end.
Also use the same file for local dev.


shortcuts (dismissed):
do we need to update renovate?

shortcuts (approved):

PR #4698, opened 2025-04-02
fix(ruby): remove integer empty check

lucas-aragno:
## 🧭 What and Why

Maybe I'm missing something but after upgrading to the latest version of the gem my timeout settings started breaking:

```ruby
::Algolia::SearchClient.create(
    app_id,
    api_key,
    {
      connect_timeout: 5_000
    }
  )
```

The error seemed to be caused by the new `empty?` checks that were added previous to the client initialization. Why are we now checking for both `nil?` and `empty?` is the idea to use strings now? that doesn't seem to be the case since the default seem to be numbers. So let me know if this was an error or maybe we need to update the documentation on how to set up timeouts.

Thanks!

🎟 JIRA Ticket: N/A

### Changes included:

- List changes
   - Removing .empty? checks on integer settings

### DIFF
--- a/clients/algoliasearch-client-ruby/lib/algolia/api/abtesting_client.rb
+++ b/clients/algoliasearch-client-ruby/lib/algolia/api/abtesting_client.rb
@@ -38,15 +38,15 @@ module Algolia
     end
 
     def self.create_with_config(config)
-      if config.connect_timeout.nil? || config.connect_timeout.empty?
+      if config.connect_timeout.nil?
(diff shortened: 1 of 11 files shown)


millotp:
Hey, thanks for your contribution, this is a miss indeed !
The files under `clients/algoliasearch-client-ruby` are generated from the template, can you remove them from your PR pls ?

lucas-aragno:
@millotp done!

PR #4702, opened 2025-04-03
fix(specs): userData is any type

millotp:
## 🧭 What and Why

fixes https://github.com/algolia/algoliasearch-client-go/issues/769

🎟 JIRA Ticket: [DI-3692](https://algolia.atlassian.net/browse/DI-3692)

The `userData` of the index settings can be anything since it's not serialized by the API, I tried with a string, an array, and a object and they are all accepted by the API.

I found out that we can generate the `any` type in all languages by just omitting the `type`, but this means its also nullable, not sure how to remove that.

[DI-3692]: https://algolia.atlassian.net/browse/DI-3692?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/specs/common/schemas/IndexSettings.yml
+++ b/specs/common/schemas/IndexSettings.yml
@@ -768,7 +768,6 @@ hitsPerPage:
     - Pagination
 
 userData:
-  type: object
   example:
     settingID: f2a7b51e3503acc6a39b3784ffb84300
     pluginVersion: 1.6.0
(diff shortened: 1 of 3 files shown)


shortcuts (dismissed):
HUGE

let's see if it works for any language 🥶

shortcuts (approved):
EASY

PR #4725, opened 2025-04-08
docs(personalization): add rate limit information for personalization API

Keylor42:
## 🧭 What and Why
Document the new rate limit for the personalization API.

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/PRED-3537

### Changes included:

- Section added in specs/personalization/spec.yml that details the rate limit and the related headers.

## 🧪 Test
N/A

### DIFF
--- a/specs/personalization/paths/deleteUserProfile.yml
+++ b/specs/personalization/paths/deleteUserProfile.yml
@@ -14,6 +14,13 @@ delete:
   responses:
     '200':
       description: OK
+      headers:
+        x-ratelimit-limit:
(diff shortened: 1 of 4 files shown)


millotp (dismissed):
thanks for the contribution !
You can also specify the headers for each endpoint directly, like in [abtests](https://github.com/algolia/api-clients-automation/blob/main/specs/abtesting/paths/abtest.yml#L14).
This is not used in the doc generation, but it can be used in the documentation to know for each endpoint which headers are returned.

millotp (approved):
perfecto

PR #4726, opened 2025-04-09
fix(specs): condition cant be numeric filter

kai687:
## 🧭 What and Why

Amend the description for the 'condition' object about the limitations of the 'filter' parameter.

🎟 JIRA Ticket: N/A

### Changes included:

- Fix a typo
- Add that you can't use numeric filters or combinations of filters with `OR`


Fluf22 (approved):
Thanks ✅

PR #4732, opened 2025-04-12
chore(playground): always use local version of the go client

millotp:
## 🧭 What and Why

Even if renovate wants to update the version in the go.mod, it will still use the local version.
This was not the case for the playground, it was using a remote one.


Fluf22 (approved):
Easy one

PR #4756, opened 2025-04-14
fix(clients): correctly deserialize SearchResult

millotp:
## 🧭 What and Why

🎟 JIRA Ticket: [CR-7011](https://algolia.atlassian.net/browse/CR-7011) [CR-8307](https://algolia.atlassian.net/browse/CR-8307)

The `SearchResult` deserialization is not working in java because the generic type `T` is getting erased by java, we need to use all the information about the type to deserialize it into the proper type.


Fluf22 (approved):
Thanks 🙏🏻

shortcuts (approved):
lovely

PR #4784, opened 2025-04-25
feat(scripts): add push to mcp-node on release

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket:

### Changes included:

This PR adds mcp-node to the repository to spread the specs to, it also updates the support for JSON output as this is what's expected there.


specs/abtesting/paths/abtests.yml
```diff
@@ -57,6 +57,7 @@ get:
   tags:
     - abtest
   operationId: listABTests
+  x-mcp-tool: true
```
shortcuts, line 60: as per @kombucha's comment this might be easier/better to handle at the mcp repository level, unless we see some documentation purposes that we want to reflect on our API reference pages?

millotp (approved):
gg !

specs/abtesting/paths/abtests.yml
morganleroi: Could be indeed reflected in the documentation at some point! I'd stick with your solution !

morganleroi (approved):
God save the OpenAPI spec!

specs/abtesting/paths/abtests.yml
kombucha: I think this metadata should leave in the source of truth (ie this repository) eventually, I just thought we might want to experiment _first_ in the mcp repo while we're figuring things out before doing it here; especially if we start defining mcp specific description.
Also we might want to think whether we want to opt-in endpoints vs opt-out depending on the ratio

shortcuts: you see that as a non blocker so? We can also make it clearer that this is where the source of truth lives if a PR is opened on the mcp side as well, the good part of having it here is mostly reflecting it from the documentation

PR #4800, opened 2025-04-30
docs: rework the new client section and small guide on pre-releases

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-3720

### Changes included:

It wasn't clear for some Algolians that it is possible to release pre-releases (in JavaScript) and how to do so, so I've reworked our doc a bit plus added a really simple guide on how to do so just so we can point it to people.

Also fixes the generator to default unknown versions to pre-releases, as this should be less common now.


millotp (approved):
looks good !

PR #4801, opened 2025-04-30
feat(swift): generate composition client

Fluf22:
## 🧭 What and Why

🎟 JIRA Ticket: [DI-3716]

### Changes included:

- enable the `composition` client in the swift generator config
- add the `composition` client in the reserved name check so all models are prefixed
- add a prefix lambda for response and generic types in tests

## 🧪 Test
- `composition` client tests are enabled


millotp (commented):
nice !
There are 6 models that starts with `CompositionCompisition`, like `CompositionCompositionRunSearchResponse`, is it possible to not add the prefix in this case ?

Fluf22:
Sadly, the prefixing is required on all models to avoid conflicts with other models from the client. eg:
- BaseSearchResponse -> CompositionBaseSearchResponse
- CompositionBaseSearchResponse -> CompositionCompositionBaseSearchResponse

millotp (approved):

PR #4805, opened 2025-05-02
chore(test): check that the response can contain unknown field

millotp:
## 🧭 What and Why

Add a test that assert that unknown fields doesn't make the deserialization fail, at the root of the object, or nested in a oneOf inside an array inside a object.


scripts/cts/testServer/algoliaMock.ts
```diff
@@ -37,6 +37,51 @@ function addRoutes(app: Express): void {
       processingTimeMS: 0,
     });
   });
+
+  // languages that just put the response in a map, there is no strict parsing or types to match.
+  const isLaxLanguage = (lang: string) => {
+    return lang === 'dart' || lang === 'javascript' || lang === 'python' || lang === 'php';
```
shortcuts, line 43: can this be moved as an option in the tests field of the clients config? I tried to update some scripts to remove those inline conditions where possible, having the json schema there as well helps us better describing things

shortcuts (approved):
idk about the swift part but the rest looks wonderful

PR #4819, opened 2025-05-05
feat(specs): enrich response for list composition endpoint

ClaraMuller:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/CMP-424

### Changes included:

- List changes

List composition endpoint response is enriched with response fields added since:
* https://github.com/algolia/metis/pull/11213
* https://github.com/algolia/metis/pull/11214

## 🧪 Test

* CI
* Locally run
    * `yarn cli build specs all`

### DIFF
--- a/specs/composition-full/common/schemas/listCompositionsResponse.yml
+++ b/specs/composition-full/common/schemas/listCompositionsResponse.yml
@@ -11,6 +11,18 @@ listCompositionsResponse:
       type: integer
       description: Number of pages.
       example: 100
+    page:
+      type: integer
+      description: Current page.
+      example: 0
+    hitsPerPage:
+      type: integer
+      description: Number of items per page.
+      example: 20
+    nbHits:
+      type: integer
+      description: Number of items.
+      example: 200
   required:
     - items
     - nbPages


Fluf22 (approved):
✅

millotp:
> Sorry I just rebased as Github was complaining about not being up to date with `main`

no worries, this is required because the CI will push a generated commit once this is merged

ClaraMuller:
Ok I see. I checked if I have to generate other stuff please do not hesitate if some stuff are missing in this PR 🙏

millotp:
I think this is all good 😄

specs/composition-full/common/schemas/listCompositionsResponse.yml
```diff
@@ -11,6 +11,18 @@ listCompositionsResponse:
...
+      description: Current page.
+      example: 0
+    hitsPerPage:
+      type: integer
+      description: Number of items per page.
+      example: 20
+    nbHits:
+      type: integer
+      description: Number of items.
+      example: 200
```
e-krebs, line 25: I'm a bit late to the party, but shouldn't those be required as they are always returned @ClaraMuller ?
cc: @millotp

e-krebs: (the Typescript types [generated by the CI](https://github.com/algolia/api-clients-automation/commit/ca622392b3b005f87e9035ada74027418414870c) are optional which made me realize)

ClaraMuller: What would happen if the field are not in the response (for whatever reason, for example due to a bug).
I would have though that by default no fields is required in the response to prevent having issue in the client side if something is going wrong in the backend.
But I don't have a strong opinion, I will update if we think this is important.

e-krebs: > What would happen if the field are not in the response (for whatever reason, for example due to a bug).

There would be an error because we wouldn't handle it.
But if there's a bug in the backend, that's fine imho.

The specs should represent the expected shape of data when there is no bug.
Otherwise it's unmanageable on the frontend to handle all cases when all data is optional.

ClaraMuller: I will make them required then.

PR #4820, opened 2025-05-06
fix(clients): InvalidOperationException in C# Fill() due to unsafe dictionary enumeration

reneemundie:
## 🧭 What and Why

Fixes a concurrency issue in the generated C# client (HttpRequestHeadersExtensions.Fill) where a shared IDictionary<string, string> is enumerated without being cloned. This causes intermittent InvalidOperationException under load when the collection is modified during enumeration.

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-3456

### Changes included:

- Clone the dictionary with .ToList() before enumeration to prevent unsafe concurrent iteration

## 🧪 Test

- Verified this change compiles in local environment
- Fix has already been validated against the customer case and linked GitHub issue ([algoliasearch-client-csharp#865](https://github.com/algolia/algoliasearch-client-csharp/issues/865))
- Safe, non-breaking change


morganleroi (approved):
thanks!

PR #4849, opened 2025-05-13
feat(clients): flatten body parameters

millotp:
## 🧭 What and Why

Flatten body parameters when possible to avoid having to wrap properties in a model.
To avoid breaking changes and avoid releasing a new major, all the new code lives under `algolia/next`, and the previous method are not generated anymore, they just live here.


shortcuts:
yooooo @millotp keep it for HISTORY

PR #4901, opened 2025-05-30
feat(specs): update transformation specs for no-code

mehmetaligok:
## 🧭 What and Why

We are updating the transformation endpoints to support the upcoming no-code transformations. 

### Changes included:

- Both request and response formats updated.
- `type` field added with an enum.
- `input` field added for both types. 

**NOTE:** The no-code input is left as a generic input because there is a need for recursion, but the generation doesn't seem to work properly for this complex structure. 

We expect this to be used internally, so we will move forward with a generic field for now.

### DIFF
--- a/clients/algoliasearch-client-python/algoliasearch/abtesting/client.py
+++ b/clients/algoliasearch-client-python/algoliasearch/abtesting/client.py
@@ -79,9 +79,9 @@ class AbtestingClient:
             config = AbtestingConfig(
                 transporter.config.app_id, transporter.config.api_key, region
             )
-
-        if config is None:
+        elif config is None:
             config = AbtestingConfig(app_id, api_key, region)
+
         self._config = config
         self._request_options = RequestOptions(config)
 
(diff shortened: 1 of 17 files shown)


shortcuts (commented):
nice

specs/ingestion/common/schemas/transformation.yml
```diff
@@ -25,9 +30,49 @@ Transformation:
...
+        $ref: '#/TransformationNoCodeStep'
+  required:
+    - steps
+
+TransformationNoCodeStep:
+  type: object
+  description: A step in a no-code transformation.
+
 Code:
   type: string
```
shortcuts, line 74:
```suggestion
  type: string
  deprecated: true
```

I guess we can move it there now as well

mehmetaligok: sure, it is moved with https://github.com/algolia/api-clients-automation/pull/4901/commits/ea0503cc9d194645cadba27f6905cdca26296491

shortcuts (approved):
perfect gg!

PR #4922, opened 2025-06-02
feat(clients): add generic parameter

millotp:
## 🧭 What and Why

Some parameters deserve to be generic, to make it easier for the user.
Starting with `saveObject`.

This is similar to a previous attempt a long time ago: https://github.com/algolia/api-clients-automation/pull/3151


shortcuts (approved):
so nice

PR #4923, opened 2025-06-02
feat(python): re-export models from __init__.py files

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-3872

### Changes included:

adds long awaited exports to the init files, in order to use the imports of the models


millotp (dismissed):
thanks a lot for this ❤️

PR #4929, opened 2025-06-03
fix(specs): add titles to oneOf options

kai687:
## 🧭 What and Why

Add `title` properties to the `oneOf` options for `highlightResult` and `snippetResult` in the search response.

Without them, our work-in-progress [Mintlify docs](https://algolia2-feat-apis.mintlify.app/doc/rest-api/search/search/search-single-index) just show 'option 1', 'option 2' etc. in the `oneOf` selector.

🎟 JIRA Ticket:

### Changes included:

- List changes


shortcuts:
Looks great! Is this required for every oneOf? We could make it a custom eslint rule if necessary

PR #4941, opened 2025-06-06
fix(scripts): push to repository

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket:

### Changes included:

the push silently fails https://github.com/algolia/api-clients-automation/actions/runs/15463968082/job/43532734476 with `fatal: could not read Username for 'https://github.com/': No such device or address`

works properly now https://github.com/algolia/api-clients-automation/actions/runs/15486841829/job/43603897640?pr=4941#step:12:1

### DIFF
--- a/scripts/common.ts
+++ b/scripts/common.ts
@@ -197,11 +197,14 @@ export async function runComposerInstall(): Promise<void> {
 }
 
 export function ensureGitHubToken(): string {
+  const githubToken = process.env.GITHUB_TOKEN || process.env.GH_TOKEN || '';
+
   // use process.env here to mock with vitest
-  if (!process.env.GITHUB_TOKEN) {
-    throw new Error('Environment variable `GITHUB_TOKEN` does not exist.');
+  if (githubToken == '') {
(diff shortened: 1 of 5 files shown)


scripts/common.ts
```diff
@@ -197,11 +197,14 @@ export async function runComposerInstall(): Promise<void> {
 }
 
 export function ensureGitHubToken(): string {
+  const githubToken = process.env.GITHUB_TOKEN || process.env.GH_TOKEN || '';
```
millotp, line 200: should you check that they are both present ?

shortcuts: the doc is a bit vague, https://cli.github.com/manual/gh_auth_login#:~:text=To%20use%20gh%20in%20GitHub%20Actions%2C%20add%20GH_TOKEN%3A%20%24%7B%7B%20github.token%20%7D%7D%20to%20env. states that it reads GH_TOKEN on the CI, however we might use GITHUB_TOKEN locally, costs nothing to have both

millotp: okay so it expects `GITHUB_TOKEN` locally but `GH_TOKEN` on the CI, this is super weird

PR #4944, opened 2025-06-06
feat(specs): add user profile computation & amend existing user profile endpoints [PRED-3876]

bengreenbank:
## 🧭 What and Why

🎟 JIRA Ticket: PRED-3876

See [RFC](https://algolia.atlassian.net/wiki/spaces/PREDICT/pages/5933728059/RFC+Realtime+Perso+Partial+Engine+Integration)

### Changes included:
- Add the new endpoint described in the linked RFC
- Amend the existing endpoint as described in the RFC

## 🧪 Test
n/a


shortcuts (approved):
awesome

francoischalifour (approved):

shortcuts (dismissed):

raed667:
Should we merge `realtime-personalization` client into `personalization`?

The fact that it is "real-time" is just a technical detail and not a different product.

bengreenbank:
> Should we merge `realtime-personalization` client into `personalization`?
> 
> The fact that it is "real-time" is just a technical detail and not a different product.

I'm not sure we can, since the [base URLs are different](https://github.com/algolia/api-clients-automation/blob/7e226c4559f6cff190632a8b15df9103f4e3fc15/specs/personalization/spec.yml#L11-L12).

@shortcuts can we support different base URLs within the same spec?

PR #4973, opened 2025-06-11
chore: skip e2e composition

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

seems like we have an issue on our metis app which prevents this e2e from running, we can skip it for now


shortcuts:
> qq: do you think it would be hard to implement a `skip` boolean in the test spec? 😬 Instead of fully removing the test
> 
> Otherwise tell me and I'll approve that I'm just afraid we won't ever put it back

I wanted to go with that first but then realized it would be something we might overuse for simplicity/easier for anyone to toggle so... I've went with removing and will add it back 🥶

PR #4974, opened 2025-06-11
feat(specs): update try transformation specs for no-code

mehmetaligok:
## 🧭 What and Why

A follow-up to https://github.com/algolia/api-clients-automation/pull/4901. 

This PR adds no-code transformations support to the try endpoint specs. 

### Changes included:

- The try endpoint spec is updated.

### DIFF
--- a/specs/ingestion/common/schemas/transformation.yml
+++ b/specs/ingestion/common/schemas/transformation.yml
@@ -107,8 +107,6 @@ TransformationCreate:
       $ref: '#/AuthenticationIDs'
   required:
     - name
-    - type
-    - input
 
 TransformationCreateResponse:
   type: object
@@ -165,8 +163,6 @@ TransformationTry:
       items:
         $ref: './authentication.yml#/AuthenticationCreate'
   required:
-    - type
-    - input
     - sampleRecord
 
 TransformationTryResponse:
--- a/specs/ingestion/common/schemas/transformation.yml
+++ b/specs/ingestion/common/schemas/transformation.yml
@@ -153,6 +153,10 @@ TransformationTry:
   properties:
     code:
       $ref: '#/Code'
+    type:
+      $ref: '#/TransformationType'
+    input:
+      $ref: '#/TransformationInput'
     sampleRecord:
       description: The record to apply the given code to.
       type: object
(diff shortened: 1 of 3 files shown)


specs/ingestion/common/schemas/transformation.yml
```diff
@@ -161,7 +165,8 @@ TransformationTry:
       items:
         $ref: './authentication.yml#/AuthenticationCreate'
   required:
-    - code
+    - type
+    - input
```
Fluf22, line 169: I don't think we should require them here
We could create another object where it's required, but here, it would be a bc I think

mehmetaligok: Sorry, I didn't get it. 

The `code` field was previously required; now, instead of the `code` field, we want to use `type` and `input`. So that is why I made them required, because the endpoint wouldn't work without them.

Fluf22: Yes, but if people use this prop in their code, it will not work anymore.
For example, if you have this in your JS code:
```js
const transformation: Transformation = { code: '' }
```
Typescript will be yelling at you because you don't implement `type` and `input`, when you'll upgrade to the latest version

millotp: it still works with the legacy payload right ? @Fluf22 is right, this is a breaking change if you put them as required

Fluf22: @millotp my bad! I read yours too fast! I deleted it 😁

mehmetaligok: Okay. I didnt aware of this will effect and break the generated code. Btw, this is also the case for create payload so I need to revert that change. 

But what is the recommendation in here? Make nothing required? 

(The api supports legacy payload and the new payload at the same time. So there is no breaking change in the actual API.)

Fluf22: @millotp I'd suggest creating a second payload and using a oneOf? 
But maybe this would be bc in some languages?

Fluf22: Only other option is to make nothing required, yes

millotp: having a oneOf is breaking too, I would suggest making everything optional for now, and in a month or so, make the new payload required and leave `code` as optional.
We don't have a lot of way to deal with breaking changes because of the SLA...

millotp (approved):

PR #5012, opened 2025-06-17
fix(specs): fix type in realtime_user

raed667:
## 🧭 What and Why

Fix the current type `RealtimeUserUser` -> `RealtimeUser`


shortcuts (approved):
good now! thanks

PR #5029, opened 2025-06-24
fix(specs): tag for push

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-3930

### Changes included:

the tag is missing, so the operation isn't documented on the algolia docs


morganleroi (approved):
Thanks!

PR #5060, opened 2025-07-04
fix(specs): allow one sided rule validity

millotp:
## 🧭 What and Why

🎟 JIRA Ticket: [CR-8889](https://algolia.atlassian.net/browse/CR-8889)

Time ranges for rule can be one sided, we can't express that in openapi without using a one of, so we just make them both optional.

[CR-8889]: https://algolia.atlassian.net/browse/CR-8889?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


shortcuts:
> > this is a bc for go no?
> 
> indeed, and scala, but the spec was not correct so it's still a fix right ?

yup, just makes me think that we should have a way to properly reflect those in changelogs

PR #5082, opened 2025-07-07
fix(specs): correct base url info for Insights URL

kai687:
## 🧭 What and Why

Update the statement about `insights.algolia.io` for the docs.

🎟 JIRA Ticket:

### Changes included:

- List changes


specs/insights/spec.yml
```diff
@@ -18,7 +18,7 @@ info:
 
     - `https://insights.us.algolia.io`
     - `https://insights.de.algolia.io`
-    - `https//insights.algolia.io` (routes requests to the closest of the above servers, based on your geographical location)
+    - `https//insights.algolia.io` (alias of `insights.us.algolia.io`)
```
shortcuts, line 21: is this true? IIRC we read somewhere that it was geo routed to the closest server compared to e.g. analytics that defaults

shortcuts: see https://algolia.atlassian.net/wiki/spaces/EX/pages/4515987479/Internal+routes+of+the+Insights+API#Endpoints

shortcuts: not sure if up to date but worth asking

shortcuts: cc @Keylor42 maybe

PR #5087, opened 2025-07-08
feat(specs): document runMetadata parameter

DevinCodes:
## 🧭 What and Why

🎟 JIRA Ticket: DI-3625


specs/ingestion/common/schemas/source.yml
```diff
@@ -474,6 +474,12 @@ RunSourcePayload:
       description: List of entityIDs to update.
     entityType:
       $ref: '#/EntityType'
+    runInput:
+      type:
+        - type: string
+        - type: object
+        - type: array
```
DevinCodes, line 481: @millotp do you have any idea how I would describe a free-form parameter? Or do I not need to add that in the spec in that case?

PR #5088, opened 2025-07-09
chore(scripts): remove dotenv

millotp:
## 🧭 What and Why

The owner of dotenv wants to promote his paid feature, but actually we don't need the lib at all.
https://github.com/motdotla/dotenv/pull/877


shortcuts (approved):
Image

PR #5090, opened 2025-07-10
fix(specs): abtest v3 naming conflict with Metric type

cdhawke:
## 🧭 What and Why

image

The typescript types being generated for the v3 abtesting client don't match the object definitions in the openapi spec file. We think this is due to a naming collision when generating the types - `Metric` is defined elsewhere and being reused improperly.

### Changes included:

- rename the object reference in the spec

## 🧪 Test

N/A


leonardogavaudan (approved):

leonardogavaudan (dismissed):

leonardogavaudan (approved):
LGTM

PR #5109, opened 2025-07-14
Update api.mustache

alcohol:
The spec for these looks correct, so I am guessing this template is the cause for the erroneous docblocks in the generated client code.

## 🧭 What and Why

Incorrect type docblocks for parameters

🎟 JIRA Ticket:

### Changes included:

- Fixed docblocks


alcohol:
Thanks 😄

PR #5121, opened 2025-07-15
fix(specs): abtests metadata mean

cdhawke:
## 🧭 What and Why

image

The value is missing from the v3 definitions, but is returned from the API

### Changes included:

Add the property to the metadata metric definition


stevenMevans (approved):

shortcuts (approved):

cdhawke:
Could you release this then? it's blocking https://github.com/algolia/AlgoliaWeb/pull/25052 (I could add type overrides but wanted to avoid the tech debt and just fix it here)

PR #5122, opened 2025-07-16
Clara/refact/group files in dir

ClaraMuller:
## 🧭 What and Why

🎟 JIRA Ticket: Work preparation for https://algolia.atlassian.net/browse/CMP-484

Composition API team is preparing update of the API clients with writes endpoints (see full list of endpoints [here](https://algolia.atlassian.net/wiki/spaces/CMP/pages/5104042001/Endpoints)).
Goals of this PR is to cleanup / refactor the current Open API spec to ease the integrations of the new endpoints.

### Changes included:

➡️ No API client should be updated with this PR, it purely about reorganising code.

Create 3 new directories within `composition-full/schemas`:
- `components`: to group reusable models together
- `requestBodies`: to group and identify request bodies easily
- `response`: to group and identify request response easily

#### Next steps:
* Create `composition-full/params` to add an interface between shared parameter between  Composition API and Search API.

#### Review tips

PR can be review commit by commit for ease.

## 🧪 Test

* CI
* Locally run
    * `yarn cli build specs all`


specs/composition-full/common/schemas/SearchParams.yml
shortcuts, line 1: Does this still differ from the regular search params? If not, we can remove it and use the search spec instead

specs/composition-full/common/schemas/components/CompositionRule.yml
```diff
@@ -3,7 +3,7 @@ compositionRule:
   additionalProperties: false
   properties:
     objectID:
-      $ref: '../../../common/parameters.yml#/objectID'
+      $ref: '../../../../common/parameters.yml#/objectID'
```
shortcuts, line 6: should this use the composition objectID instead?

specs/composition-full/common/schemas/components/CompositionRule.yml
```diff
@@ -77,3 +77,8 @@ anchoring:
 
     Empty queries are only allowed as patterns with `anchoring: is`.
   enum: [is, startsWith, endsWith, contains]
+
+objectID:
+  type: string
+  example: 'my_composition_rule_object_id'
+  description: Composition rule unique identifier.
```
shortcuts, line 84: can this and the one in `specs/composition-full/common/schemas/components/Composition.yml` be two separated object with descriptive name in a file named `specs/composition-full/common/parameters.yml`? that's the pattern we use everywhere

specs/composition-full/common/schemas/components/Injection.yml
```diff
@@ -0,0 +1,953 @@
+mainInjectionQueryParameters:
+  title: Composition Main Injection Query parameters as object
+  type: object
+  additionalProperties: false
+  properties:
+    attributesToRetrieve:
```
shortcuts, line 6: if those are expected to be exactly the same as the one from the search API, you can import them from either the [Search Params](https://github.com/algolia/api-clients-automation/blob/main/specs/common/schemas/SearchParams.yml) or the [Index Settings](https://github.com/algolia/api-clients-automation/blob/main/specs/common/schemas/IndexSettings.yml), see [recommend spec example](https://github.com/algolia/api-clients-automation/blob/main/specs/recommend/common/schemas/IndexSettings.yml#L12)

specs/composition-full/common/schemas/components/CompositionRule.yml
shortcuts: (the name of the parameter file doesn't matter much i'm fine with the specs/composition-full/paths/urlParams.yml below or other but it's better if schemas are standalone)

shortcuts (commented):
clean

specs/composition-full/common/schemas/components/Injection.yml
shortcuts: ah maybe it's tackled by `Create composition-full/params to add an interface between shared parameter between Composition API and Search API.`?

specs/composition-full/common/schemas/components/CompositionRule.yml
ClaraMuller: Ah it as actually the previous format, sorry.
I will revert this part.
Bu why not grouping schema that lives together in the same file ? After all the Composition rule objectID and the Composition ObjectID exists only within a context of a largeur model. By themselves they have no meanings.

shortcuts: it might be easier to manage at a spec level indeed but when searching for `objectID` in every specs for example, it's better if we understand the difference by just having a peak at the variable name

for objects of primitive type it won't generate a model so we can make the name as specific as possible without impacting the code

specs/composition-full/common/schemas/SearchParams.yml
ClaraMuller: The parameter are the same than the Search API but the usahe is not allowed in all part of the API.
We are way mores restricted.
I will try to clarify the usage in my next PR

specs/composition-full/common/schemas/components/CompositionRule.yml
ClaraMuller: Indeed this is a bad ref 👍

PR #5125, opened 2025-07-16
refactor(specs): mutualise code between for search query parameter

ClaraMuller:
## 🧭 What and Why

🎟 JIRA Ticket:

### Changes included:

- List changes

### DIFF
--- a/specs/common/schemas/IndexSettings.yml
+++ b/specs/common/schemas/IndexSettings.yml
@@ -333,25 +333,7 @@ indexSettingsAsSearchParams:
   additionalProperties: false
   properties:
     attributesToRetrieve:
-      type: array
-      items:
-        type: string
-      example:
-        - author
-        - title
-        - content
-      description: |
-        Attributes to include in the API response.
-
-        To reduce the size of your response, you can retrieve only some of the attributes.
-        Attribute names are case-sensitive.
-
-        - `*` retrieves all attributes, except attributes included in the `customRanking` and `unretrievableAttributes` settings.
-        - To retrieve all attributes except a specific one, prefix the attribute with a dash and combine it with the `*`: `["*", "-ATTRIBUTE"]`.
-        - The `objectID` attribute is always included.
-      default: ['*']
-      x-categories:
-        - Attributes
+      $ref: '#/attributesToRetrieve'
     ranking:
       type: array
       items:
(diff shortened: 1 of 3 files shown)


specs/composition-full/common/params/Search.yml
```diff
@@ -1,77 +1,23 @@
 advancedSyntax:
-  type: boolean
-  description: |
-    Whether to support phrase matching and excluding words from search queries
-    Use the `advancedSyntaxFeatures` parameter to control which feature is supported.
-  default: false
-  x-categories:
-    - Query strategy
+  $ref: '../../../common/schemas/IndexSettings.yml#/advancedSyntax'
```
shortcuts, line 2: might not be as clean but reduce the work overall, you can import a property even if it's in a schema, e.g. `common/schemas/IndexSettings.yml#/indexSettingsAsSearchParams/properties/attributesToRetrieve`

shortcuts: I don't mind personally it's up to how much work you want to dedicate in this, exporting them is cleaner indeed

ClaraMuller: I will continue that way then, that won't be lost 👍

specs/common/schemas/IndexSettings.yml
```diff
@@ -779,6 +744,26 @@ userData:
...
     - Advanced
 
+queryLanguages:
+  type: array
+  items:
+    $ref: '#/supportedLanguage'
+  example:
+    - es
+  description: |
+    Languages for language-specific query processing steps such as plurals, stop-word removal, and word-detection dictionarie
```
shortcuts, line 754: some copy/paste seems to miss the last char, e.g. the s here

specs/common/schemas/IndexSettings.yml
```diff
@@ -779,6 +744,26 @@ userData:
...
+  type: array
+  items:
+    $ref: '#/supportedLanguage'
+  example:
+    - es
+  description: |
+    Languages for language-specific query processing steps such as plurals, stop-word removal, and word-detection dictionarie 
+    This setting sets a default list of languages used by the `removeStopWords` and `ignorePlurals` settings.
+    This setting also sets a dictionary for word detection in the logogram-based [CJK](https://www.algolia.com/doc/guides/managing-results/optimize-search-results/handling-natural-languages-nlp/in-depth/normalization/#normalization-for-logogram-based-languages-cjk) languages.
+    To support this, you must place the CJK language **first*
```
shortcuts, line 757: here the *

specs/common/schemas/IndexSettings.yml
```diff
@@ -563,181 +414,260 @@ indexSettingsAsSearchParams:
...
-    minProximity:
-      type: integer
-      minimum: 1
-      maximum: 7
-      description: |
-        Minimum proximity score for two matching words.
+attributeCriteriaComputedByMinProximity:
+  type: boolean
+  description: |
+    Whether the best matching attribute should be determined by minimum proximit
```
shortcuts, line 526:
```suggestion
    Whether the best matching attribute should be determined by minimum proximity
```

specs/common/schemas/IndexSettings.yml
```diff
@@ -563,181 +414,260 @@ indexSettingsAsSearchParams:
...
+    - Query strategy
 
-        - `exactPhrase`.
-          Phrases in quotes must match exactly.
-          For example, `sparkly blue "iPhone case"` only returns records with the exact string "iPhone case".
+advancedSyntaxFeaturesEnum:
+  type: string
+  enum: [exactPhrase, excludeWords]
+  x-categories:
+    - Query strategy
```
shortcuts, line 485: All of those `Enum` suffix should be renamed to their original name (without the Enum), otherwise it changes the model generated and the import, which is a breaking change
Screenshot 2025-07-23 at 12 38 42

ClaraMuller: `yarn cli generate javascript` is working locally

gavinwade12:
@Fluf22 @shortcuts 👋 Clara handed this PR off to me before she left for PTO. I pushed a commit to fix the issue with her last commit, and the CI is green again. Could you please have another look? It seems like there was an unresolved comment: https://github.com/algolia/api-clients-automation/pull/5125#discussion_r2225121318

PR #5126, opened 2025-07-17
feat(specs): add lastUpdatedAt to personalization real-time user

raed667:
## 🧭 What and Why

Add `lastUpdatedAt` property to Advanced-Personalization Real-time user response.

### DIFF
--- a/specs/advanced-personalization/common/schemas/RealtimeUser.yml
+++ b/specs/advanced-personalization/common/schemas/RealtimeUser.yml
@@ -5,6 +5,7 @@ RealtimeUser:
   required:
     - version
     - userID
+    - lastUpdatedAt
     - search
   properties:
     version:
@@ -13,6 +14,8 @@ RealtimeUser:
     userID:
       type: string
       description: User ID of the user.
+    lastUpdatedAt:
+      type: string
     search:
       type: object
       description: |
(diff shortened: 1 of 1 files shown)


specs/advanced-personalization/common/schemas/RealtimeUser.yml
```diff
@@ -13,6 +14,8 @@ RealtimeUser:
     userID:
       type: string
       description: User ID of the user.
+    lastUpdatedAt:
+      type: string
```
shortcuts, line 18: usually we add the format it's useful for languages that need to parse it .e.g Golang

shortcuts (approved):
noice

specs/advanced-personalization/common/schemas/RealtimeUser.yml
raed667: 👍  addressed in 6e7f754401e97bda6419a33d90131c44217f9945

shortcuts: I think we purposely not use date-time because it forces the user to construct objects (e.g. new Date()), in some languages it's a lot of overhead

bengreenbank (approved):

PR #5130, opened 2025-07-18
fix(specs): Add customSearchParameters to Variant schema, and add direction query param for GET abtests

leonardogavaudan:
Two changes:

- added query parameter `direction` for `GET /abtests`
- added `customSearchParameters` in the `Variant` schema which is then used in the `GET /abtests/` response

## 🧭 What and Why

🎟 JIRA Ticket:

### Changes included:

- List changes


leonardogavaudan:
Hey @shortcuts, would it be possible re-release the alpha client after this gets merged? Thanks in advance for your help / review!

shortcuts (approved):
hey sorry I was already OOO at that time, I'd recommend you to set your team as codeowners here https://github.com/algolia/api-clients-automation/blob/main/CODEOWNERS so that you can go ahead with the specs change and potentially release the clients using this GitHub action https://github.com/algolia/api-clients-automation/actions/workflows/scheduled-release.yml

cc @cdhawke @leonardogavaudan

PR #5153, opened 2025-07-22
fix(clients): reduce chances of Push rate limiting

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-4062

### Changes included:

the current implementation let no time for the rate limiting rules to breathe, since we are only waiting for the tasks after every batch have been sent to the api, but for some customer that send e.g. 1_000_000 records, they'd hit that 500 RPM limit quite quickly, before even finishing sending the records to process.

this pr updates the helper code of every clients in order to wait every `batchSize/10` API calls, this smoothes the load a bit and potentially let the rule reset, we could even make that `10` configurable to give more control over potentially rate limited customers.


morganleroi (approved):
Huuuuge!

PR #5223, opened 2025-08-04
chore(scripts): allow retry of the release

millotp:
## 🧭 What and Why

Make the release non-blocking, it will only create the github release for the language that succeeded to release, and send a slack message for the one that failed.
We can retry the CI since it will only try to release languages that have changes.


shortcuts (approved):
gg

PR #5224, opened 2025-08-06
feat(clients): add collections

dhayab:
## 🧭 What and Why

This PR creates a client and describes endpoints for the Collections API.

🎟 JIRA Ticket: [FX-3433]

### Changes included:

- new Collections OpenAPI specs
- Standalone JavaScript Collections API client

## 🧪 Test

- added requests tests
- added JavaScript playground

[FX-3433]: https://algolia.atlassian.net/browse/FX-3433?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/config/clients.config.json
+++ b/config/clients.config.json
@@ -388,7 +388,6 @@
       "abtesting",
       "abtesting-v3",
       "analytics",
-      "collections",
       "composition",
       "ingestion",
       "insights",


specs/collections/spec.yml
```diff
@@ -0,0 +1,102 @@
...
+
+    Use Algolia's API clients and libraries to reliably integrate Algolia's APIs with your apps.
+
+    See: [Algolia's ecosystem](https://www.algolia.com/doc/guides/getting-started/how-algolia-works/in-depth/ecosystem/)
+
+    ## Base URLs
+
+    The base URL for requests to the Collections API is:
+
+    - `https://experiences.algolia.com`
```
Haroenv, line 17: maybe 
```suggestion
    - `https://experiences.algolia.com/1/collections`
```

as in theory we use this endpoint for other things too

specs/collections/spec.yml
```diff
@@ -0,0 +1,102 @@
...
+  # ######################
+  # ### Custom request ###
+  # ######################
+  /{path}:
+    $ref: '../common/paths/customRequest.yml'
+
+  # #############################
+  # ### Collections Endpoints ###
+  # #############################
+  /1/collections:
```
Haroenv, line 90: same question here

specs/collections/spec.yml
```diff
@@ -0,0 +1,102 @@
...
+    $ref: 'paths/collections.yml'
```
Haroenv, line 91: is the yaml always based on the path? If not I think we should call this one list, next one upsert, third one commit makes sense

specs/collections/paths/collectionID.yml
```diff
@@ -0,0 +1,45 @@
...
+      $ref: '../common/responses/CollectionNotFound.yml'
+    '200':
+      description: The requested collection.
+      content:
+        application/json:
+          schema:
+            title: getCollectionResponse
+            $ref: '../common/schemas/collection.yml#/Collection'
+
+delete:
```
Haroenv, line 20: maybe unrelated, but really this is soft-deleting. I wonder if we could build in a "active" status in the collection separate from whether it's existing or not, and a separate endpoint that really deletes the collection. Maybe doesn't matter as we can use a different endpoint for that

specs/collections/common/schemas/condition.yml
```diff
@@ -0,0 +1,8 @@
+Conditions:
+  type: object
+  description: Conditions to filter records.
+  properties:
+    facetFilters:
+      $ref: '../../../common/schemas/SearchParams.yml#/facetFilters'
+    numericFilters:
+      $ref: '../../../common/schemas/SearchParams.yml#/numericFilters'
```
Haroenv, line 8: do we actually support everything? I think the no-code/go transformation didn't implement `TO` in numeric right?

Haroenv (commented):
v cool

PR #5241, opened 2025-08-12
feat(specs): add fields for metadata in composition injectedItems

gavinwade12:
## 🧭 What and Why

🎟 JIRA Ticket: [CMP-485](https://algolia.atlassian.net/browse/CMP-485)

Add support for the new `metadata` field on `injectedItems` and the corresponding `_extra` field on each `hit`.

### Changes included:

- Add the `metadata` property to `injectedItems`
  - It currently only has a `hits` property. You can see the full details [here](https://algolia.atlassian.net/wiki/spaces/CMP/pages/5149753345/Models#Key-definition).
- Add the `_extra` property to `hit`
  - This contains the metadata specified in `injectedItems` + an optional `_injectedItemKey` property

## 🧪 Test

- Green CI

[CMP-485]: https://algolia.atlassian.net/browse/CMP-485?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/specs/composition-full/common/schemas/responses/Hit.yml
+++ b/specs/composition-full/common/schemas/responses/Hit.yml
@@ -19,6 +19,17 @@ hit:
       $ref: '#/rankingInfo'
     _distinctSeqID:
       $ref: '../../../../common/schemas/Hit.yml#/distinctSeqID'
+    _extra:
+      $ref: '#/hitMetadata'
+
+hitMetadata:
+  type: object
+  description: An object that contains the extra key-value pairs provided in the injectedItem definition.
+  additionalProperties: true
+  properties:
+    _injectedItemKey:
+      type: string
+      description: The key of the injectedItem that inserted this metadata.
 
 rankingInfo:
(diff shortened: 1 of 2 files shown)


specs/composition-full/common/schemas/responses/Hit.yml
```diff
@@ -19,6 +19,17 @@ hit:
...
       $ref: '../../../../common/schemas/Hit.yml#/distinctSeqID'
+    _extra:
+      $ref: '#/hitMetadata'
+
+hitMetadata:
+  type: object
+  description: An object that contains the extra key-value pairs provided in the injectedItem definition.
+  additionalProperties: true
+  properties:
+    _injectedItemKey:
```
shortcuts, line 30: is the _ to indicate some private fields? could be worth adding it to the description if so

PR #5277, opened 2025-08-26
refactor(specs): add interface between Search and Composition

ClaraMuller:
## 🧭 What and Why

**What**

This PR refactors the Composition API client specs to introduce a dedicated `Search.yml` file that aggregates all Search API resource references.
Composition API should continue to expose resources from other clients as we plan to grow in API you can directly call. Having a clear separation of concern will make the evolution of the Composition API client easier.

**Why**
  - Avoids duplicating direct `$ref` calls to Search API schemas in multiple files (all parameter that are shared between run and injection: `filters`, `hitsPerPage`, etc...).
  - Simplifies maintenance: future updates to Search API specs can be managed in one place.
  - Makes the dependency between Composition and Search explicit and easier to track.

### Changes included:

- Created a new file `specs/composition-full/common/params/Search.yml` containing all Search API–related `$ref` definitions used in Composition.
- Updated `Injection.yml` and `RunParams.yml` to reference this new file instead of directly pointing to `common/schemas/IndexSettings.yml`, `common/schemas/SearchParams.yml` (the difference means nothing for Composition API) or complexe one like `common/schemas/SearchParams.yml#/baseSearchParamsWithoutQuery/properties/facets`
- This change is a refactor only, no functional behavior is altered.

## 🧪 Test

* CI
* Locally run
    * `yarn cli build specs all`


specs/composition-full/common/schemas/requestBodies/RunParams.yml
```diff
@@ -4,54 +4,54 @@ params:
...
     queryLanguages:
-      $ref: '../../../../common/schemas/IndexSettings.yml#/queryLanguages'
+      $ref: '../../params/Search.yml#/queryLanguages'
     naturalLanguages:
-      $ref: '../../../../common/schemas/SearchParams.yml#/naturalLanguages'
+      $ref: '../../params/Search.yml#/naturalLanguages'
     enableRules:
       $ref: '../../params/Composition.yml#/enableRules'
     ruleContexts:
       $ref: '../../params/Composition.yml#/ruleContexts'
```
shortcuts, line 45: should those, that are initially search parameters, but edited to match composition's requirement, also moved to that new Search.yml?

shortcuts (approved):
nitpick comment, looks good

specs/composition-full/common/schemas/requestBodies/RunParams.yml
ClaraMuller: As they are not related anymore to Search API client, I don't believe so.
The Search.yml file also contains the ref to the actual search query paramter that we used in other part of the composition API clients.
For these query paramater. available for run request for Composition, we want them to point to Composition specific parameters

PR #5284, opened 2025-08-29
fix(python): remove warnings for HighlightResult and SnippetResult

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/CR-8953 https://algolia.atlassian.net/browse/DI-4020

### Changes included:

closes https://github.com/algolia/algoliasearch-client-python/issues/569

When calling `to_dict` on a search response, pydantic yells about recursion issue and incorrect typings of our HighlightResult and SnippetResults models.

image

I tried many ways to solve this, and sadly the only other solution that would work (except the one proposed in this PR) is to implement a custom serializer for every models, which would be a much harder work to put in place.

In the meantime, I propose using a custom serializer (with the `field_serializer` feature from Pydantic), that allows us to simply remove warnings for those two models.

### DIFF
--- a/playground/python/app/search.py
+++ b/playground/python/app/search.py
@@ -11,48 +11,47 @@ load_dotenv("../.env")
 def main():
     print("SearchClient version", __version__)
 
-    # client = SearchClientSync(
-    #     environ.get("ALGOLIA_APPLICATION_ID"), environ.get("ALGOLIA_ADMIN_KEY")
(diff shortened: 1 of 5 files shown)


millotp (approved):
nice fix, we should never have infinite recursion, unless someone malicious forges the object

PR #5339, opened 2025-09-11
fix(specs): apply the right category tags to composition query parameter

ClaraMuller:
## 🧭 What and Why

Fix some category tags for composition API run query parameters.

🎟 JIRA Ticket: contributes to https://algolia.atlassian.net/browse/CMP-487

## 🧪 Test

* CI
* Locally run
    * `yarn cli build specs all`


Fluf22 (approved):
🆗

PR #5364, opened 2025-09-17
feat(specs): deprecate Predict real-time endpoints

raed667:
## 🧭 What and Why

Deprecate the `compute real-time user` endpoint.

The Real-time Personalization feature now does this behind the scenes automatically.

I also updated the description of the `getRealtimeUser` endpoint to make it clearer to differentiate with the "regular" get-profile endpoint


shortcuts (approved):
nice!

PR #5365, opened 2025-09-17
chore(scripts): add an option to attach a java debugger

millotp:
## 🧭 What and Why

Thanks to @Fluf22 for trying this out, we can attach a debugger to the generator !
This works with IntelliJ, VSCode, and the `jdb` cli.

For example, run `apic cts generate csharp search` and then attach the debugger, the java process will wait until it's attached.


shortcuts (approved):
awesome

shortcuts (approved):
gg

PR #5368, opened 2025-09-19
snippets: remove snippets for non-existing helpers

millotp:
## 🧭 What and Why

Noticed by @kai687, we don't validate dart snippets and some helpers that we call in the snippets don't exists.
We can remove the snippets for now, and validate dart files so this doesn't happen again.


shortcuts (approved):
awesome

PR #5443, opened 2025-10-06
fix(clients): move default timeouts to the constructor

millotp:
## 🧭 What and Why

Noticed by @kombucha, the default timeouts are not set if the client is created using `new(config)` instead of `create_with_config(config)`, we can move them there.


morganleroi (approved):
🆗

PR #5448, opened 2025-10-08
chore(scripts): dont use verbose when releasing from the CI

millotp:
## 🧭 What and Why

The release is not working from the CI because when we use `setVerbose(true)`, the behavior of the `run` function is different and might return unexpected results.

Removing `setVerbose(CI)` might solve the issues.


morganleroi (approved):
What a fix!

PR #5477, opened 2025-10-14
feat(specs): add specific documentation for `facets` query parameter in Composition API

ClaraMuller:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/CMP-594

### Changes included:

* Enrich the documentation for `facets` query parameter with features that are only implemented within Composition API.
For that, remove the link toward Search API doc and create a `facets` section within Composition API query parameter.
* Add this parameter as a potential value for `run` endpont as it is allowed in the payload

## 🧪 Test

* CI
* Run locally `yarn cli build specs all && yarn cli build clients javascript`

### DIFF
--- a/specs/composition/common/params/Composition.yml
+++ b/specs/composition/common/params/Composition.yml
@@ -1,3 +1,23 @@
+# #########################
+# ### Category Faceting ###
+# #########################
+
+facets:
+  type: array
+  items:
+    type: string
+  description: |
+    Facets for which to retrieve facet values that match the search criteria and the number of matching facet values
+    To retrieve all facets, use the wildcard character `*`.
+    To retrieve disjunctive facets lists, annotate any facets with the `disjunctive` modifier.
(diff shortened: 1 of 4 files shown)


specs/composition/common/params/Composition.yml
```diff
@@ -1,3 +1,23 @@
+# #########################
+# ### Category Faceting ###
+# #########################
+
+facets:
```
millotp: `facets` is already defined in the `IndexSettings.yml` file, and since it's imported in the composition spec, it will clash with this one and produce unpredictable naming, like `IndexSettingsFacets`.

It's not possible to override a model like that, either you would have to copy the entire `IndexSettings.yml` in here and edit the `facets`, or you need to rename this to something else like `compositionFacets`, but the `facets` object will still be generated with the other spec since it comes first.

PR #5479, opened 2025-10-14
snippets: remove all the >LOG

millotp:
## 🧭 What and Why

The `>LOG` marker was intended to be used internally only, but it leaked into the doc and it's not very nice to look at.

Screenshot 2025-10-15 at 00 01 28

Also harmonize the log for all languages.


scripts/specs/__tests__/snippets.test.ts
millotp: this test was broken and not testing anything

shortcuts (dismissed):
perfect

shortcuts (approved):
gg!

PR #5516, opened 2025-10-20
fix(java): dont use watch=true in replaceAllObjectsWithTransformation

millotp:
## 🧭 What and Why

The `replaceAllObjectsWithTransformation` helper already uses `getEvent`, the `watch` parameter in push should be false, otherwise it waits twice.


shortcuts (approved):
oopsie

PR #5529, opened 2025-10-24
fix(specs): add ACL to missing endpoints

millotp:
## 🧭 What and Why

🎟 JIRA Ticket: [DI-4359](https://algolia.atlassian.net/browse/DI-4359)

Add all missing ACLs, and a linter to assert that all endpoints have the `x-acl` prop.

[DI-4359]: https://algolia.atlassian.net/browse/DI-4359?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


eslint/src/rules/validACL.ts
```diff
@@ -43,6 +44,36 @@ export const validACL = createRule('validACL', {
 
     return {
       YAMLPair(node): void {
+        const spec = context.getFilename().match(/specs\/([a-z-]+?)\//)?.[1];
+        if (!spec) {
+          return;
+        }
+        if (spec === 'monitoring') {
```
shortcuts: should this be the case for the crawler as well? or is it the case because it' doesn't have a client?

shortcuts (approved):

PR #5620, opened 2025-11-12
chore(csharp): test the support of multiple dotnet versions

millotp:
## 🧭 What and Why

Since switching to .NET 10 in #5585, we need to make sure that we still support older versions.


Fluf22 (approved):
Very nice to see all these languages getting version matrices

PR #5621, opened 2025-11-12
chore(go): test the support of multiple go versions

millotp:
## 🧭 What and Why

It's very easy to break the go client if we upgrade `.go-version`, we need to make sure we always support at least 1.21.11


shortcuts (approved):
this is incredibly huge

PR #5645, opened 2025-11-17
fix(specs): use summary fields

kai687:
## 🧭 What and Why

OpenAPI 3.1 allows for a `summary` field in the `info` object.
This makes it easier to render documentation from the specs, since I can use the `summary` directly.

🎟 JIRA Ticket:

### Changes included:

- Moves the first sentences from `openapi.info.description` to `openapi.info.summary`.


kai687:
So, this page seems to be wrong: https://api-clients-automation.netlify.app/docs/add-a-new-api/write-a-specification

The spec validation is surprised by the `openapi.info.summary` field which is valid in OpenAPI 3.1.

PR #5650, opened 2025-11-18
fix(specs): fix example in Composition API specs

ClaraMuller:
## 🧭 What and Why

I saw that the final documentation of `facets` query parameter in Comp API in the documentation is not valid because we made an error in the specs:

Screenshot 2025-11-18 at 10 13 29

_(the value in example is not valid)_ ⬆️ 

This PR aims to fix the examples

### Changes included:

* Update of specs for Composition API

## 🧪 Test

* CI
* Run locally `yarn cli build specs all && yarn cli build clients javascript`


Fluf22 (dismissed):
Ah yes apparently it must be an object, not a list
Let's see if it works this way

ClaraMuller:
@Fluf22 unclear what is expected, documentation (https://swagger.io/docs/specification/v3_0/adding-examples/) asks for an object, but the lint specs failed because it expects an iterator it seems ?

PR #5673, opened 2025-11-24
fix(go): avoid body copy

millotp:
## 🧭 What and Why

Small optimization that avoids doing a full body read before each request in go.
Instead, it uses `io.TeeReader` that will do a copy of the body while it's being written to the http client, and save it for the next call in the retry chain.


Fluf22 (approved):
Let's go

PR #5676, opened 2025-11-25
chore(specs): change oneOf order for docs

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: -

### Changes included:

so it displays the object shape first on the doc instead of the search string alternative, you can see https://www.algolia.com/doc/rest-api/search/browse that:
1. it recommends the string by default
2. option2 is the object because of the missing title


millotp (approved):
good good

PR #6098, opened 2026-03-12
fix(templates): handle null parameters and nullable fields in test generation

Fluf22:
Jira: API-454
## Summary
Extracted from #5860 — generic template fixes for null parameter handling across Go, Kotlin, Scala, and Swift.

### Changes
- **Go**: Use nameInPascalCase for optional With* methods; skip null optional params in tests; reformatted long calls
- **Kotlin**: Handle null params in test request generation; handle nullable required fields (? = null)
- **Scala**: Handle null params in test request generation
- **Swift**: Emit nil for null values outside client tests


Fluf22:
@MarioAlexandruDan the goal here is to be able to send a literal `null` value to the API, compared to just not sending the field, which is not the same behavior

PR #6100, opened 2026-03-12
fix(clients): deserialization and HTTP header handling

Fluf22:
## Summary
Extracted from #5860 — generic client fixes for Python and Scala.

### Changes
- **Python** (api_response.py): Handle non-string input for list types; add dict type deserialization
- **Scala** (HttpRequest.scala): Add withHeader(key, Option[Any]) overload that skips None values


eric-zaharia (approved):
Looks good

PR #3567, opened 2024-08-20
fix(clients): highlight and snippet results e2e

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-2767 https://algolia.atlassian.net/browse/DI-2776 https://algolia.atlassian.net/browse/DI-2779

### Changes included:

reported in https://github.com/algolia/algoliasearch-client-kotlin/issues/423, should close https://github.com/algolia/algoliasearch-client-kotlin/issues/423

also enables e2e for kotlin, can't remember why it was missing but... it was, and highlights some problems

PR #6199, opened 2026-04-02
fix(generators): auto-infer discriminator fields from unique required properties

Fluf22:
## Summary

- For oneOf variants without explicit `x-discriminator-fields`, the generator now auto-infers discriminators from required fields that are unique to each variant
- Explicit `x-discriminator-fields` in the spec take precedence (no behavioral change for existing annotations)
- Covers ~86% of existing annotated schemas automatically, eliminating the need for manual spec annotations in most cases

## Context

Go's `json.Unmarshal` is too lenient for oneOf disambiguation — it doesn't fail on missing/extra fields, so multiple variants can "succeed" on the same payload. The existing `x-discriminator-fields` mechanism solves this by adding `hasKey()` guards before unmarshal attempts, but requires manual annotation per schema.

This change makes the generator compute those guards automatically from the `required` fields that are unique to each variant (e.g., `SearchSource` requires `search`, `ExternalSource` requires `external` — no overlap → auto-injected as discriminators).

The 14% of schemas where discriminators are based on non-required fields still need explicit `x-discriminator-fields`.

## Related

- #6170 — manual fix for `InjectedItemSource` (would be covered automatically by this change)
- #5854 — Go oneOf unmarshal strategy change
- #3567 — original `x-discriminator-fields` mechanism


eric-zaharia (approved):
Looks good!

PR #6224, opened 2026-04-09
refactor(specs): split injection schema

ClaraMuller:
## 🧭 What and Why

🎟 Contributes to https://algolia.atlassian.net/browse/CMP-761

### Stacked PR:

* https://github.com/algolia/api-clients-automation/pull/6224 <- This PR
* https://github.com/algolia/api-clients-automation/pull/6227

#### → File reorg

- **Split `Behaviour.yml`** into dedicated schema files:
    - **`InjectedItem.yml`** (new) for `injectionInjectedItem` schema
    - **`Main.yml`** (new) for `injectionMain` and `injectionMainSource` schemas
    - **`MainSource.yml`** (new) for `SearchSource` and `RecommendSource` (placeholder) for the main injection source `oneOf`
  - **Renamed schema files** for clarity:
    - `Source.yml` → `InjectedItemSource.yml`
    - `QueryParams.yml` → `SearchQueryParams.yml`

#### → Strongly typed client impact generation

  - **Separated main and injected item source types**: `main.source and injectedItems[].source` now resolve to distinct `oneO`f types (`injectionMainSource` vs `injectedItemSource`), producing separate strongly-typed models
  in generated clients
  - **Added `RecommendSource` placeholder** to `injectionMainSource` `oneOf` to ensures the generator creates a proper `oneOf` wrapper type instead of inlining the single variant

#### → A few comments

Because all models are generated at the root level in the Go client (no sub-packages), two constraints arise:
- **Every `title` must be unique across the entire package.** All generated struct names must be globally unique. In the context of the Composition API, it would have been more natural to use package separation allowing the same struct name (e.g., `SearchSource`) in different packages (`main` vs `injectedItem`). This would better reflect the intent of the API, where both sources share the same shape but serve different roles.
- **Unique names lead to a significant increase in verbosity.** Each struct must encode its full context in its name to avoid collisions (e.g., `InjectedItemSourceSearchSource` instead of just `SearchSource`), making the generated code harder to read.
- Key `title` does not behave the same way according to the language generated. From what I saw, it is for example completely ignored by the Go language for direct specs objects and at the same time require for inner object. According to what I read, in C# this handled for each value of `title` 🤷‍♀️ 

Additionally, since feature teams only modify the specs, **they cannot always anticipate that a future feature will require an object with a very similar shape.** This is forcing retroactive renames to maintain uniqueness, and therefore **producing breaking changes**. For example, the previous titles `compositionSourceSearch` or `compositionSource` no longer make sense now that both `main` and `injectedItems` have their own source types.

_I am speaking for the Go client but because this one I understand the most but I am assuming this is the same thing for other strongly typed clients_

## 🧪 Test

* Run locally `yarn cli build specs all && yarn cli generate go composition`
* CI


specs/composition/common/schemas/components/injection/Main.yml
```diff
@@ -0,0 +1,14 @@
+injectionMain:
+  type: object
+  title: main
```
gavinwade12: Why set the `title` instead of naming the component "main"? IIUC the `title` is more intended for objects being defined inline inside another object

```suggestion
main:
  type: object
```

specs/composition/common/schemas/components/injection/MainSource.yml
```diff
@@ -0,0 +1,43 @@
...
+  additionalProperties: false
+  properties:
+    search:
+      title: mainSearchSource
+      type: object
+      additionalProperties: false
+      properties:
+        index:
+          type: string
+          description: Composition Index name.
```
gavinwade12: Do you think having "Composition" here could be confusing?

```suggestion
          description: Targeted index name.
```

specs/composition/common/schemas/components/injection/MainSource.yml
```diff
@@ -0,0 +1,43 @@
...
+  additionalProperties: false
+  properties:
+    recommend:
+      title: mainRecommendSource
+      type: object
+      additionalProperties: false
+      properties:
+        index:
+          type: string
+          description: Composition Index name.
```
gavinwade12: Same

specs/composition/common/schemas/components/injection/Behaviour.yml
```diff
@@ -4,108 +4,23 @@ injection:
...
-            - search
-      required:
-        - source
+      $ref: './Main.yml#/injectionMain'
     injectedItems:
       type: array
       description: list of injected items of the current Composition.
       minItems: 0
-      maxItems: 2
+      maxItems: 3
```
gavinwade12, line 12: Do you know if we have anyone that was manually increased to > 3? Should we just put this at the true max we'll allow?

specs/composition/common/schemas/components/injection/InjectedItem.yml
```diff
@@ -0,0 +1,60 @@
+injectionInjectedItem:
+  type: object
+  title: injectedItem
```
gavinwade12: Same

```suggestion
injectedItem:
  type: object
```

tests/output/go/tests/manual/composition_oneof_test.go
```diff
@@ -19,9 +19,9 @@ func TestInjectedItemSource_UnmarshalSearchSource(t *testing.T) {
 	err := json.Unmarshal([]byte(input), &source)
 	require.NoError(t, err)
 
-	require.NotNil(t, source.SearchSource, "expected SearchSource to be set")
+	require.NotNil(t, source.InjectedItemSourceSearchSource, "expected InjectedItemSourceSearchSource to be set")
```
gavinwade12: Should we keep the actual source definitions in `InjectedItem.yml` so it reads `InjectedItemSearchSource` instead of `InjectedItemSourceSearchSource`?

specs/composition/common/schemas/components/injection/Behaviour.yml
ClaraMuller: This is actually a fix, the default value is 3 (see our [doc](https://algolia.atlassian.net/wiki/spaces/CMP/pages/5656150035/Composition+API+-+Application+Configuration))

specs/composition/common/schemas/components/injection/InjectedItem.yml
ClaraMuller: I admit I wanted to give context to the object.
I am not sure about how exactly it is ending into the global `specs` file, but the last time we struggled so much with common parameter with Search API that I did not want to take any risks.

But maybe for this struct, we can keep the initial name 🤔

specs/composition/common/schemas/components/injection/Main.yml
ClaraMuller: No because actually, if title is not set, the structure generated is the name of the object.
Imo `main` is too generic to be included in specs directly

specs/composition/common/schemas/components/injection/MainSource.yml
ClaraMuller: In the context of multifeeds, this is about the index targeted of the feed, not necessarily "the index tied to the composition".

tests/output/go/tests/manual/composition_oneof_test.go
ClaraMuller: I am not sure the split of the file is leading to the shift of the name 🤔 .
But I can see if I manage to turn it into `InjectedItemSearchSource`

ben-kalmus:
> Separated main and injected item source types: main.source and injectedItems[].source now resolve to distinct oneOf types (injectionMainSource vs injectedItemSource), producing separate strongly-typed models
> in generated clients

Does this solve the problem of api clients deprecating previous versions as soon as new enumerable is added? Is the only downside of this that we have to maintain both versions of sources: in main and injectedItem ?

ClaraMuller:
@ben-kalmus 
> _Does this solve the problem of api clients deprecating previous versions as soon as new enumerable is added?_

Yes because in the next iterations, the `oneOf` will be properly handled.
No because this introduces a breaking change in this version

> _Is the only downside of this that we have to maintain both versions of sources: in main and injectedItem ?_

We were already maintaining 2 different structures. Just the main one was directly put at injection level.
The query parameter restrictions according to main source or injected item source produced from day one 2 different structures.

tests/output/go/tests/manual/composition_oneof_test.go
gavinwade12: I believe it's because you have `SearchSource` defined in more than one spot, so this one gets the file name as a prefix to prevent a clash

specs/composition/common/schemas/components/injection/Main.yml
gavinwade12: Is it too generic to be included in the client then, too? i.e. should it just be `injectionMain` *everywhere*?

specs/composition/common/schemas/components/injection/MainSource.yml
gavinwade12: IMO "Composition Index name" implies the index *is* tied to the composition

specs/composition/common/schemas/components/injection/Behaviour.yml
gavinwade12: 3 is the default maximum, but we can manually increase it up to 10; yet, anyone we do that for won't be able to use the official client because of this maximum. I'm asking if we should put this at 10 to prevent that situation

specs/composition/common/schemas/components/injection/InjectedItem.yml
gavinwade12: Maybe we should just stick with the "injection" prefix for anything related to that behavior so we don't have more breaking changes when we introduce new behaviors that also might include e.g. `main` 🤔

specs/composition/common/schemas/components/injection/Main.yml
ClaraMuller: Maybe 🤷‍♀️

specs/composition/common/schemas/components/injection/MainSource.yml
ClaraMuller: Oh my bad, I misread your comment, I agree, I though I made your changes

specs/composition/common/schemas/components/injection/Behaviour.yml
ClaraMuller: Personally I want to avoid advertise the actual maximal value and just stick to  what service limits document.

specs/composition/common/schemas/components/injection/InjectedItem.yml
ClaraMuller: I would tend to agree

specs/composition/common/schemas/components/injection/InjectedItemSource.yml
```diff
@@ -1,13 +1,13 @@
 SearchSource:
-  title: searchSource
+  title: injectedItemSearchSource
```
gavinwade12: I believe `title` is only handy when defining things inline and we should stick with directly using the name for the component where possible

```suggestion
injectedItemSearchSource:
```

specs/composition/common/schemas/components/injection/MainSource.yml
```diff
@@ -0,0 +1,45 @@
+SearchSource:
+  title: mainSearchSource
```
gavinwade12: Same, plus I noticed we are inconsistent with casing. Sometimes we start with lowercase and sometimes uppercase

```suggestion
mainSearchSource:
```

specs/composition/common/schemas/components/injection/MainSource.yml
```diff
@@ -0,0 +1,45 @@
...
+          example: Products
+        params:
+          $ref: './SearchQueryParams.yml#/mainInjectionQueryParameters'
+      required:
+        - index
+  required:
+    - search
+
+RecommendSource:
+  title: mainRecommendSource
```
gavinwade12: Same
```suggestion
mainRecommendSource:
```

specs/composition/common/schemas/components/injection/MainSource.yml
```diff
@@ -0,0 +1,45 @@
+SearchSource:
+  title: mainSearchSource
+  description: Organic result set will originate from a search request performed on the specified index.
+  x-discriminator-fields:
+    - search
+  type: object
+  additionalProperties: false
+  properties:
+    search:
+      title: mainSearchSource
```
gavinwade12: Wait, all these inner objects have the same `title` as the containing objects. I'm surprised we don't get some sort of error from this 🤔

tests/output/go/tests/manual/composition_oneof_test.go
ClaraMuller: So I checked, Go generator ignores the `title` key 🙄 Let's make the situation even more complicated

gavinwade12: I think we should try to avoid `title` when possible anyways
https://github.com/algolia/api-clients-automation/pull/6224#discussion_r3086872194

specs/composition/common/schemas/components/injection/MainSource.yml
ClaraMuller: Yes, I am wondering if title is not ignored when not with the object 🤔

specs/composition/common/schemas/components/injection/InjectedItem.yml
```diff
@@ -0,0 +1,59 @@
...
+    - source
+    - position
+    - length
+
+injectedItemSource:
+  oneOf:
+    - $ref: './InjectedItemSource.yml#/injectedItemSearchSource'
+    - $ref: './InjectedItemSource.yml#/injectedItemExternalSource'
+
+dedupPositioning:
```
gavinwade12, line 47: It's a little annoying/verbose but could save us from more breaking changes in the future. WDYT?
```suggestion
injectionDedupPositioning:
```

specs/composition/common/schemas/components/injection/InjectedItem.yml
```diff
@@ -0,0 +1,59 @@
...
+              type: object
+              additionalProperties: true
+              description: The user-defined key-value pairs that will be placed in the `_extra` field of each affected hit.
+  required:
+    - key
+    - source
+    - position
+    - length
+
+injectedItemSource:
```
gavinwade12, line 42: Same, but even less concerned about this one
```suggestion
injectionInjectedItemSource:
```

specs/composition/common/schemas/components/injection/InjectedItemSource.yml
```diff
@@ -1,13 +1,12 @@
-SearchSource:
-  title: searchSource
+injectedItemSearchSource:
```
gavinwade12, line 1: Same could apply to most types in this file, but I think the chance of a clash/breaking change in the future is quite low (but not impossible). The cost of safety is an annoying level of verbosity 😓 
```suggestion
injectionInjectedItemSearchSource:
```

specs/composition/common/schemas/components/injection/Behaviour.yml
```diff
@@ -4,108 +4,23 @@ injection:
...
       items:
-        $ref: '#/injectedItem'
+        $ref: './InjectedItem.yml#/injectionInjectedItem'
     deduplication:
       title: deduplication
```
gavinwade12, line 16:
```suggestion
      title: injectionDeduplication
```

gavinwade12 (approved):
I think we'd be fine as-is if we want, so none of my comments are blocking. Like I mentioned in one of the comments, I think we could choose to be even safer via naming, but at the cost of an annoying level of verbosity

ClaraMuller:
@gavinwade12 I hesitated with the structure you mention in comment tbh.
I though we can start like this

Fluf22 (approved):
We need to find a way to avoid breaking changes when feature teams add more objects somewhere

PR #6225, opened 2026-04-09
docs(insights): Update the user token deletion summary

fophillips:
## 🧭 What and Why

Update the summary of the user token deletion API to make it clear we delete events, not user tokens

corresponding docs change: https://github.com/algolia/docs-new/pull/714

### Changes included:

- user token deletion summary

## 🧪 Test
N/A


Fluf22 (approved):
✅

PR #6247, opened 2026-04-13
fix(clients): push helper smooth traffic

shortcuts:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/DI-4807

### Changes included:

a push request takes in average 4.1s for completion, so we can increase the retry timeout logic to avoid spamming the API


Fluf22 (approved):
Great teamwork

PR #6255, opened 2026-04-15
feat(python): add TransformationOptions object

eric-zaharia:
## 🧭 What and Why

🎟 JIRA Ticket: [API-373]

### Summary:
Adds `TransformationOptions` to the Python search client. The old `config.set_transformation_region()` method had two problems: it leaked the parent search config (hosts, timeouts) into the ingestion transporter, and it didn't work after client construction. Rather than fixing it, this PR deprecates it and introduces `TransformationOptions`, which is passed via `SearchConfig` at construction or via `client.set_transformation_options()` after construction. The ingestion transporter is created using ingestion API defaults, only overriding what's explicitly specified.

This PR also fixes the ingestion transporter's HTTP session not being closed when the search client is closed.

### Changes included:
- Add `TransformationOptions` class with `region` (required) + optional ingestion config fields (timeouts, hosts, proxies, headers, compression settings)
- Add `transformation_options` to `SearchConfig` to eagerly create the ingestion transporter at client construction using ingestion defaults — the parent search config no longer leaks into the ingestion transporter
- Add `client.set_transformation_options()` to allow updating transformation options after client creation (closes the previous ingestion transporter on replacement)
- Deprecate `config.set_transformation_region()` with a warning explaining it does not affect an already constructed client
- Update error messages in `*_with_transformation` methods to reference both `SearchConfig` and `client.set_transformation_options()`
- Fix `close()` not closing the ingestion transporter — previously, closing the search client (or using `async with`) would leave the ingestion transporter's HTTP session open

## 🧪 Test
- CTS e2e tests for `saveObjectsWithTransformation`, `partialUpdateObjectsWithTransformation`, `replaceAllObjectsWithTransformation`
- Manual unit tests (`tests/output/python/tests/manual/test_transformation_options.py`) covering the `close()` fix and the `set_transformation_options` state-swap for both async and sync clients

[API-373]: https://algolia.atlassian.net/browse/API-373?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clients/algoliasearch-client-python/algoliasearch/http/base_config.py
```diff
@@ -30,8 +30,8 @@ def __init__(self, app_id: Optional[str] = None, api_key: Optional[str] = None):
 
         self._user_agent: UserAgent = UserAgent()
 
-        self.compression_type = None
-        self.compression_threshold = 750
+        self.compression_type: Optional[str] = None
```
Fluf22: Since you're updating this, don't we have an enum for available compression types?

Fluf22 (dismissed):
Small nitpicks; looks good otherwise

clients/algoliasearch-client-python/algoliasearch/http/base_config.py
eric-zaharia: There's no enum in Python for the compression type (values can only be "gzip" or None), we could tighten it a bit more with `Optional[Literal["gzip"]]` which gives the same guarantee as an enum with no new type to introduce.

Fluf22: Would it be centralized here, or would we have to update other occurrences?

eric-zaharia: It would need to be added in the `TransformationOptions` constructor as well but that should be it. Should I apply this modification?

Fluf22: Okay
Well, no strong opinion on my side: I believe the types are helpful to avoid doing too many back and forth between the coding env and the docs (for humans) and to avoid too many wrong guesses (for AI agents)
But I don't want to introduce too much friction either

I'll let you decide what makes most sense here!

eric-zaharia: I'll add it, it'll help with IDE auto-completion and pyright flagging values different to `gzip`

Fluf22 (approved):

PR #6254, opened 2026-04-15
feat(specs): add extensions field for Query Categorization to SearchForHits and SearchResponse

dmytrosarzhan-mono:
---                                                                                                                                                                       
  feat(specs): add extensions field for Query Categorization
                                                                                                                                                                            
  Summary                                                
                                                                                                                                                                            
  Adds first-class spec support for the https://www.algolia.com/doc/guides/algolia-ai/query-categorization/ AI feature, which is currently undocumented in the OpenAPI spec 
  and therefore missing from all generated clients.                                                                                                                         
                                                                                                                                                                            
  ---                                                                                                                                                                       
  Problem
                                                                                                                                                                            
  The Algolia Search API already supports extensions.queryCategorization in both the request and the response, but it is not modelled in the spec. This means:
                                                                                                                                                                            
  - Clients have no typed way to send extensions.queryCategorization.enableCategoriesRetrieval / enableAutoFiltering                                                        
  - The response field extensions.queryCategorization (with category predictions, confidence levels, and auto-filtering metadata) is silently swallowed as an unknown       
  additionalProperty                                                                                                                                                        
  - Users who want to use Query Categorization must resort to raw Map manipulation or subclassing generated models
                                                                                                                                                                            
  ---                                                    
  Changes                                                                                                                                                                   
                                                                                                                                                                            
  specs/search/common/schemas/SearchQuery.yml
                                                                                                                                                                            
  - Added extensions property to searchForHitsOptions referencing a new searchExtensions schema                                                                             
  - New schemas:                                                                                                                                                            
    - searchExtensions — top-level container for AI feature parameters                                                                                                      
    - searchExtensionsQueryCategorization — enableCategoriesRetrieval (bool, default false) and enableAutoFiltering (bool, default false)                                   
                                                                                                                                                                            
  specs/common/schemas/SearchResponse.yml                                                                                                                                   
                                                                                                                                                                            
  - Added extensions property to baseSearchResponse referencing a new responseExtensions schema                                                                             
  - New schemas (all top-level, per the out-of-line enum rule):
    - responseExtensions — container for AI-generated response metadata                                                                                                     
    - queryCategorization — normalizedQuery, type, categories[], autofiltering                                                                                              
    - queryCategorization_type — enum: narrow | broad | ambiguous | none                                                                                                    
    - categoryPrediction — bin, hierarchyPath[]                                                                                                                             
    - categoryPrediction_bin — enum: certain | very high | high | low | very low                                                                                            
[...]

PR #6258, opened 2026-04-17
feat(specs): query categorization

eric-zaharia:
## 🧭 What and Why

Adds extensions.queryCategorization to the search response spec (credit to @dmytrosarzhan-mono, original PR #6254), plus the maintenance tooling needed to keep the CTS e2e test green long-term.

### Changes included:

- Add extensions.queryCategorization to the search response schema (categories, normalizedQuery, count, type, autofiltering)
- Add CTS e2e test covering extensions.queryCategorization on query sofa, asserting only the minimum shape that survives model drift across all 11 languages
- Add TypeScript scripts under `scripts/index-maintenance/query-categorization/` to populate the QC test index and generate click/conversion events
- Add .github/workflows/qc-events-cron.yml: monthly cron that re-seeds events so the 90-day QC window never lapses; Slack failure notification included

## 🧪 Test

- Ran the scripts locally end-to-end against a throwaway index; QC trained within 24h and produced predictions matching the CTS assertion


Fluf22 (approved):
Great 👍🏻

PR #6283, opened 2026-04-21
fix(java): prevent chunkedPush subList index errors

louisfisch:
## 🧭 What and Why

`chunkedPush` processes records one-by-one, but it only gets a new server response when it actually sends a chunk. I think there is a bug in the “wait for responses” logic that can run on loop iterations where no chunk was sent. So it keeps moving its response window (`offset`) even though the response list didn’t grow. After enough iterations, the window becomes invalid and it crashes.

In short: it mixes two counters (records vs pushed responses), and the response cursor can drift out of sync.

### Changes included:

- Polling only runs on iterations where a new push response was actually created (`pushed == true`), which prevents repeated re-processing while iterating records inside the same response window.
- Changing offset advancement from `offset += waitBatchSize` to `offset = nextOffset`, where `nextOffset` is capped to `responses.size()`.
- Changing the name `waitBatchSize` to `pollInterval` to avoid the confusion with `batchSize`

### DIFF
--- a/templates/java/api_helpers.mustache
+++ b/templates/java/api_helpers.mustache
@@ -42,19 +42,27 @@ public <T> List<WatchResponse> chunkedPush(
   String referenceIndexName,
   RequestOptions requestOptions
 ) {
+  if (batchSize < 1) {
+      throw new AlgoliaRuntimeException("`batchSize` must be greater than 0");
+  }
+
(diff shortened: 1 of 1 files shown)


shortcuts (approved):
thanks a lot and good catch!!!!!!!!!

PR #6284, opened 2026-04-22
fix(specs): sort bundled specs for docs

kai687:
## 🧭 What and Why

The docs get a special bundle that we use for rendering the API references.
This PR makes it so that parameters, request body fields, and response body fields appear sorted in the specs: required before optional, then alphabetical.
This increases scannability of our API references and shouldn't have any side effect on the regular specs.

Most of the code just deals with resolving references and flattening composite schemas like `allOf`, to allow sorting their properties.

🎟 JIRA Ticket:

### Changes included:

- List changes


Fluf22 (approved):
👌🏻

PR #6289, opened 2026-04-23
feat(java): add TransformationOptions for ingestion configuration

MarioAlexandruDan:
## 🧭 What and Why

🎟 JIRA Ticket [API-370](https://algolia.atlassian.net/browse/API-370):

### Changes included:

- List changes

## 🧪 Test

[API-370]: https://algolia.atlassian.net/browse/API-370?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clients/algoliasearch-client-java/algoliasearch/src/main/java/com/algolia/config/ClientOptions.java
```diff
@@ -33,6 +33,8 @@ public static Builder builder() {
 
   private final ExecutorService executor;
 
+  private final TransformationOptions transformationOptions;
```
Fluf22: This is specific to the Search client, it shouldn't live in the ClientOptions file, which is generic to all clients

clients/algoliasearch-client-java/algoliasearch/src/main/java/com/algolia/config/TransformationOptions.java
Fluf22, line 1: Can you use composition so the TransformationOptions can accept the region and a generic ClientOptions that you can forward to the IngestionClient
It would be cleaner than redeclaring all fields

templates/java/api.mustache
```diff
@@ -60,10 +60,14 @@ public class {{classname}} extends ApiClient {
...
-     *
-     * @param region (required)
+     * @deprecated Use {@link ClientOptions.Builder#setTransformationOptions(TransformationOptions)}
+     *     instead. The old setter forwarded the parent search {@link ClientOptions} to the ingestion
+     *     transporter, which caused search timeouts to bleed into ingestion calls. The new
+     *     {@code transformationOptions} defaults to Ingestion API defaults (25 s timeouts) and only
+     *     overrides what is explicitly specified.
+     *     See https://www.algolia.com/doc/libraries/sdk/methods/ingestion
      */
+    @Deprecated
```
Fluf22, line 93: Deprecated path still leaks search config; should use ingestion defaults
Check the Python [PR](https://github.com/algolia/api-clients-automation/pull/6255) for reference

templates/java/api.mustache
Fluf22, line 1: Missing close() override to clean up ingestion transporter (check Python PR for reference)

Fluf22: It would be even better with a specific test for it 😁

clients/algoliasearch-client-java/algoliasearch/src/main/java/com/algolia/config/ClientOptions.java
MarioAlexandruDan: Done — removed `transformationOptions` field, getter, and builder setter from `ClientOptions`. It now lives only on `SearchClient` via the new `TransformationOptions` constructor parameter.

clients/algoliasearch-client-java/algoliasearch/src/main/java/com/algolia/config/TransformationOptions.java
MarioAlexandruDan: Done — `TransformationOptions` now uses composition: accepts `region` + an optional `ClientOptions` instead of redeclaring individual fields. The `ClientOptions` is forwarded directly to `IngestionClient`.

templates/java/api.mustache
MarioAlexandruDan: Fixed — `setTransformationRegion` now delegates to the new `setTransformationOptions` method which calls `new IngestionClient(appId, apiKey, region)` without forwarding search `ClientOptions`, so ingestion defaults (25s) apply.

templates/java/api.mustache
MarioAlexandruDan: Added — `close()` override in `api.mustache` closes the ingestion transporter first (if set), then calls `super.close()`, mirroring the Python PR pattern.

MarioAlexandruDan: Added — `tests/output/java/src/test/java/com/algolia/client/manual/TransformationOptionsTest.java` with tests covering: region validation, composition, error when not configured, `close()` behavior, and `setTransformationOptions` replacing the transporter.

templates/java/api.mustache
```diff
@@ -60,12 +60,35 @@ public class {{classname}} extends ApiClient {
...
-     * Sets the region of the current algolia application to the configuration, this is required to be called if you wish to leverage the transformation pipeline (via the *WithTransformation methods).
-     *
-     * @param region (required)
+     * Sets (or replaces) the ingestion transporter used by {@code *WithTransformation} helpers.
+     * Closes the previous transporter if one exists.
+     * See https://www.algolia.com/doc/libraries/sdk/methods/ingestion
      */
-    public void setTransformationRegion(String region) {
-      this.ingestionTransporter = new IngestionClient(this.authInterceptor.getApplicationId(), this.authInterceptor.getApiKey(), region, this.clientOptions);
+    public void setTransformationOptions(TransformationOptions transformationOptions) throws java.io.IOException {
```
eric-zaharia: This will introduce a Breaking Change: all clients using this method will be required to add a try/catch block for their code to compile. 

I think what you are looking for is `AlgoliaRuntimeException` as that is unchecked.

templates/java/api.mustache
```diff
@@ -92,12 +115,41 @@ public class {{classname}} extends ApiClient {
     {{/hasRegionalHost}}
     {{^hasRegionalHost}}
     public {{classname}}(String appId, String apiKey) {
-        this(appId, apiKey, null);
+        this(appId, apiKey, (ClientOptions) null);
```
eric-zaharia: The cast leaks into other clients than search, which is not functionally incorrect but a bit misleading (check the generated commit).

You could either restrict the cast to the SearchClient or entirely remove the need for it using a static factory method (e.g. `withTransformation`) instead of the constructor. What do you think @Fluf22?

eric-zaharia: Also, at the moment an user using the weird but valid (until now) syntax:
`new SearchClient("app-id", "api-key", null);`
will get a compile error.

This is what the static factory method would solve

templates/java/api.mustache
eric-zaharia: Clarification: this is not the BC per-se as it is a new method, but the fact that it is called in the deprecated `setTransformationRegion` which now also throws `java.io.IOException`

templates/java/api.mustache
```diff
@@ -60,12 +60,35 @@ public class {{classname}} extends ApiClient {
...
+     * See https://www.algolia.com/doc/libraries/sdk/methods/ingestion
      */
-    public void setTransformationRegion(String region) {
-      this.ingestionTransporter = new IngestionClient(this.authInterceptor.getApplicationId(), this.authInterceptor.getApiKey(), region, this.clientOptions);
+    public void setTransformationOptions(TransformationOptions transformationOptions) throws java.io.IOException {
+      IngestionClient previous = this.ingestionTransporter;
+      this.ingestionTransporter = new IngestionClient(
+        this.authInterceptor.getApplicationId(),
+        this.authInterceptor.getApiKey(),
+        transformationOptions.getRegion(),
```
eric-zaharia, line 72: The setter should also have a null-check for the options object

templates/java/api_helpers.mustache
eric-zaharia, line 850: The Javadocs still reference passing `region` at client instantiation. Could you update them to reference `TransformationOptions` instead?

eric-zaharia:
The test is still not properly wired up. Locally it runs and fails, the CI passes because it has a workflow which deletes the path you put it at before generating and running the CTS.

config/generation.config.mjs
```diff
@@ -63,6 +63,7 @@ export const patterns = [
   'clients/algoliasearch-client-java/algoliasearch/src/main/java/com/algolia/BuildConfig.java',
 
   'tests/output/java/build.gradle',
+  '!tests/output/java/src/test/java/com/algolia/client/manual/**',
```
Fluf22: Shouldn't this be:

```suggestion
  '!tests/output/java/src/test/java/com/algolia/manual/**',
```

eric-zaharia (approved):

PR #6292, opened 2026-04-23
feat(csharp): add TransformationOptions for ingestion transporter configuration

MarioAlexandruDan:
## 🧭 What and Why

🎟 JIRA Ticket [API-371:](https://algolia.atlassian.net/browse/API-371)

### Changes included:

- List changes


templates/csharp/api.mustache
```diff
@@ -208,40 +213,75 @@ namespace Algolia.Search.Clients;
...
-            ReadTimeout = _transport._algoliaConfig.ReadTimeout,
-            WriteTimeout = _transport._algoliaConfig.WriteTimeout,
-            Compression = _transport._algoliaConfig.Compression,
-            CustomHosts = _transport._algoliaConfig.CustomHosts
-          },
-          factory
-        );
+            transformationOptions.Region
+          );
+        _ingestionTransporter = new IngestionClient(ingestionConfig, _loggerFactory);
```
Fluf22, line 245: Note for follow-up: both Java and Python close/dispose the previous ingestion transporter when this is called again. C# doesn't, but since IngestionClient doesn't implement IDisposable today, there's nothing to call. Worth tracking as a follow-up when IDisposable is added to the C# clients more broadly.

templates/csharp/api.mustache
```diff
@@ -208,40 +213,75 @@ namespace Algolia.Search.Clients;
...
+      /// Sets the region of the transformation pipeline.
+      /// </summary>
+      /// <param name="region">The region ("us" or "eu")</param>
+      [Obsolete(
+        "SetTransformationRegion is deprecated. Use SetTransformationOptions instead. "
+          + "The old setter forwarded the parent search config to the ingestion transporter. "
+          + "The new approach defaults to Ingestion API defaults and only overrides what's explicitly specified. "
+          + "See https://www.algolia.com/doc/libraries/sdk/methods/ingestion"
+      )]
+      public void SetTransformationRegion(string region)
```
Fluf22: The old signature was `SetTransformationRegion(string region, ILoggerFactory factory = null)`. Dropping the factory parameter is a source-breaking change for callers who were passing it.

clients/algoliasearch-client-csharp/algoliasearch/Clients/TransformationOptions.cs
```diff
@@ -0,0 +1,45 @@
...
+public class TransformationOptions
+{
+  /// <summary>The ingestion region ("eu" or "us"). Required.</summary>
+  public string Region { get; }
+
+  /// <summary>
+  /// Optional ingestion config forwarded to the ingestion transporter.
+  /// Only fields explicitly set here replace the Ingestion API defaults.
+  /// </summary>
+  public IngestionConfig IngestionConfig { get; }
```
Fluf22: I see there's no C# equivalent of Java's `ClientOptions`, which explains why `IngestionConfig` was chosen here. However, accepting a full `IngestionConfig` means users must redundantly provide appId/apiKey/region just to override a timeout, and could accidentally provide different credentials.
`TransformationOptions.Region` is also silently ignored below when IngestionConfig is provided.

I believe for this language, the best is to used a new custom class...

templates/csharp/api.mustache
```diff
@@ -208,40 +213,75 @@ namespace Algolia.Search.Clients;
...
-            ConnectTimeout = _transport._algoliaConfig.ConnectTimeout,
-            ReadTimeout = _transport._algoliaConfig.ReadTimeout,
-            WriteTimeout = _transport._algoliaConfig.WriteTimeout,
-            Compression = _transport._algoliaConfig.Compression,
-            CustomHosts = _transport._algoliaConfig.CustomHosts
-          },
-          factory
-        );
+            transformationOptions.Region
+          );
```
Fluf22: Here, you ignore the `transformationOptions.Region` if `IngestionConfig` is provided
That's one of the reasons I think it would be best to create a new custom class for C#

templates/csharp/api.mustache
```diff
@@ -208,40 +213,99 @@ namespace Algolia.Search.Clients;
...
+        if (transformationOptions.ClientOptions != null)
         {
-          throw new ArgumentException("AppId and ApiKey are required for transformation pipeline");
+          var opts = transformationOptions.ClientOptions;
+          if (opts.CustomHosts != null) ingestionConfig.CustomHosts = opts.CustomHosts;
+          if (opts.ConnectTimeout.HasValue) ingestionConfig.ConnectTimeout = opts.ConnectTimeout;
+          if (opts.ReadTimeout.HasValue) ingestionConfig.ReadTimeout = opts.ReadTimeout;
+          if (opts.WriteTimeout.HasValue) ingestionConfig.WriteTimeout = opts.WriteTimeout;
+          if (opts.Compression.HasValue) ingestionConfig.Compression = opts.Compression.Value;
+          if (opts.DefaultHeaders != null) ingestionConfig.DefaultHeaders = opts.DefaultHeaders;
```
eric-zaharia: This will wipe all existing headers (app id, api key etc) if set and not explicitly passed again by the user, I think merging the existing object with the one received in ClientOptions would be safer

clients/algoliasearch-client-csharp/algoliasearch/Clients/ClientOptions.cs
```diff
@@ -0,0 +1,33 @@
+using System;
+using System.Collections.Generic;
+using Algolia.Search.Models.Common;
+using Algolia.Search.Transport;
+
+namespace Algolia.Search.Clients;
+
+/// <summary>
+/// Optional overrides for the ingestion transporter created by <c>*WithTransformation</c> helpers.
```
eric-zaharia: I think this class name oversells the scope. We have `ClientOptions` in other langs as well but as the general client config primitive. In this case it's something sounding generic but only used in one place (as the docstring also mentions). 

I see 2 options: either add the fields directly on the TransformationOptions object or actually make ClientOptions generally used but that's probably a bigger refactor and out of scope. I'd lean towards the first one

templates/csharp/api.mustache
MarioAlexandruDan: Right, good catch!

Fluf22 (approved):

PR #6293, opened 2026-04-24
feat(go): add TransformationOptions for ingestion transporter configuration

MarioAlexandruDan:
## 🧭 What and Why

🎟 JIRA Ticket [API-374:](https://algolia.atlassian.net/browse/API-374)

### Changes included:

- TransformationOptions fields are flat: Go has no general-purpose ClientOptions primitive, so duplicating 6 fields directly avoids a single-use wrapper.
- ClientOption func(*SearchConfiguration) functional option: Go's idiomatic way to pass optional construction-time config; WithTransformationOptions(opts) fits naturally alongside other future With* options.
- SetTransformationOptions also exposed as a method: Lets callers configure after construction without rebuilding the client, consistent with SetClientApiKey.
-  DefaultHeader direct assignment, not merge: In Go, auth headers are injected per-request from cfg.AppID/cfg.ApiKey, not stored in DefaultHeader, so replacing it doesn't wipe credentials.
- transformation_options.mustache registered only for search client: Only the search client needs the transporter helpers; conditional registration avoids generating a useless file for every other client.


templates/go/configuration.mustache
```diff
@@ -2,6 +2,7 @@
 package {{packageName}}
 
 import (
+  "github.com/algolia/algoliasearch-client-go/v4/algolia/ingestion"
```
Fluf22: Can you add a `{{#isIngestionClient}}` around this import please?
It stripped by goimports, but better be clean from the beginning

tests/output/go/tests/manual/transformation_options_test.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func TestSetTransformationOptionsCreatesTransporter(t *testing.T) {
+	client, err := search.NewClient("appID", "apiKey")
+	require.NoError(t, err)
+	_, err = client.SaveObjectsWithTransformation("index", []map[string]any{{"objectID": "1"}})
+	require.ErrorContains(t, err, "TransformationOptions must be set")
+	require.NoError(t, client.SetTransformationOptions(search.TransformationOptions{Region: "us"}))
+	_, err = client.SaveObjectsWithTransformation("index", []map[string]any{{"objectID": "1"}})
+	require.NotContains(t, err.Error(), "TransformationOptions must be set")
```
Fluf22, line 60: Better assert `err != nil` before calling `err.Error()`


```suggestion
        require.Error(t, err)
	require.NotContains(t, err.Error(), "TransformationOptions must be set")
```

templates/go/client.mustache
```diff
@@ -165,12 +157,56 @@ func (c *APIClient) SetClientApiKey(apiKey string) error {
...
+  }
+  if opts.DefaultHeader != nil {
+    ingestionConfig.DefaultHeader = opts.DefaultHeader
+  }
+  ingestionClient, err := ingestion.NewClientWithConfig(ingestionConfig)
+  if err != nil {
+    return err //nolint:wrapcheck
+  }
+  c.ingestionTransporter = ingestionClient
+  return nil
```
Fluf22, line 206: The SetTransformationOptions method creates the ingestion transporter correctly, but never writes c.cfg.TransformationOptions = &opts

Fluf22 (approved):

PR #6313, opened 2026-04-29
feat(ruby): add TransformationOptions for ingestion transporter configuration

MarioAlexandruDan:
## 🧭 What and Why

🎟 JIRA Ticket [API-375:](https://algolia.atlassian.net/browse/API-375)

### Changes included:

- TransformationOptions fields are flat: Ruby has no general-purpose ClientOptions primitive and keyword args handle optional fields natively, so a nested wrapper would add indirection with no benefit.
- with_transformation class method instead of constructor overload: Ruby has no constructor overloading; a named factory method is the idiomatic way to express "create this object pre-configured for a specific use case."
- _build_ingestion_transporter is a private method prefixed with _: Ruby convention for private-ish helpers; keeps the transporter creation logic isolated without the strict visibility enforcement that would prevent internal calls.
- assert_ingestion_transporter! bang method: Ruby convention, bang suffix signals the method raises on failure. Centralises the "transporter not configured" guard so the three *_with_transformation helpers stay a single assert_ingestion_transporter! call each.
- config.header_params = config.header_params.merge(...) not replace: Ruby's merge returns a new hash leaving the original untouched; auth headers already in config.header_params survive alongside any user-supplied overrides.
- ingestion_helpers.mustache as a separate partial included via {{#isIngestionClient}}: The chunked_push logic belongs to the ingestion client, not the search helpers. Splitting it into its own partial keeps search_helpers.mustache focused and reuses the code without duplication.
- transformation_options stored on the config object, not the client: Consistent with how Ruby configures all other client options (timeouts, compression, requester), the config object is the single place to set up a client before handing it to the constructor.


templates/ruby/search_helpers.mustache
```diff
@@ -1,3 +1,195 @@
...
+
+# Sets (or replaces) the ingestion transporter used by *_with_transformation helpers.
+# See https://www.algolia.com/doc/libraries/ruby/v3/methods/ingestion
+#
+# @param transformation_options [TransformationOptions] transformation options
+def set_transformation_options(transformation_options)
+  @ingestion_transporter = _build_ingestion_transporter(transformation_options)
+end
+
+# @deprecated Use with_transformation or set_transformation_options instead.
```
eric-zaharia: Why add a method you are instantly deprecating? Nobody is using it, the point of deprecation is to discourage existing usage without breaking it. There's no existing usage here

eric-zaharia:
The helper specs and `runCts` were not updated, Ruby is not in the `x-available-languages` list therefore the CTS will not generate tests for them

templates/ruby/search_helpers.mustache
```diff
@@ -1,3 +1,158 @@
...
+  @ingestion_transporter = _build_ingestion_transporter(transformation_options)
+end
+
+# @deprecated Use {SearchClient.with_transformation} or {#set_transformation_options} instead.
+#   The old setter forwarded the parent search config to the ingestion transporter, causing search
+#   timeouts to bleed into ingestion calls. Use TransformationOptions to set Ingestion API defaults.
+def set_transformation_region(region)
+  warn '[DEPRECATION] set_transformation_region is deprecated. Use set_transformation_options with a TransformationOptions object instead.'
+  set_transformation_options(Algolia::TransformationOptions.new(region))
+end
```
Fluf22: No need to create a new method that you're deprecating right away
We are deprecating existing methods in the other languages if they existed before

templates/ruby/search_helpers.mustache
```diff
@@ -1,3 +1,158 @@
...
+  )
+
+  if co&.header_params
+    config.header_params = config.header_params.merge(co.header_params)
+  end
+
+  Algolia::IngestionClient.create_with_config(config)
+end
+
+def set_transformation_options(transformation_options)
```
Fluf22, line 36: Can you guard against passing `nil` here, please?

templates/ruby/api.mustache
```diff
@@ -8,6 +8,40 @@ require 'base64'
...
+  # the ingestion transporter is eagerly created using Ingestion API defaults (25s timeouts,
+  # no compression). Pass a ClientOptions as the second argument to override specific defaults.
+  # See https://www.algolia.com/doc/libraries/ruby/v3/methods/ingestion
+  class TransformationOptions
+    attr_accessor :region, :client_options
+
+    def initialize(region, client_options = nil)
+      raise ArgumentError, "`region` is required in `TransformationOptions`. See https://www.algolia.com/doc/libraries/ruby/v3/methods/ingestion" if region.nil? || region.to_s.strip.empty?
+      @region = region
+      @client_options = client_options.is_a?(Hash) ? ClientOptions.new(client_options) : client_options
```
Fluf22: Shouldn't we test that the hash actually match the expected shape for ClientOptions?

templates/ruby/search_helpers.mustache
```diff
@@ -1,3 +1,158 @@
+# The parent search config MUST NOT leak into the ingestion transporter.
+def _build_ingestion_transporter(transformation_options)
+  co = transformation_options.client_options
```
Fluf22: Can we use `client_options` instead of `co`, please?
Or is it idiomatic?

I don't really find it easy to understand

templates/ruby/search_helpers.mustache
```diff
@@ -1,3 +1,184 @@
...
+    search_watch_responses = watch_responses.map do |wr|
+      Search::WatchResponse.build_from_hash(wr.to_hash)
+    end
+
+    Search::ReplaceAllObjectsWithTransformationResponse.new(
+      copy_operation_response: copy_operation_response,
+      watch_responses: search_watch_responses,
+      move_operation_response: move_operation_response
+    )
+  rescue StandardError => e
```
Fluf22: Is there a specific reason you didn't use

```suggestion
  rescue Exception => e
```

templates/ruby/ingestion_helpers.mustache
```diff
@@ -0,0 +1,53 @@
...
+# @param wait_for_tasks [Boolean] whether to wait until every task has been processed. (optional, default: false)
+# @param batch_size [Integer] the size of each chunk of objects sent in a single push call. (optional, default: 1000)
+# @param reference_index_name [String] the reference index name used for replace-all operations. (optional)
+# @param request_options [Hash] the request options to send along with the query. (optional)
+#
+# @return [Array<Ingestion::WatchResponse>]
+def chunked_push(index_name, objects, action = Ingestion::Action::ADD_OBJECT, wait_for_tasks = false, batch_size = 1000, reference_index_name = nil, request_options = {})
+  responses = []
+  offset = 0
+  wait_batch_size = [batch_size / 10, 1].max
```
Fluf22: We want to fallback on batch_size, I believe

```suggestion
  wait_batch_size = batch_size / 10
  wait_batch_size = batch_size if wait_batch_size < 1
```

templates/ruby/ingestion_helpers.mustache
```diff
@@ -0,0 +1,54 @@
...
+      false,
+      reference_index_name,
+      request_options
+    )
+    responses << response
+
+    if wait_for_tasks && (responses.length % wait_batch_size == 0 || batch_index == total_batches - 1)
+      responses[offset, wait_batch_size].each do |watch_response|
+        if watch_response.event_id.nil?
+          raise ArgumentError, "received unexpected response from the push endpoint, eventID must not be undefined"
```
Fluf22: undefined -> nil
```suggestion
          raise ArgumentError, "received unexpected response from the push endpoint, eventID must not be nil"
```

templates/ruby/search_helpers.mustache
```diff
@@ -1,3 +1,184 @@
...
+  end
+
+  Algolia::IngestionClient.create_with_config(config)
+end
+
+# Helper: Sets (or replaces) the ingestion transporter used by *_with_transformation helpers.
+#
+# @param transformation_options [TransformationOptions] the transformation options including region and optional ingestion transporter overrides. (required)
+def set_transformation_options(transformation_options)
+  raise ArgumentError, "`transformation_options` must be a TransformationOptions instance, got #{transformation_options.inspect}" if transformation_options.nil?
```
Fluf22:
```suggestion
  raise ArgumentError, "`transformation_options` must not be nil" if transformation_options.nil?
```

Fluf22 (approved):

PR #6319, opened 2026-04-30
fix(csharp): ReplaceAllObjectsAsync intermittently fails with "maximum number of retries exceeded (51/50)" response (C# API client)

MarioAlexandruDan:
## 🧭 What and Why

We have a scheduled task that runs method ReplaceAllObjectsAsync is run on the prod_shawcontract_resource index weekly on Sundays, starting at 12pm UTC, and usually taking about 20-30 minutes to complete, but fails more often than not with the response “maximum number of retries exceeded (51/50)”. 

🎟 JIRA Ticket [CR-11342:](https://algolia.atlassian.net/browse/CR-11342)

### Changes included:

- List changes


MarioAlexandruDan:
already started in https://github.com/algolia/api-clients-automation/pull/6318

PR #6321, opened 2026-04-30
feat(dart): add TransformationOptions for ingestion transporter configuration

MarioAlexandruDan:
## 🧭 What and Why

🎟 JIRA Ticket [API-379:](https://algolia.atlassian.net/browse/API-379)

### Changes included:

- **`TransformationOptions`** added as a dedicated named parameter on the generated `SearchClient` constructor (gated by `{{#isSearchClient}}`). It does **not** live on the generic `ClientOptions`, so no other client (Ingestion, Composition, Recommend, Insights, Abtesting, …) can receive or silently forward it. No breaking changes to `ClientOptions` — its public API is identical to pre-PR.
- `TransformationOptions.ingestionClientOptions: ClientOptions?` — reuses Dart's existing general-purpose `ClientOptions` to configure the ingestion transporter, instead of duplicating 8 fields as other languages had to.
- `setTransformationOptions(...)` post-init method on `SearchClient`, matching C#'s `SetTransformationOptions`.
- `import 'dart:io' as io`: The ingestion spec generates a `Platform` model that shadows `dart:io.Platform` at import time.
- `lib/src/extension.dart` placeholder in `client_ingestion`: `lib.mustache` unconditionally exports this file in every package barrel; missing it breaks the build.
- `DartCTSManager` snake_case → camelCase: Ingestion spec uses snake_case; CTS generator passed raw names producing uncompilable Dart constructor calls.
- `param_object.mustache` aggregates `additionalProperties`: Old template generated one entry per extra field; Dart forbids duplicate named args in a single call.
- `runCts.ts` `dart` in both mock validators: `replaceAllObjectsWithTransformation` calls `chunkedPush` internally, hitting the pushMock server indirectly.

## 🧪 Test


clients/algoliasearch-client-dart/packages/client_search/lib/src/extension/transformation.dart
```diff
@@ -0,0 +1,258 @@
...
+        requestOptions: requestOptions,
+      );
+      return;
+    } on AlgoliaApiException catch (e) {
+      if (e.statusCode != 404) rethrow;
+    }
+    await Future<void>.delayed(
+      Duration(milliseconds: (retries * 1500).clamp(0, 5000)),
+    );
+  }
```
Fluf22: Silently returns after 50 retries
Add a throw exception at the end

clients/algoliasearch-client-dart/packages/client_search/lib/src/extension/transformation.dart
```diff
@@ -0,0 +1,258 @@
...
+}
+
+Future<void> _waitForEvent({
+  required ingestion.IngestionClient transporter,
+  required String runID,
+  required String eventID,
+  RequestOptions? requestOptions,
+}) async {
+  for (var retries = 0; retries < 50; retries++) {
```
Fluf22: The default is 100 now, and we want to pass maxRetries from the calling method I believe (check Dart jira ticket for maxRetries update)

templates/dart/api.mustache
```diff
@@ -72,13 +81,53 @@ final class {{classname}} implements ApiClient {
...
+  ingestion.IngestionClient _buildIngestionTransporter(TransformationOptions opts) {
+    final ingestionOpts = opts.ingestionClientOptions;
+    return ingestion.IngestionClient(
+      appId: _appId,
+      apiKey: _apiKey,
+      region: opts.region,
+      options: ClientOptions(
+        connectTimeout: ingestionOpts?.connectTimeout ?? const Duration(seconds: 25),
+        readTimeout: ingestionOpts?.readTimeout ?? const Duration(seconds: 25),
+        writeTimeout: ingestionOpts?.writeTimeout ?? const Duration(seconds: 25),
```
Fluf22: Shouldn't we let the ingestion code default the right duration?
Because if we change in the future, we would have two places to change instead of one

templates/dart/api.mustache
```diff
@@ -172,6 +221,10 @@ final class {{classname}} implements ApiClient {
           writeTimeout: Duration(milliseconds: {{write}}),
           readTimeout: Duration(milliseconds: {{read}}),
           connectTimeout: Duration(milliseconds: {{connect}}),
+      ) + {{/vendorExtensions.x-timeouts}}{{^vendorExtensions.x-timeouts}}RequestOptions(
+          writeTimeout: Duration(milliseconds: {{{serverWriteTimeout}}}),
+          readTimeout: Duration(milliseconds: {{{serverReadTimeout}}}),
+          connectTimeout: Duration(milliseconds: {{{serverConnectTimeout}}}),
```
Fluf22: Not sure to understand why you added that

Fluf22 (commented):
I don't understand the changes you made around server timeouts, I might be missing something

MarioAlexandruDan:
> The default is 100 now, and we want to pass maxRetries from the calling method I believe (check Dart jira ticket for maxRetries update)

Related PR: configurable retries on chunked helpers via ChunkedHelperOptions:
https://github.com/algolia/api-clients-automation/pull/6405

templates/dart/api.mustache
MarioAlexandruDan: Added inline comment for better understanding

clients/algoliasearch-client-dart/packages/client_search/lib/src/extension/transformation.dart
```diff
@@ -0,0 +1,262 @@
...
+        runID: runID,
+        eventID: eventID,
+        requestOptions: requestOptions,
+      );
+      return;
+    } on AlgoliaApiException catch (e) {
+      if (e.statusCode != 404) rethrow;
+    }
+    await Future<void>.delayed(
+      Duration(milliseconds: (retries * 1500).clamp(0, 5000)),
```
Fluf22: This makes the first retry wait 0ms, and thus making it useless, if I'm not mistaken

templates/dart/api.mustache
Fluf22: Okay, I was asking for myself, not for documentation: I don't think the comment is useful here, especially since it will be put on every method, adding a lot of noise

Now I understand better. This is not where this should live: default timeouts per client should be at the client init level, overloading the ClientOptions default when it calls the constructor

... which should make client-level timeouts available without the need to inject the new variables ( 🤞🏻 )

clients/algoliasearch-client-dart/packages/client_ingestion/lib/src/extension.dart
Fluf22, line 1: 👻

MarioAlexandruDan: Created a PR ticket: https://algolia.atlassian.net/browse/API-495

clients/algoliasearch-client-dart/packages/client_search/lib/src/extension/transformation.dart
```diff
@@ -0,0 +1,288 @@
...
+      if (e.statusCode != 404) rethrow;
+    }
+    await Future<void>.delayed(
+      Duration(milliseconds: min((retries + 1) * 1500, 5000)),
+    );
+  }
+  throw StateError(
+    'The maximum number of retries exceeded. ($maxRetries/$maxRetries)',
+  );
+}
```
Fluf22, line 264: Isn't this the same thing than `waitUntil`?
If not, can you add a small documentation above the method so we understand its specific behaviour, please?

Fluf22 (commented):
Just one question

clients/algoliasearch-client-dart/packages/client_core/lib/src/transport/retry_strategy.dart
```diff
@@ -23,29 +23,57 @@ final class RetryStrategy {
...
+    final readTimeout = options.readTimeout == ClientOptions.unsetTimeout
+        ? defaultReadTimeout
+        : options.readTimeout;
+    final writeTimeout = options.writeTimeout == ClientOptions.unsetTimeout
+        ? defaultWriteTimeout
+        : options.writeTimeout;
+
+    // A caller-provided requester (e.g. the CTS echo requester) carries its own
+    // connect timeout, so push the resolved value onto it; the default
+    // [DioRequester] receives it through its constructor.
```
Fluf22: No need for a comment

clients/algoliasearch-client-dart/packages/client_core/test/retry_strategy_timeout_test.dart
Fluf22, line 1: Don't we already have tests generated for the RetryStrategy?

clients/algoliasearch-client-dart/packages/client_core/test/retry_strategy_timeout_test.dart
```diff
@@ -0,0 +1,89 @@
...
+}
+
+class _StopException implements Exception {}
+
+void main() {
+  Iterable<Host> defaultHosts() => [const Host(url: 'localhost')];
+
+  // The ingestion spec sets every server timeout to 25s; the generated client
+  // forwards those as the per-client defaults below.
+  const ingestionDefault = Duration(milliseconds: 25000);
```
Fluf22: If the ingestion spec sets the default to 25s, why do you have to add it here?

clients/algoliasearch-client-dart/packages/client_core/test/retry_strategy_timeout_test.dart
```diff
@@ -0,0 +1,89 @@
...
+      final requester = _CaptureRequester();
+      final strategy = build(ClientOptions(requester: requester));
+
+      expect(strategy.readTimeout, ingestionDefault);
+      expect(strategy.writeTimeout, ingestionDefault);
+      // The resolved connect timeout is pushed onto the provided requester.
+      expect(requester.connectTimeout, ingestionDefault);
+
+      // A read call should go out with the per-client read + connect timeouts —
+      // this is what the "calls api with default read timeouts" CTS test asserts.
```
Fluf22: If already asserted, I don't see the value of adding a redundant test but I might be missing something here

clients/algoliasearch-client-dart/packages/client_search/lib/src/extension/transformation.dart
```diff
@@ -0,0 +1,304 @@
...
+            from: polledUpTo,
+            to: responses.length,
+            maxRetries: maxRetries,
+            requestOptions: requestOptions,
+          );
+          polledUpTo = responses.length;
+        }
+      }
+
+      if (isLast) break;
```
Fluf22, line 77: Sorry I forgot: can't we reuse the `_waitUntil` method here?

clients/algoliasearch-client-dart/packages/client_search/lib/src/extension/transformation.dart
```diff
@@ -0,0 +1,304 @@
...
+      return ReplaceAllObjectsWithTransformationResponse(
+        copyOperationResponse: copyResponse,
+        watchResponses: watchResponses,
+        moveOperationResponse: moveResponse,
+      );
+    } catch (_) {
+      try {
+        await deleteIndex(indexName: tmpIndex);
+      } catch (_) {}
+      rethrow;
```
Fluf22, line 213: In other languages, do we usually rethrow the parent exception or override it with why we couldn't execute `deleteIndex`?

clients/algoliasearch-client-dart/packages/client_search/lib/src/extension/transformation.dart
```diff
@@ -0,0 +1,304 @@
...
+    );
+  }
+}
+
+/// Polls the Ingestion API for a single event until it becomes available.
+///
+/// Unlike [waitTask]'s generic `_waitUntil` (which retries until a *result*
+/// predicate holds), this is exception-driven: `getEvent` returns 404 until the
+/// event exists, so we retry on 404 and stop on the first successful response.
+/// It therefore can't reuse `_waitUntil` directly.
```
Fluf22: Okay you added a comment but I don't think it holds: I believe that's literally what we do for `waitKeyCreation`

clients/algoliasearch-client-dart/packages/client_core/test/retry_strategy_timeout_test.dart
MarioAlexandruDan: The hardcoded 25s was only there to feed the unit test

clients/algoliasearch-client-dart/packages/client_search/lib/src/extension/transformation.dart
MarioAlexandruDan: We rethrow the parent (original) exception same as Java (catch (e) { deleteIndex(...); throw e; }). 
Dart additionally guards the deleteIndex cleanup so a cleanup failure can't mask the original error.

clients/algoliasearch-client-dart/packages/client_search/lib/src/extension/transformation.dart
MarioAlexandruDan: The chunkedPush loop itself is just batch iteration, but the event-polling it triggers (_pollBatch → _waitForEvent) now goes through waitUntil.

Fluf22: Okay 👍🏻

templates/dart/api.mustache
```diff
@@ -72,13 +88,48 @@ final class {{classname}} implements ApiClient {
...
+        apiKey: _apiKey,
+        region: opts.region,
+        options: ingestionOpts,
+      );
+    }
+    return ingestion.IngestionClient(
+      appId: _appId,
+      apiKey: _apiKey,
+      region: opts.region,
+    );
```
Fluf22: Both branches do the same thing: you can collapse


```suggestion
    return ingestion.IngestionClient(
      appId: _appId,
      apiKey: _apiKey,
      region: opts.region,
      options: ingestionOpts,
    );
```

Fluf22 (commented):
Okay I believe this is my final comment 😁

Fluf22 (approved):
Let's go!

PR #6338, opened 2026-05-05
fix(clients): validate empty string for required string parameters

### LINKED ISSUE #872: Passing empty objectId to DeleteObject deletes the ENTIRE index, configuration, settings.
Passing an empty string as objectId to the SearchClient.DeleteObject deletes the entire index (not just a record, but the entire index, configuration, settings, etc).  Example mySearchClient.DeleteObjectAsync("my_index", "")

https://github.com/algolia/algoliasearch-client-csharp/blob/32ecd39fb48d9b5b6ac94e5c257a1bc91da54126/algoliasearch/Clients/SearchClient.cs#L6654

  1. The null check is the only guard (SearchClient.cs:6654):
  if (objectID == null)                                                
    throw new ArgumentException("Parameter `objectID` is required  when calling `DeleteObject`.");                                    
  An empty string passes this check.

  2. ParameterToString("") returns "" (QueryStringHelper.cs:32-53)
  — the default branch hits Convert.ToString("", ...) which is just
   "". No empty-string handling.

  3. The path template gets a literal empty substitution
  (HttpTransport.cs:399):
  path = path.Replace("{" + parameter.Key + "}",
  Uri.EscapeDataString(parameter.Value));
  Template /1/indexes/{indexName}/{objectID} becomes
  /1/indexes/myIndex/ (trailing slash, empty segment).
  Uri.EscapeDataString("") is "", so no protection there either.

  4. What Algolia's API does with DELETE /1/indexes/myIndex/ — the
  delete-index endpoint is DELETE /1/indexes/{indexName} (you can
  see it just above at SearchClient.cs:6628). Algolia's gateway
  normalizes trailing slashes, so DELETE /1/indexes/myIndex/ is
  treated as DELETE /1/indexes/myIndex — which deletes the entire
  index (records + settings). It is not the same as "clear index"
  (POST /1/indexes/{name}/clear); a delete is more destructive
  because it also drops settings, replicas config, etc.

  Conclusion: Calling DeleteObject(indexName, "") with the C# SDK
  at this commit will send a request that Algolia routes to the
  delete-index endpoint and wipes the whole index. The SDK does not
   validate against empty, and the URL builder happily produces the
   trailing-slash path. Given your recent commits (e996d8f
  Implement null/empty item_id check, e24a68c Don't do Algolia
  deletes), it sounds like you've already homed in on this — the
  null/empty guard at the call site is the right fix; do not rely
  on the SDK to catch it.

Fluf22:
[API-414]

## Summary

fixes algolia/algoliasearch-client-csharp#872

Passing an empty string as a required path parameter (e.g. `objectID = ""`) bypasses null guards and corrupts the URL path — causing the wrong API endpoint to be called (e.g. delete index instead of delete object).

This PR adds empty string validation for required string parameters across the **7 vulnerable languages**. 4 languages (JavaScript, Go, Swift, Kotlin) were already safe.

### Templates (guard empty string for `{{#isString}}` + `{{#required}}` params)

| Language | Guard added | File |
|----------|-------------|------|
| **C#** | `string.IsNullOrWhiteSpace()` | `templates/csharp/api.mustache` |
| **Java** | `Parameters.requireNonEmpty()` | `templates/java/api.mustache` |
| **Python** | `if not paramName:` (falsy) | `templates/python/api.mustache` |
| **Ruby** | `.empty?` (respects `client_side_validation` flag) | `templates/ruby/api.mustache` |
| **PHP** | `=== ''` strict comparison | `templates/php/api.mustache` |
| **Scala** | `requireNotEmpty()` | `templates/scala/api.mustache` |
| **Dart** | `throw ArgumentError` (replaces `assert()` which was stripped in release builds) | `templates/dart/api.mustache` |

### Hand-written utilities

| File | Change |
|------|--------|
| `clients/.../java/.../utils/Parameters.java` | Added `requireNonEmpty(String, String)` |
| `clients/.../scala/.../internal/util/package.scala` | Added `requireNotEmpty(String, String)` |

### Dart bonus fix

Removed a duplicate buggy line in the `isFreeFormObject` block (had trailing spaces in variable name and error message — copy-paste artifact).

## Already safe (no changes needed)

| Language | Existing guard |
|----------|---------------|
| **JavaScript** | `!paramName` (falsy catches empty) |
| **Go** | `paramName == ""` explicit check |
| **Swift** | `guard !paramName.isEmpty` |
| **Kotlin** | `require(paramName.isNotBlank())` |

## Next steps

After merge, regenerate clients for the 7 affected languages:
```bash
yarn cli generate csharp java python ruby php scala dart
```

[API-414]: https://algolia.atlassian.net/browse/API-414?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/clients/algoliasearch-client-java/algoliasearch/src/main/java/com/algolia/utils/Parameters.java
+++ b/clients/algoliasearch-client-java/algoliasearch/src/main/java/com/algolia/utils/Parameters.java
@@ -13,4 +13,10 @@ public class Parameters {
       throw new AlgoliaRuntimeException(error);
     }
   }
+
+  public static void requireNonEmpty(String param, String error) {
+    if (param == null || param.trim().isEmpty()) {
+      throw new AlgoliaRuntimeException(error);
+    }
+  }
 }
--- a/clients/algoliasearch-client-scala/src/main/scala/algoliasearch/internal/util/package.scala
+++ b/clients/algoliasearch-client-scala/src/main/scala/algoliasearch/internal/util/package.scala
@@ -16,6 +16,10 @@ package object util {
     if (param == null) throw new IllegalArgumentException(error)
   }
 
+  private[algoliasearch] def requireNotEmpty(param: String, error: String): Unit = {
+    if (param == null || param.trim.isEmpty) throw new IllegalArgumentException(error)
+  }
+
   /** Escape the given string to be used as URL query value.
     */
   private[algoliasearch] def escape(input: Any): String = input match {
--- a/templates/csharp/api.mustache
+++ b/templates/csharp/api.mustache
@@ -261,6 +261,10 @@ namespace Algolia.Search.Clients;
           if ({{paramName}} == null)
               throw new ArgumentException("Parameter `{{paramName}}` is required when calling `{{operationId}}`.");
             {{/isEnumRef}}{{/vendorExtensions.x-csharp-value-type}}
+          {{#isString}}
+          if (string.IsNullOrWhiteSpace({{paramName}}))
+              throw new ArgumentException("Parameter `{{paramName}}` is required when calling `{{operationId}}`.");
+          {{/isString}}
           {{/required}}
           {{/allParams}}
           var requestOptions = new InternalRequestOptions(options);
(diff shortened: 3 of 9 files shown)


MarioAlexandruDan (approved):
🆗

PR #6347, opened 2026-05-09
fix(javascript): handle response stream error in `createHttpRequester`

zigzagdev:
## 🧭 What and Why

The response stream in `createHttpRequester` had no `error` event handler.
According to the Node.js stream documentation, if an `error` event handler is not added, the error will be thrown as an unhandled exception.

ref: https://nodejs.org/api/stream.html#event-error

### Changes included:

When a network error occurs during response streaming:
- `response.on('error')` was not handled
- `connectTimeout` and `responseTimeout` were never cleared
- Node.js would throw an unhandled exception

Added `response.on('error')` to properly clear timeouts and resolve
the promise on response stream errors.

## 🧪 Test
スクリーンショット 2026-05-09 15 33 02


Fluf22 (approved):

Fluf22:
Thank you for your contribution!

PR #6349, opened 2026-05-11
feat(php): add TransformationOptions for ingestion transporter

eric-zaharia:
## 🧭 What and Why

🎟 JIRA Ticket: [API-372]

Brings PHP to parity with Java/Python/Scala/JS by introducing `TransformationOptions` as the configurable surface for the existing `*WithTransformation` write helpers (`saveObjectsWithTransformation`, `partialUpdateObjectsWithTransformation`, `replaceAllObjectsWithTransformation`).

The previous PHP shape exposed only `SearchConfig::setTransformationRegion`, which had two problems: the parent `SearchConfig`'s `hosts` leaked into the ingestion transporter, and there was no way to override timeouts, headers, or compression for the ingestion path. This PR adds a fluent `TransformationOptions` class, builds the ingestion transporter from Ingestion API defaults, and only applies the fields the caller explicitly sets — `SearchConfig` no longer leaks.

### Changes included
- Add `TransformationOptions` class (`region` required, fluent setters for `hosts`/`fullHosts`, `read`/`write`/`connectTimeout`, `waitTaskTimeBeforeRetry`, `defaultHeaders`, `compressionType`)
- Add `SearchConfig::setTransformationOptions(...)` for configuring the transporter before client construction
- Add `SearchClient::setTransformationOptions(...)` for swapping the transporter on an already-built client
- Build the ingestion transporter from Ingestion defaults, applying only the overrides on `TransformationOptions` — `SearchConfig` hosts/timeouts/headers no longer leak in
- Deprecate `SearchConfig::setTransformationRegion` with `@deprecated`; it still works for back-compat but can't affect an already-built client
- Update `api.mustache` / `client_config.mustache` / `createClient.mustache` to emit the new shape
- Allowlist `TransformationOptions.php` in `generation.config.mjs` so the pre-commit hook doesn't strip it as generated

## 🧪 Test
- `yarn cli cts run php search` → all tests pass

[API-372]: https://algolia.atlassian.net/browse/API-372


templates/php/api.mustache
```diff
@@ -107,18 +108,8 @@ use Algolia\AlgoliaSearch\Exceptions\NotFoundException;
...
-
-            if ($hosts = $config->getHosts()) {
-                if ($config->getHasFullHosts()) {
-                    $ingestionConfig = $ingestionConfig->setFullHosts($hosts);
-                } else {
-                    $ingestionConfig = $ingestionConfig->setHosts($hosts);
-                }
-            }
-
-            $client->ingestionTransporter = IngestionClient::createWithConfig($ingestionConfig);
```
Fluf22, line 121: Don't you believe this is a breaking change?
If a customer is using this method today, the default is to forward hosts from the search client to the ingestion client, and they might be custom: this behavior is lost with the new version, and they could fallback on the default hosts

templates/php/api.mustache
Fluf22, line 48: Is this generated for all clients?
This is out of what you changed, but it feels overlooked. We should wrap this in an `if search client` right?

templates/php/api.mustache
eric-zaharia: The way I see it this was part of the leak that we are trying to fix: hosts from the Search Config leaked into the ingestion transporter config. This way, they have to be explicitly set, else defaults are used. It's also what we did for Java, Python.

Indeed customers may have relied on this behaviour but still I'd frame it as a fix rather than a BC

Fluf22 (approved):

PR #6353, opened 2026-05-12
feat(javascript): expose maximum number of retries

eric-zaharia:
## 🧭 What and Why

🎟 JIRA Ticket: [API-472](https://algolia.atlassian.net/browse/API-472)

Mirrors the C# work in #6318 for the JavaScript client: makes `maxRetries` configurable on the chunked / long-polling helpers so callers aren't stuck with the 100-retry default when they need a longer (or shorter) ceiling.

### Changes

- Adds an optional `maxRetries` field to the option bags of every helper that polls `waitForTask` under the hood:
  - `chunkedBatch`, `saveObjects`, `deleteObjects`, `partialUpdateObjects`, `replaceAllObjects`, `accountCopyIndex` (search)
  - `chunkedPush` (ingestion)
- Threads the value through to each internal `waitForTask` call so it actually takes effect (defaults to 100, unchanged behavior when omitted).
- Introduces a shared `ChunkedHelperOptions` type in `@algolia/client-common` and mixes it into each helper's options type. Name + intent matches the C# `ChunkedHelperOptions` class so the two clients stay aligned, and gives us one place to add future shared helper config without growing every option type individually.

## 🧪 Test

- CTS passes for the JS client

[API-472]: https://algolia.atlassian.net/browse/API-472?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/clients/algoliasearch-client-javascript/packages/client-common/src/types/chunkedHelperOptions.ts
+++ b/clients/algoliasearch-client-javascript/packages/client-common/src/types/chunkedHelperOptions.ts
@@ -1,0 +1,10 @@
+/**
+ * Shared configuration for chunked helpers that poll for task completion.
+ * Designed to grow over time so future helper configs can be added without ballooning option types.
+ */
+export type ChunkedHelperOptions = {
+  /**
+   * The maximum number of retries when polling for task completion. 100 by default.
+   */
+  maxRetries?: number | undefined;
+};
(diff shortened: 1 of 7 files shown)


Fluf22 (approved):
`maxRetries` default went from 50 to 100 (mentioning it for future reference)

PR #6401, opened 2026-05-12
feat(php): expose maximum number of retries

eric-zaharia:
## 🧭 What and Why

🎟 JIRA Ticket: [API-471](https://algolia.atlassian.net/browse/API-471)

Mirrors the C# work in #6318 and the JavaScript work in #6353 for the PHP client: makes `maxRetries` configurable on every chunked / long-polling helper so callers aren't stuck with the default ceiling when they need a longer (or shorter) one.

### Changes

- New `Algolia\AlgoliaSearch\Support\ChunkedHelperOptions` class. One field for now (`?int $maxRetries`); designed to grow so future shared helper config (e.g. timeout, batchSize defaults) can be added without ballooning each helper's signature. Name matches the C# class so the two stay aligned.
- Added an optional trailing `?ChunkedHelperOptions $chunkedOptions = null` parameter to:
  - `chunkedBatch`, `saveObjects`, `deleteObjects`, `partialUpdateObjects`, `replaceAllObjects` (search)
  - `chunkedPush` (ingestion)
  - `saveObjectsWithTransformation`, `partialUpdateObjectsWithTransformation`, `replaceAllObjectsWithTransformation`
- Threads the value through to every internal `waitForTask` and to the `chunkedPush` poll loop. Forwarders pass the options object through unchanged; the consuming layer (`waitForTask` resolving via `SearchConfig::getDefaultMaxRetries()` on search, `chunkedPush` resolving to `100` on ingestion) owns the default.
- **Ingestion `chunkedPush` default bumped 50 → 100.** The previous loop was hardcoded to 50; callers can now lift it via `new ChunkedHelperOptions(maxRetries: 200)`. Flagging here so release notes don't miss it.
- Harmonized the poll-exhausted error message in `chunkedPush` to match the wording already used by `Support\Helpers::retryUntil`:
  > Stopped waiting for the task after {maxRetries} retries. This does not mean the operation failed; it may still complete. If you need to keep polling, retry with a higher \$maxRetries.

### Usage

```php
$client->replaceAllObjects(
    $indexName,
    $objects,
    chunkedOptions: new ChunkedHelperOptions(maxRetries: 200),
);
```

Existing call sites are unaffected — the new parameter is optional and trailing.

## 🧪 Test

- `yarn cli cts run php` — 941 tests, 4322 assertions, 0 failures.

[API-471]: https://algolia.atlassian.net/browse/API-471

### DIFF
--- a/clients/algoliasearch-client-php/lib/Support/ChunkedHelperOptions.php
+++ b/clients/algoliasearch-client-php/lib/Support/ChunkedHelperOptions.php
@@ -1,0 +1,16 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Support;
+
+/**
+ * Optional configuration for chunked helpers that batch records and poll for task completion.
+ *
+ * Designed to grow over time; future shared helper config (e.g. timeout, batchSize defaults)
+ * should be added here instead of widening every helper's parameter list.
+ */
+final class ChunkedHelperOptions
+{
+    public function __construct(
+        public ?int $maxRetries = null,
+    ) {}
+}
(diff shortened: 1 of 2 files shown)


Fluf22 (approved):
Nothing to add, great work

PR #6406, opened 2026-05-14
feat(go): expose number of retries

eric-zaharia:
## 🧭 What and Why

🎟 JIRA Ticket: [API-412](https://algolia.atlassian.net/browse/API-412)

Mirrors the C# work in #6318, the JavaScript work in #6353 and the PHP work in #6401 for the Go client: makes `MaxRetries` configurable on every chunked / long-polling helper so callers aren't stuck with the default ceiling when they need a longer (or shorter) one.

### Changes

- New unexported `chunkedHelperOption` named function type (emitted into both the `search` and `ingestion` packages) that satisfies `ChunkedBatchOption`, `IterableOption`, `PartialUpdateObjectsOption`, `ReplaceAllObjectsOption`, and `WaitForApiKeyOption`. Lets a single `WithMaxRetries(...)` value flow through every helper signature. Equivalent in intent to the `ChunkedHelperOptions` class in the sibling clients — kept as a bridge type rather than an options struct so the existing variadic functional-option API (`WithBatchSize`, `WithWaitForTasks`, …) stays consistent.
- Re-aimed `WithMaxRetries(maxRetries int)` to return `chunkedHelperOption` (previously returned `iterableOption`). Non-breaking: the returned value still satisfies `IterableOption`, and now also satisfies `ChunkedBatchOption` and friends, so existing `WaitForTask` / `CreateIterable` call sites keep working while `ChunkedBatch` / `SaveObjects` / `ReplaceAllObjects` / `PartialUpdateObjects` gain support.
- Threaded the value through `ChunkedPush` (ingestion) and the search→ingestion forwarder `toIngestionChunkedBatchOptions`, so a `WithMaxRetries(...)` passed to a search helper reaches the ingestion-side polling loop.
- **Default polling budget bumped `50` → `100`** in `WaitForTask`, `WaitForAppTask`, `WaitForApiKey` (search) and `ChunkedPush` (ingestion). Matches the global bump in C# #6318 and the explicit `100` defaults in JS #6353 and PHP #6401.
- Adopted the friendlier poll-exhausted error wording from PHP #6401 in `CreateIterable` — replaces `"The maximum number of retries exceeded. (X/Y)"` with `"Stopped waiting for the task after N retries. This does not mean the operation failed; it may still complete. If you need to keep polling, retry with a higher maxRetries."`.

### Usage

```go
client.SaveObjects(indexName, objects, search.WithMaxRetries(200))

client.WaitForTask(indexName, taskID, search.WithMaxRetries(200))
```

Existing call sites are unaffected — `WithMaxRetries` remains a variadic option and behaviour is unchanged when it's not passed.

[API-412]: https://algolia.atlassian.net/browse/API-412?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Fluf22 (approved):
Looks good 🆗

PR #6428, opened 2026-05-18
feat(dart): configurable retries on chunked helpers via ChunkedHelperOptions

MarioAlexandruDan:
## 🧭 What and Why

🎟 JIRA Ticket [API-415]:

### Changes included:

Introduce a ChunkedHelperOptions configuration class for all Dart chunked helpers that batch records and poll for task completion. Bump the default defaultMaxRetries from 50 → 100.

Why:
Customers hitting "maximum number of retries exceeded" on long-running batch operations (e.g. weekly reindex via replaceAllObjects) because the polling budget was hardcoded at 50. The wait helper gives up before the operation actually completes, misleading users into thinking the operation failed.

Changes:

New ChunkedHelperOptions class exposing maxRetries (defaults to defaultMaxRetries). Designed as an options object so future helper configs (e.g. batchSize) can be added without ballooning method signatures.

Add optional ChunkedHelperOptions? chunkedOptions parameter to all chunkedBatch, saveObjects, replaceAllObjects, deleteObjects, partialUpdateObjects helpers (including *WithTransformation variants) and chunkedPush on the ingestion side. Non-breaking — defaults to null, resolved internally via chunkedOptions?.maxRetries ?? defaultMaxRetries.

Bump defaultMaxRetries: 50 → 100.

## 🧪 Test

[API-415]: https://algolia.atlassian.net/browse/API-415?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


eric-zaharia:
The PR description and ticket mention the implementation of a `ChunkedHelperOptions` object. The purpose of this object is to not further extend method signatures in the future, when we might want to add more configurable params. This PR only threads the maxRetries param through methods. 

Is there a reason Dart can't support this?

MarioAlexandruDan:
> The PR description and ticket mention the implementation of a `ChunkedHelperOptions` object. The purpose of this object is to not further extend method signatures in the future, when we might want to add more configurable params. This PR only threads the maxRetries param through methods.
> 
> Is there a reason Dart can't support this?

Sorry, haven't finished the PR and i put it by mistake in ready to review.

clients/algoliasearch-client-dart/packages/client_search/lib/src/extension/transformation.dart
```diff
@@ -119,11 +125,13 @@ extension Transformation on SearchClient {
     required String indexName,
     required Iterable<Map<String, dynamic>> objects,
     int batchSize = 1000,
+    ChunkedHelperOptions? chunkedOptions,
```
Fluf22: Is it allowed to put a new prop in the middle of a list of args like that without a bc?

clients/algoliasearch-client-dart/packages/client_search/lib/src/extension/wait_task.dart
```diff
@@ -47,7 +47,6 @@ extension WaitTask on SearchClient {
   ///  Wait on an API key creation operation.
   Future<void> waitKeyCreation({
     required String key,
-    int maxRetries = 50,
```
Fluf22, line 50: Same, this is a bc for customers that already implemented the method

clients/algoliasearch-client-dart/packages/client_search/lib/src/extension/transformation.dart
MarioAlexandruDan: In Dart yes, the order in the declaration is purely cosmetic, but i will add it before requestOption. (requestOption is the last parameter in every method)

clients/algoliasearch-client-dart/packages/client_search/lib/src/extension/wait_task.dart
MarioAlexandruDan: Yes

MarioAlexandruDan: Added a Deprecarted warning

Fluf22 (approved):

PR #6444, opened 2026-05-21
fix(specs): use correct map format for facets examples

MarioAlexandruDan:
## 🧭 What and Why

Replace the list syntax (- key: ...) with the proper OpenAPI map syntax so each entry is treated as a named Example Object with a summary label and an actual value, rather than an object whose key happens to be the example name.

🎟 JIRA Ticket [API-359]:

[API-359]: https://algolia.atlassian.net/browse/API-359?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


specs/composition/common/params/Composition.yml
```diff
@@ -13,6 +13,13 @@ facets:
     For more information, see [facets](https://www.algolia.com/doc/guides/managing-results/refine-results/faceting/#contextual-facet-values-and-counts) and [disjunctive faceting for Smart Groups](https://www.algolia.com/doc/guides/managing-results/compositions/search-based-groups#facets-including-disjunctive-faceting).
   default: []
   example: ['category', 'disjunctive(brand)', 'price']
+  examples:
```
ClaraMuller, line 16: Is it expected to have the `example` + `examples` key ?

ClaraMuller: Can you add
```yaml
mix:
    summary: Retrieve all facets and specify which one are disjunctive
    value: ['*', 'disjunctive(brand)']
```

MarioAlexandruDan: > Is it expected to have the `example` + `examples` key ?

Yes, because they serve two different consumers:
- The code sample generator
- The docs renderer.

We might be able in format.ts to update it to fall back to the first examples entry's value when example is absent. (if you want I can explore the idea further)

ClaraMuller: Oh I see.
I will let you decide what is the more convenient for you 👍

ClaraMuller (dismissed):

Fluf22:
@MarioAlexandruDan sadly, this prop shape (map of string -> object) is not compatible with our doc system:
image

What we need to do is expose multiple examples as a list. It must be an array (so for `params.facets` you’d have an array of an array of strings):
image

Then I believe our linter will refuse this format, and that's what we need to fix in this PR: accepting an array for the `examples` prop (iirc)

Fluf22 (approved):
This looks good!

eric-zaharia (approved):

ClaraMuller (approved):

PR #6404, opened 2026-05-13
feat(python): expose maximum number of retries

eric-zaharia:
## 🧭 What and Why

🎟 JIRA Ticket: [API-411](https://algolia.atlassian.net/browse/API-411)

Mirrors the C# work in #6318 and the PHP work in #6401 for the Python client: makes `max_retries` configurable on every chunked / long-polling helper so callers aren't stuck with the default ceiling when they need a longer (or shorter) one.

### Changes

- New `algoliasearch.http.chunked_helper_options.ChunkedHelperOptions` dataclass (`max_retries: int = 100`). Name matches the C# / PHP / JS types so the four clients stay aligned; designed to grow so future shared helper config can be added without ballooning each helper's signature.
- Added an optional trailing `chunked_options: Optional[ChunkedHelperOptions] = None` parameter to:
  - `chunked_batch`, `save_objects`, `delete_objects`, `partial_update_objects`, `replace_all_objects` (search)
  - `chunked_push` (ingestion)
  - `save_objects_with_transformation`, `partial_update_objects_with_transformation`, `replace_all_objects_with_transformation`
- Threads the value through to every internal `wait_for_task` and to the `chunked_push` poll loop.
- **Default retry budget bumped 50 → 100** across the board: ingestion `chunked_push`, the search chunked helpers (`wait_for_tasks=True` path), and the `wait_for_task` / `wait_for_app_task` / `wait_for_api_key` primitives. Brings Python in line with C# / PHP / JS, which already use 100. Flagging here so release notes don't miss it.
- Harmonized the poll-exhausted error message in `chunked_push` to match sibling clients.

### Usage

```python
await client.replace_all_objects(
    index_name=index_name,
    objects=objects,
    chunked_options=ChunkedHelperOptions(max_retries=200),
)
```

Existing call sites are unaffected — the new parameter is optional and trailing.

## 🧪 Test

- `yarn cli cts run python --no-e2e` — exit 0, no failures.

[API-411]: https://algolia.atlassian.net/browse/API-411

PR #6458, opened 2026-05-25
feat(kotlin): expose maximum number of retries

eric-zaharia:
## 🧭 What and Why

🎟 JIRA Ticket: [API-418](https://algolia.atlassian.net/browse/API-418)

Mirrors #6318 (csharp — original), #6353 (javascript), #6401 (php), #6404 (python), and #6406 (go) for the kotlin client: makes `maxRetries` configurable on every chunked / long-polling helper through a new `ChunkedHelperOptions` data class, and bumps the default poll budget on the wait-family helpers 50 → 100 so weekly `replaceAllObjects` reindexes stop hitting the previous ceiling.

### Changes included:

- New `ChunkedHelperOptions` data class in `com.algolia.client.extensions`, plus a top-level `DEFAULT_MAX_RETRIES = 100` constant.
- Optional trailing `chunkedOptions: ChunkedHelperOptions? = null` added to `chunkedBatch`, `saveObjects`, `deleteObjects`, `partialUpdateObjects`, `replaceAllObjects`, `chunkedPush`, `saveObjectsWithTransformation`, `partialUpdateObjectsWithTransformation`, and `replaceAllObjectsWithTransformation`. The value threads into every inner `waitForTask` and ingestion event-poll call.
- Default `maxRetries` bumped 50 → 100 on the wait-family helpers (`waitForTask`, `waitForAppTask`, `waitForApiKey`, `waitKeyCreation`, `waitKeyUpdate`, `waitKeyDelete`, and the deprecated `waitTask` / `waitAppTask`). Practical worst-case wait grows ~250s → ~500s (still capped by `maxDelay = 5.seconds`).
- `IngestionClient.chunkedPush` no longer hard-codes `maxRetries = 50` in its event-poll loop — it uses `chunkedOptions?.maxRetries ?: DEFAULT_MAX_RETRIES`, matching the sibling clients.
- Harmonized the poll-exhausted exception message in `AlgoliaWaitException`: now reads "Stopped waiting for the task after N retries. This does not mean the operation failed; it may still complete. If you need to keep polling, retry with a higher maxRetries."

### Usage

```kotlin
import com.algolia.client.extensions.ChunkedHelperOptions
import com.algolia.client.extensions.replaceAllObjects

client.replaceAllObjects(
  indexName = "products",
  objects = records,
  chunkedOptions = ChunkedHelperOptions(maxRetries = 250),
)
```

Existing call sites are unaffected — `chunkedOptions` is optional and trailing.

[API-418]: https://algolia.atlassian.net/browse/API-418?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clients/algoliasearch-client-kotlin/client/src/commonMain/kotlin/com/algolia/client/extensions/IngestionClient.kt
```diff
@@ -41,9 +42,11 @@ public suspend fun IngestionClient.chunkedPush(
   batchSize: Int = 1000,
   referenceIndexName: String? = null,
   requestOptions: RequestOptions? = null,
+  chunkedOptions: ChunkedHelperOptions? = null,
```
Fluf22: Unnecessary optionality
We should default to our defined sane defaults

```suggestion
  chunkedOptions: ChunkedHelperOptions = ChunkedHelperOptions(),
```

clients/algoliasearch-client-kotlin/client/src/commonMain/kotlin/com/algolia/client/extensions/IngestionClient.kt
```diff
@@ -41,9 +42,11 @@ public suspend fun IngestionClient.chunkedPush(
...
 ): List<WatchResponse> {
   require(batchSize > 0) { "`batchSize` must be greater than 0" }
 
+  val maxRetries = chunkedOptions?.maxRetries ?: DEFAULT_MAX_RETRIES
```
Fluf22: That way, this becomes

```suggestion
  val maxRetries = chunkedOptions.maxRetries
```

clients/algoliasearch-client-kotlin/client/src/commonMain/kotlin/com/algolia/client/extensions/ChunkedHelperOptions.kt
```diff
@@ -0,0 +1,15 @@
+package com.algolia.client.extensions
+
+/** The default maximum number of retries for chunked helpers polling for task completion. */
+public const val DEFAULT_MAX_RETRIES: Int = 100
```
Fluf22: I wonder if we could have this in a more neutral place, like the client or extensions folder?
It would less look like this is a chunked helper const, but more like it's a max retries one that we use in the chunked helper options

Fluf22 (approved):

PR #6459, opened 2026-05-25
feat(scala): expose maximum number of retries

eric-zaharia:
## 🧭 What and Why

🎟 JIRA Ticket: [API-416](https://algolia.atlassian.net/browse/API-416)

Port of "configurable maxRetries on chunked helpers" to scala. Mirrors:
- #6318 (csharp — original)
- #6353 (javascript)
- #6401 (php)
- #6404 (python)
- #6406 (go)

Makes `maxRetries` configurable on every chunked / long-polling helper so callers aren't stuck with the default ceiling when they need a longer (or shorter) one. Default budget bumped 50 → 100 across the wait-family helpers and in `chunkedPush`'s internal poll loop (previously hardcoded to 50); callers who never set `maxRetries` will now wait up to ~2× longer before the helper gives up (worst case ~250s → ~500s, with `DEFAULT_DELAY = retries * 200ms` capped at `5000ms`). This matches the global bump applied in the sibling clients.

### Changes included:

- New `ChunkedHelperOptions` case class in `algoliasearch.extension`, with companion `DefaultMaxRetries = 100`.
- Optional trailing `chunkedOptions: Option[ChunkedHelperOptions] = None` added to `chunkedBatch`, `saveObjects`, `deleteObjects`, `partialUpdateObjects`, `replaceAllObjects`, `saveObjectsWithTransformation`, `partialUpdateObjectsWithTransformation`, `replaceAllObjectsWithTransformation`, and `IngestionClient.chunkedPush`.
- Default `maxRetries` bumped 50 → 100 on the wait-family helpers (`waitForTask`, `waitForAppTask`, `waitForApiKey`, …) and threaded into `pollEvent` (previously hardcoded to 50).
- Harmonized poll-exhausted exception message in `AlgoliaWaitException`.

## 🧪 Test

CI regenerates `clients/`, `tests/output/`, and `specs/bundled/` from these source edits and runs the CTS.

[API-416]: https://algolia.atlassian.net/browse/API-416?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clients/algoliasearch-client-scala/src/main/scala/algoliasearch/extension/package.scala
```diff
@@ -304,8 +308,10 @@ package object extension {
         action: Action = Action.AddObject,
         waitForTasks: Boolean,
         batchSize: Int = 1000,
-        requestOptions: Option[RequestOptions] = None
+        requestOptions: Option[RequestOptions] = None,
+        chunkedOptions: Option[ChunkedHelperOptions] = None
     )(implicit ec: ExecutionContext): Future[Seq[BatchResponse]] = {
+      val maxRetries = chunkedOptions.map(_.maxRetries).getOrElse(DefaultMaxRetries)
```
Fluf22: I have the same comment to make than on the ruby PR, but it adds a lot of overhead in Scala...
I think it'd be more straightforward to just use a non-optional type here, since you're already defaulting the values
You can just do:

```suggestion
        chunkedOptions: ChunkedHelperOptions = ChunkedHelperOptions()
)(implicit ec: ExecutionContext): Future[Seq[BatchResponse]] = {
```

clients/algoliasearch-client-scala/src/main/scala/algoliasearch/extension/package.scala
```diff
@@ -325,7 +331,11 @@ package object extension {
       if (waitForTasks)
         futureResponses.flatMap(rs =>
           Future
-            .sequence(rs.map(r => client.waitForTask(indexName, r.taskID, requestOptions = requestOptions)))
+            .sequence(
+              rs.map(r =>
+                client.waitForTask(indexName, r.taskID, maxRetries = maxRetries, requestOptions = requestOptions)
```
Fluf22: And then:

```suggestion
                client.waitForTask(indexName, r.taskID, maxRetries = chunkedOptions.maxRetries, requestOptions = requestOptions)
```

Fluf22 (commented):
I think this time it's a little more blocking than in ruby

You can do that for all call sites

Fluf22 (approved):

PR #6461, opened 2026-05-25
feat(ruby): expose maximum number of retries

eric-zaharia:
## 🧭 What and Why

🎟 JIRA Ticket: [API-417](https://algolia.atlassian.net/browse/API-417)

Mirrors the C# work in #6318, the JavaScript work in #6353, the Go work in #6406, and the PHP work in #6401 for the Ruby client: makes `max_retries` configurable on every chunked / long-polling helper so callers aren't stuck with the default ceiling when they need a longer (or shorter) one.

### Changes included:

- New `Algolia::ChunkedHelperOptions` class exposing `max_retries` (defaults to `ChunkedHelperOptions::DEFAULT_MAX_RETRIES`). Constructor takes a single kwarg (`new(max_retries: 200)`); designed to grow so future shared helper config can be added without ballooning each helper's signature. Name matches the C# / JS / PHP `ChunkedHelperOptions` class so the clients stay aligned.
- Added an optional trailing positional `chunked_options = nil` parameter to `chunked_batch`, `save_objects`, `delete_objects`, `partial_update_objects`, `replace_all_objects`, the three `*_with_transformation` variants, and ingestion `chunked_push`. Threads `max_retries` through to every nested `wait_for_task` and into `chunked_push`'s event-poll loop.
- **Default polling budget bumped `50` → `100`** on `wait_for_task`, `wait_for_app_task`, `wait_for_api_key`, and `chunked_push`'s inner loop. Matches the global bump in C# #6318, Go #6406, and PHP #6401.

### Usage

```ruby
client.save_objects(
  index_name, objects,
  false, 1000, {},
  Algolia::ChunkedHelperOptions.new(max_retries: 200)
)

client.wait_for_task(index_name, task_id, 200)
```

Existing call sites are unaffected — the new parameter is optional and trailing.

[API-417]: https://algolia.atlassian.net/browse/API-417?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clients/algoliasearch-client-ruby/lib/algolia/chunked_helper_options.rb
```diff
@@ -0,0 +1,12 @@
...
+  # Optional configuration for chunked helpers that batch records and poll for task completion.
+  class ChunkedHelperOptions
+    DEFAULT_MAX_RETRIES = 100
+
+    attr_reader :max_retries
+
+    def initialize(max_retries: nil)
+      @max_retries = max_retries
+    end
+  end
```
Fluf22, line 14:
```suggestion
  class ChunkedHelperOptions
    DEFAULT_MAX_RETRIES = 100
    attr_reader :max_retries

    def initialize(max_retries: DEFAULT_MAX_RETRIES)
      @max_retries = max_retries
    end

    def self.resolve(options)
      options || new
    end
  end
```

clients/algoliasearch-client-ruby/lib/algolia.rb
```diff
@@ -5,6 +5,7 @@
 # Common files
 require "algolia/api_client"
 require "algolia/api_error"
+require "algolia/chunked_helper_options"
```
Fluf22, line 6: This shouldn't be here but in the template file I believe?

```suggestion
```

templates/ruby/ingestion_helpers.mustache
```diff
@@ -9,13 +9,13 @@
...
 # @return [Array<Ingestion::WatchResponse>]
-def chunked_push(index_name, objects, action = Ingestion::Action::ADD_OBJECT, wait_for_tasks = false, batch_size = 1000, reference_index_name = nil, request_options = {})
+def chunked_push(index_name, objects, action = Ingestion::Action::ADD_OBJECT, wait_for_tasks = false, batch_size = 1000, reference_index_name = nil, request_options = {}, chunked_options = nil)
   responses = []
   offset = 0
   wait_batch_size = batch_size / 10
   wait_batch_size = batch_size if wait_batch_size < 1
   total_batches = (objects.length.to_f / batch_size).ceil
-  max_retries = 50
+  max_retries = chunked_options&.max_retries || Algolia::ChunkedHelperOptions::DEFAULT_MAX_RETRIES
```
Fluf22: And then, with the changes in the class, we can just do:

```suggestion
  opts = Algolia::ChunkedHelperOptions.resolve(chunked_options)
```

templates/ruby/ingestion_helpers.mustache
```diff
@@ -42,7 +42,7 @@ def chunked_push(index_name, objects, action = Ingestion::Action::ADD_OBJECT, wa
             raise e unless e.code == 404
           end
           retries += 1
-          raise ApiError, "The maximum number of retries exceeded. (#{max_retries})" if retries >= max_retries
+          raise ApiError, "Stopped waiting for the task after #{max_retries} retries. This does not mean the operation failed; it may still complete. If you need to keep polling, retry with a higher max_retries." if retries >= max_retries
```
Fluf22: And

```suggestion
          raise ApiError, "Stopped waiting for the task after #{opts.max_retries} retries. This does not mean the operation failed; it may still complete. If you need to keep polling, retry with a higher max_retries." if retries >= opts.max_retries
```

Fluf22 (dismissed):
Two non-blocking comments
The second one is replicable on all 4 call sites if you want to implement it (not required)

clients/algoliasearch-client-ruby/lib/algolia.rb
eric-zaharia: `gem.mustache` seems to be excluded from generation (see `AlgoliaRubyGenerator.java` line 58). Should we remove the header instead?

Fluf22: Ah indeed
But is it in the exclude list?

eric-zaharia: It's not explicitly named in the config but it's implied by having only `clients/algoliasearch-client-ruby/lib/algolia/` which matches files under lib/algolia/. `algolia.rb` is in lib, so there is no need to exclude it explicitly.

With that being said, we can add it if you believe it reduces confusion in the future.

Fluf22: Ah yeah, if it’s under the glob pattern match, it’s okay 
Then you can remove the header just to be clear

Fluf22 (approved):

PR #6466, opened 2026-05-26
feat(swift): add TransformationOptions and WithTransformation helpers

MarioAlexandruDan:
## 🧭 What and Why

Part of the effort to standardize \`transformationOptions\` across all 11 language SDKs. Swift currently has no transformation region support at all.

Adds \`TransformationOptions\` configuration and three \`*WithTransformation\` helper methods to \`SearchClient\` that push records through the Algolia Ingestion pipeline before indexing.

### Changes included:

- **\`TransformationOptions\` struct** — flat struct accepting \`region\` (required) and optional ingestion transporter overrides: \`readTimeout\`, \`writeTimeout\`, \`hosts\`, \`compression\`, \`defaultHeaders\`
- **\`SearchClientConfiguration\`** — new \`transformationOptions: TransformationOptions?\` parameter; when set, an \`IngestionClient\` is eagerly created using Ingestion API defaults (25 s timeouts, region-derived hosts, no compression)
- **\`SearchClient.withTransformation(appID:apiKey:transformationOptions:)\`** — throwing static factory for the simple construction path
- **\`SearchClient.saveObjectsWithTransformation\`** — chunks objects and pushes via the Ingestion pipeline with \`addObject\` action
- **\`SearchClient.partialUpdateObjectsWithTransformation\`** — same as above with \`partialUpdateObject\` / \`partialUpdateObjectNoCreate\` action
- **\`SearchClient.replaceAllObjectsWithTransformation\`** — zero-downtime index swap through the Ingestion pipeline (copy → push → wait → copy → wait → move)
- **\`Package.mustache\`** — \`AlgoliaSearch\` target now depends on \`AlgoliaIngestion\`
- **CTS** — Swift added to all three transformation helper specs and to the \`runCts.ts\` assertion counts; \`createClient.mustache\` updated to build a \`SearchClientConfiguration\` with both \`hosts\` and \`transformationOptions\`

## 🧪 Test

CI — CTS covers all three helpers end-to-end against the mock push server.


clients/algoliasearch-client-swift/Sources/Search/Extra/SearchClientTransformationExtension.swift
```diff
@@ -0,0 +1,273 @@
...
+    /// - parameter indexName: The index name to push objects to.
+    /// - parameter objects: The objects to push. Each must include an `objectID` key.
+    /// - parameter action: The ingestion action to apply.
+    /// - parameter waitForTasks: Whether to wait for each batch to finish before moving to the next.
+    /// - parameter batchSize: Number of records per push call. Defaults to 1000.
+    /// - parameter referenceIndexName: Source index whose transformation is applied to `indexName`.
+    /// - parameter requestOptions: Optional per-request options.
+    /// - returns: `[IngestionWatchResponse]`
+    @discardableResult
+    func chunkedPush(
```
Fluf22: I don't remember if we made that public in the other languages
If there is no reason to, then it should be at least internal or private

clients/algoliasearch-client-swift/Sources/Search/Extra/SearchClientTransformationExtension.swift
```diff
@@ -0,0 +1,273 @@
...
+//  before calling these methods.
+//
+//  See https://www.algolia.com/doc/libraries/sdk/methods/ingestion
+
+#if canImport(AlgoliaCore)
+    import AlgoliaCore
+#endif
+#if canImport(AlgoliaIngestion)
+    import AlgoliaIngestion
+#endif
```
Fluf22, line 15: This is required here, right?

```suggestion
import AlgoliaIngestion
```

clients/algoliasearch-client-swift/Sources/Search/Extra/SearchClientTransformationExtension.swift
```diff
@@ -0,0 +1,273 @@
...
+//
+//  Helpers that push records through the Algolia Ingestion pipeline before indexing.
+//  TransformationOptions must be set via SearchClientConfiguration(transformationOptions:)
+//  before calling these methods.
+//
+//  See https://www.algolia.com/doc/libraries/sdk/methods/ingestion
+
+#if canImport(AlgoliaCore)
+    import AlgoliaCore
+#endif
```
Fluf22, line 12: Is this required in the code below?

clients/algoliasearch-client-swift/Sources/Search/Extra/SearchClientTransformationExtension.swift
```diff
@@ -0,0 +1,273 @@
...
+            )
+            try await waitForTask(
+                indexName: tmpIndexName,
+                taskID: moveOperationResponse.taskID,
+                requestOptions: requestOptions
+            )
+
+            let watchResponses = try ingestionWatchResponses.map { r -> SearchWatchResponse in
+                let data = try JSONEncoder().encode(r)
+                return try JSONDecoder().decode(SearchWatchResponse.self, from: data)
```
Fluf22, line 255: ~Please map the fields manually between both models, this is unnecessary overhead~

EDIT: please disregard this comment

clients/algoliasearch-client-swift/Sources/Search/Extra/SearchClientTransformationExtension.swift
```diff
@@ -0,0 +1,273 @@
...
+        } catch let AlgoliaError.httpError(error) where error.statusCode == 404 {
+            retryCount += 1
+            let delayMs = min((retryCount + 1) * 1500, 5000)
+            try await Task.sleep(nanoseconds: UInt64(delayMs) * 1_000_000)
+        }
+    }
+    throw AlgoliaError.runtimeError(
+        "The maximum number of retries exceeded. (\(maxRetries)/\(maxRetries))"
+    )
+}
```
Fluf22, line 311: Can you reuse the `createIterable` pattern for this loop instead of doing it manually please?
Also, we can't use Task because we must support macOS and linux, so we have a BridgedTask class especially for this use case: lookup the pattern in the codebase, it's already here

templates/swift/api.mustache
```diff
@@ -25,7 +33,46 @@ import Foundation
...
+        {{#isSearchClient}}
+        if let opts = configuration.transformationOptions {
+            do {
+                self._ingestionClient = try {{classname}}.makeIngestionClient(
+                    appID: configuration.appID,
+                    apiKey: configuration.apiKey,
+                    options: opts
+                )
+            } catch {
+                fatalError("Invalid TransformationOptions: \(error.localizedDescription)")
```
Fluf22: This is a no-go, we can't have `fatalError` leaking in customer code: this would crash their app
Please handle the error gracefully

clients/algoliasearch-client-swift/Sources/Search/Extra/TransformationOptions.swift
Fluf22, line 1: Could you explore reworking the Ingestion Client options model if you can do it without introducing a BC?
I believe it would be cleaner to have a base model we can extend to be the original IngestionClient options, and that we can also extend (or take as-is) to be the transformation options model

templates/swift/tests/client/createClient.mustache
```diff
@@ -1,3 +1,13 @@
...
+    hosts: [{{#transformationCustomHosts}}RetryableHost(url: URL(string: "http://" + (ProcessInfo.processInfo.environment["CI"] == "true" ? "localhost" : "host.docker.internal") + ":{{port}}")!){{^-last}}, {{/-last}}{{/transformationCustomHosts}}]{{/hasTransformationCustomHosts}}
+)
+let configuration = try {{client}}Configuration(appID: "{{parametersWithDataTypeMap.appId.value}}", apiKey: "{{parametersWithDataTypeMap.apiKey.value}}"{{#hasRegionalHost}}, region: {{#parametersWithDataTypeMap.region}}Region(rawValue: "{{parametersWithDataTypeMap.region.value}}"){{/parametersWithDataTypeMap.region}}{{^parametersWithDataTypeMap.region}}nil{{/parametersWithDataTypeMap.region}}{{/hasRegionalHost}}{{#hasCustomHosts}}, hosts: [{{#customHosts}}RetryableHost(url: URL(string: "http://" + (ProcessInfo.processInfo.environment["CI"] == "true" ? "localhost" : "host.docker.internal") + ":{{port}}")!){{^-last}}, {{/-last}}{{/customHosts}}]{{/hasCustomHosts}}{{#gzipEncoding}}, compression: .gzip{{/gzipEncoding}}, transformationOptions: transformationOptions)
+{{^autoCreateClient}}let client = {{/autoCreateClient}}try {{client}}(configuration: configuration)
```
Fluf22: If you skip using the EchoRequestBuilder, I believe you'll end up sending real requests instead of just printing them in a closed local loop

Fluf22 (commented):
The PR body still mentions `SearchClient.withTransformation` while it was removed from this PR

clients/algoliasearch-client-swift/Sources/Search/Extra/SearchClientTransformationExtension.swift
Fluf22: I was wrong in my comment above: let's keep the serialization round-trip. It's less efficient, but ensures data shape stability in the future.

clients/algoliasearch-client-swift/Sources/Search/Extra/SearchClientTransformationExtension.swift
MarioAlexandruDan: You are right

clients/algoliasearch-client-swift/Sources/Search/Extra/SearchClientTransformationExtension.swift
MarioAlexandruDan: It has to stay conditional. The CocoaPods build is one AlgoliaSearchClient module where AlgoliaIngestion isn't a separate module, so a plain import would fail with "no such module".

clients/algoliasearch-client-swift/Sources/Search/Extra/SearchClientTransformationExtension.swift
MarioAlexandruDan: Yes (AlgoliaError, AnyCodable, RequestOptions, RetryableHost, IngestionClientOption) all come from it. Kept conditional for the same reason as above.

clients/algoliasearch-client-swift/Sources/Search/Extra/SearchClientTransformationExtension.swift
Fluf22: Sorry 🤦🏻‍♂️ I'm so looking forward to Dec 2nd so we don't have to support Cocoapods anymore

clients/algoliasearch-client-swift/Sources/Search/Extra/SearchClientTransformationExtension.swift
```diff
@@ -0,0 +1,290 @@
...
+            return
+        } catch let AlgoliaError.httpError(error) where error.statusCode == 404 {
+            retryCount += 1
+            let delayMs = min((retryCount + 1) * 1500, 5000)
+            try await BridgedTask.sleep(nanoseconds: UInt64(delayMs) * 1_000_000)
+        }
+    }
+    throw AlgoliaError.runtimeError(
+        "The maximum number of retries exceeded. (\(maxRetries)/\(maxRetries))"
+    )
```
Fluf22, line 310: We can't use createIterable here?

templates/swift/tests/client/createClient.mustache
```diff
@@ -1,3 +1,9 @@
...
+{{#hasTransformationRegion}}
+let transformationOptions = TransformationOptions(
+    region: Region(rawValue: "{{{transformationRegion}}}"){{#hasTransformationCustomHosts}},
+    clientOptions: IngestionClientOptions(hosts: [{{#transformationCustomHosts}}RetryableHost(url: URL(string: "http://" + (ProcessInfo.processInfo.environment["CI"] == "true" ? "localhost" : "host.docker.internal") + ":{{port}}")!){{^-last}}, {{/-last}}{{/transformationCustomHosts}}]){{/hasTransformationCustomHosts}}
```
Fluf22: You're putting a specific client name in a generic template

templates/swift/client_configuration.mustache
```diff
@@ -123,4 +132,21 @@ public struct {{#lambda.client-to-name}}{{{client}}}{{/lambda.client-to-name}}Cl
...
+      region: region,
+      writeTimeout: options.writeTimeout ?? {{#lambda.toSeconds}}{{x-timeouts.server.write}}{{/lambda.toSeconds}},
+      readTimeout: options.readTimeout ?? {{#lambda.toSeconds}}{{x-timeouts.server.read}}{{/lambda.toSeconds}},
+      defaultHeaders: options.defaultHeaders,
+      hosts: options.hosts,
+      compression: options.compression ?? .none
+    )
+  }
+  {{/isIngestionClient}}
 }
```
Fluf22: This is getting too specific
The idea would be to have a {{clientName}}ClientOptions type generated that extends the ClientOptions one that we can also reuse for TransformationOptions
And we need to make sure we don't introduce any BC of course

clients/algoliasearch-client-swift/Sources/Core/Networking/IngestionClientOptions.swift
Fluf22, line 1: As I said in another comment, the idea is not to have a manually written file specific to the ingestion client, but a ClientOptions one that has the base props, and then every client has its own {{clientName}}ClientOptions file generated that it can imports from the template
Other languages use this pattern already

Fluf22 (commented):
Sorry, I wasn't clear enough about the client options model I was thinking about

clients/algoliasearch-client-swift/Sources/Search/Extra/SearchClientTransformationExtension.swift
MarioAlexandruDan: createIterable validates on a result predicate and doesn't catch exceptions, getEvent throws a 404 until the event exists, so reusing it would force try? which swallows non-404 errors and loses the retry-only-on-404 behaviour.

clients/algoliasearch-client-swift/Sources/Core/Networking/IngestionClientOptions.swift
MarioAlexandruDan: Reworked to a single base ClientOptions in Core. 
TransformationOptions(region, clientOptions: ClientOptions?), identical to Kotlin, Java and Scala. 
I checked all 11 SDKs and none generate per-client {{clientName}}ClientOptions; they all use one shared ClientOptions, so I matched that rather than make Swift the only one with per-client types.

clients/algoliasearch-client-swift/Sources/Search/Extra/SearchClientTransformationExtension.swift
Fluf22: Looks like that's what we do in `waitForApiKey`, in the SearchClientExtension.swift file:
```swift
return try await createIterable(
            execute: { _ in
                do {
                    return try await self.getApiKey(key: key, requestOptions: requestOptions)
                } catch let AlgoliaError.httpError(error) {
                    if error.statusCode == 404 {
                        return nil
                    }

                    throw error
                }
            },
            validate: { response in
                switch operation {
                case .add:
                    guard let response else {
                        return false
                    }

                    return response.createdAt > 0
                case .delete:
                    return response == nil
                default:
                    return false
                }
            },
            aggregator: { _ in
                retryCount += 1
            },
            timeout: {
                timeout(retryCount)
            },
            error: IterableError(
                validate: { _ in
                    retryCount >= maxRetries
                },
                message: { _ in
                    "The maximum number of retries exceeded. (\(retryCount)/\(maxRetries))"
                }
            )
        )
```

clients/algoliasearch-client-swift/Sources/Core/Networking/IngestionClientOptions.swift
Fluf22: Go does that `algolia/search/configuration.go`
But you're right the Java/Kotlin/etc don't

So you can copy either model

clients/algoliasearch-client-swift/Sources/Search/Extra/SearchClientTransformationExtension.swift
MarioAlexandruDan: waitForIngestionEvent now reuses createIterable like waitForApiKey

templates/swift/api.mustache
```diff
@@ -25,6 +33,9 @@ import Foundation
 
     public convenience init(configuration: {{classname}}Configuration) {
         self.init(configuration: configuration, transporter: Transporter(configuration: configuration))
+        {{! The ingestion transporter is built lazily on first `*WithTransformation` call }}
+        {{! (see SearchClientTransformationExtension) so invalid options surface as a thrown }}
+        {{! error at the call site instead of crashing the client initializer. }}
```
Fluf22: No need for code related comments in mustache template

clients/algoliasearch-client-swift/Sources/Core/Networking/ClientOptions.swift
```diff
@@ -0,0 +1,46 @@
...
+//
+//  Overridable client options shared across the SDK. Every field is optional; an unset field
+//  keeps the calling client's default. Used as-is by `TransformationOptions` to override the
+//  ingestion transporter defaults (25 s timeouts, region-derived hosts, no compression).
+//
+//  See https://www.algolia.com/doc/libraries/sdk/methods/ingestion
+
+import Foundation
+
+public struct ClientOptions {
```
Fluf22: Cool, now that we have this base client options struct, can you make the {{classname}}Configuration template extend it, if that's not bc please? 😁

MarioAlexandruDan: Tried a few ways to extend …Configuration from ClientOptions and they all either don't work in Swift or are source-breaking:
- Inheritance: both are structs, so no extends.
- Shared protocol: the fields are type-incompatible: ClientOptions is all-optional overrides (readTimeout: TimeInterval?, "unset = keep default") while the config is resolved/required (readTimeout: TimeInterval, always set). One protocol can't cover both.
- Composition (config holds a ClientOptions): breaking, since config.readTimeout/config.hosts/… are public stored properties today.

The shared contract is already the BaseConfiguration protocol that the config conforms to, and ClientOptions stays as the dedicated optional-overrides type for TransformationOptions. So I'd keep them separate.

Fluf22 (approved):
Okay, let's go

PR #6482, opened 2026-06-08
refactor(scala): split package.scala into separate traits

eric-zaharia:
## 🧭 What and Why

🎟 JIRA Ticket: [API-499](https://algolia.atlassian.net/browse/API-499)

Splits the ~1060-line `extension/package.scala` into one file per extension group for easier maintenance. No behavior change — same package, same public extension methods, same implicit scope. Hand-written files only.

The ticket asked for "imports and re-exports only," but a literal Scala 3 `export` doesn't cross-compile (fails on the 2.13 leg). So each `implicit class` group moves into its own trait, and `package object extension` mixes them in. Callers still reach everything via `import algoliasearch.extension._`.

### Changes included:

- `SecuredApiKeyExtensions.scala` — `trait SecuredApiKeyExtensions`.
- `IngestionExtensions.scala` — `trait IngestionExtensions`.
- `SearchExtensions.scala` — `trait SearchExtensions`, extending the other two so `chunkedPush` / `toUrlEncoded` stay in implicit scope.
- `package.scala` reduced to a `package object` mixing in all three traits.

Each trait's name matches its filename (one top-level template per file); the public `implicit class` names (`SearchClientExtensions`, `IngestionClientExtensions`, `SecuredApiKeyRestrictionsExtension`) are unchanged.

## 🧪 Test

- `sbt +compile` green on Scala 2.13.17 and 3.8.2.
- CTS passes; the by-name `import algoliasearch.extension.SearchClientExtensions` used by generated tests still resolves.

[API-499]: https://algolia.atlassian.net/browse/API-499?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Fluf22 (approved):
This is a good enough first split
I believe that when we will have more time to please customers using the Scala client and reading the code, we could even split it further, grouping related methods in implicit class
This will do the job for now

PR #6462, opened 2026-05-26
feat(kotlin): add TransformationOptions for ingestion transporter

eric-zaharia:
## 🧭 What and Why

🎟 JIRA Ticket: [API-376](https://algolia.atlassian.net/browse/API-376)

Mirrors #6349 (php), #6312 (scala), #6293 (go), #6255 (python), and #6252 (javascript) for the kotlin client: introduces `TransformationOptions` as the configurable surface for the ingestion transporter used by the `*WithTransformation` write helpers (`saveObjectsWithTransformation`, `partialUpdateObjectsWithTransformation`, `replaceAllObjectsWithTransformation`). When installed, an `IngestionClient` is built using Ingestion API defaults (25 s timeouts, region-derived hosts); pass a `ClientOptions` to override specific defaults, and the parent `SearchClient` config never leaks into the ingestion transporter.

The Kotlin port follows the Scala/Java shape: a hand-written `TransformationOptions(region, clientOptions?)` data class, an `internal` `ingestionTransporter` field hosted on the generated `SearchClient` via a new `{{#isSearchClient}}` template block, a `companion object` factory, an instance setter, and `suspend` extension helpers. Kotlin never had `setTransformationRegion`, so this is purely additive — no deprecations, no breaking changes.

### Usage

```kotlin
import com.algolia.client.api.SearchClient
import com.algolia.client.configuration.ClientOptions
import com.algolia.client.configuration.TransformationOptions
import kotlin.time.Duration.Companion.seconds

// 1. Factory — preferred, mirrors Java's `SearchClient.withTransformation`
val client = SearchClient.withTransformation(
  appId = "APP_ID",
  apiKey = "API_KEY",
  transformationOptions = TransformationOptions(region = "us"),
)

// 2. Factory with extra overrides on the ingestion transporter (timeouts, hosts, headers).
//    Pass a ClientOptions to override the Ingestion API defaults; only the fields you set replace
//    them. The shape matches Java/Scala — TransformationOptions(region, clientOptions?).
val tunedClient = SearchClient.withTransformation(
  appId = "APP_ID",
  apiKey = "API_KEY",
  transformationOptions = TransformationOptions(
    region = "eu",
    clientOptions = ClientOptions(
      readTimeout = 50.seconds,
      writeTimeout = 50.seconds,
    ),
  ),
)

// 3. Swap (or first-set) on an already-built client.
val client2 = SearchClient(appId = "APP_ID", apiKey = "API_KEY")
[...]

PR #6483, opened 2026-06-08
guides: add transformationOptions setup guides for the search client

MarioAlexandruDan:
## Summary

Adds generated guide templates for the new `transformationOptions` client configuration (final docs ticket of the transformationOptions standardization effort, API-369..379).

Two guides per language, reusing a new `setUpTransformationOptions` CTS guide entry (runs `saveObjectsWithTransformation` through the Push connector):

- **setUpTransformationOptions** — basic usage (region only)
- **setUpTransformationOptionsWithOverrides** — region + connect/read timeout (and host) overrides

Templates are present for all 11 languages. They are consumed by docs-new (push.mdx / customize.mdx) via the generated guides pipeline.

## Status — green ✅

`main` has been merged in (Swift #6466, Dart #6321, Kotlin #6462 are all on `main`), and **all 11 languages generate and build the guides successfully** — every `client <lang>` job, including the "Generate code guides" and "Build the guides to check validity" steps, plus `codegen` and `check_green`, pass. The temporary `FOLLOWUP-swift-dart-transformation-guides.md` planning note has been removed.

## Fixes applied while greening CI

The first CI run (before `main` was merged, and before per-language review) surfaced guide build failures. All resolved:

- **Swift & Dart** — stale only: `TransformationOptions` landed on `main` *after* that run. Resolved by merging `main`.
- **Java** — the generated `saveObjectsWithTransformation` snippet uses `Arrays.asList(new HashMap(){…})` and a raw `List`; added the `java.util.{Arrays,HashMap,List}` imports.
- **Kotlin** — the snippet builds records with `buildJsonObject`/`JsonPrimitive`; added the `kotlinx.serialization.json` imports.
- **Scala** — the guide used `object Main`, colliding with `saveObjectsMovies`'s `object Main` (Scala guides compile together); renamed to unique per-guide objects.
- **Swift overrides** — `TransformationOptions` has no `clientOptions`/`ClientOptions`; switched to the real flat override fields (`readTimeout`/`writeTimeout`).
- **JavaScript overrides** — the `Timeouts` type requires `write`; added it alongside `connect`/`read`.
- **Dart** — the guides now use the shared `{{dynamicSnippet}}` (like the other 10 languages) instead of hand-writing the `saveObjectsWithTransformation` call, so the body is generated from the CTS entry rather than duplicating its test data.

## Notes

- The per-language examples in the ticket are inaccurate for some clients; the templates follow the real, shipped APIs (e.g. C# `SearchClient.WithTransformation(...)`, Dart `transformationOptions` as a client-constructor param).
- Override timeout units differ by client but the examples are semantically equivalent (5s connect / 30s read) — Python/Ruby/JS use milliseconds, PHP uses seconds, the typed-`Duration` clients use 5s/30s.
- Generated outputs (docs/guides, bundled snippets) are produced by the regeneration pipeline in a separate step.


Fluf22 (approved):
Looks good
Curious to see how it renders in the docs

PR #6487, opened 2026-06-09
test(clients): test chunked push polling mechanism

eric-zaharia:
## 🧭 What and Why

Adds a CTS client test that locks in the `chunkedPush` task-polling contract: with `waitForTasks: true`, every pushed task must be polled via `getEvent` exactly once. The bug this guards against — the poll cursor over-advancing and silently skipping tasks — is invisible to client-side assertions, because the returned `WatchResponse` list is identical whether or not polling happened. The test therefore asserts server-side, by counting the requests the client actually makes. Full cross-language analysis: [Confluence: Chunked Push Task-Polling Bug (waitForTasks)](https://algolia.atlassian.net/wiki/spaces/APIC/pages/7227441368/Chunked+Push+Task-Polling+Bug+waitForTasks).

### Changes included:

- **`chunkedPushWait` mock server** (`scripts/cts/testServer/chunkedPushWait.ts`): simulates the ingestion `push` and `getEvent` endpoints on port 6692, counting both per language. Each push returns a distinct `eventID`, and `getEvent` answers `status: succeeded`, so a correct client polls exactly once per pushed task.
- **CTS client test** (`tests/CTS/client/search/saveObjectsWithTransformation.json`): drives `saveObjectsWithTransformation` with 25 objects, `batchSize: 10`, `waitForTasks: true` — producing 3 pushes (10/10/5) and a poll window of 1 (`waitBatchSize = batchSize / 10`), the smallest configuration that triggers the cursor bug.
- **Post-run assertion** (`scripts/cts/runCts.ts`): `assertValidChunkedPushWait` requires `pushCount >= 3` and `getEventCount == pushCount` for every participating language, and verifies the participant count itself, so a test that silently stops being generated fails loudly instead of passing vacuously.
- **New `skipLanguages` field on CTS client tests** (`generators/.../ClientTestData.java`, `TestsClient.java`): excludes a test per language at generation time. Go, PHP, and JavaScript are skipped here because they currently fail the polling contract — Go panics on an out-of-range slice, PHP silently under-polls, and JavaScript additionally never forwards `batchSize` through `saveObjectsWithTransformation`, so it needs both defects fixed before it can participate. Dart and Swift don't expose the helper and are excluded automatically.

## 🧪 Test

- An unfiltered run of this branch serves as the cross-language proof: go, php, and javascript fail exactly as documented in the Confluence page, while java, csharp, kotlin, ruby, and scala pass.
- Python participates and passes thanks to the fix in #6486 (this branch's base).
- With the `skipLanguages` filter in place, all client CI jobs are green.

### DIFF
--- a/generators/src/main/java/com/algolia/codegen/cts/tests/ClientTestData.java
+++ b/generators/src/main/java/com/algolia/codegen/cts/tests/ClientTestData.java
@@ -7,6 +7,7 @@ public class ClientTestData {
   public String testName;
   public boolean autoCreateClient = true;
   public List<Step> steps;
+  public List<String> skipLanguages;
 
   public String toString() {
     StringBuilder sb = new StringBuilder();
--- a/generators/src/main/java/com/algolia/codegen/cts/tests/TestsClient.java
+++ b/generators/src/main/java/com/algolia/codegen/cts/tests/TestsClient.java
@@ -70,6 +70,10 @@ public class TestsClient extends TestsGenerator {
       List<Object> tests = new ArrayList<>();
       int testIndex = 0;
       skipTest: for (ClientTestData test : blockEntry.getValue()) {
+        if (test.skipLanguages != null && test.skipLanguages.contains(language)) {
+          System.out.println("Skipping client test " + (test.testName == null ? client : test.testName) + " for language " + language);
+          continue skipTest;
+        }
         try {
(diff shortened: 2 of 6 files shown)


Fluf22 (approved):
👍🏻 
Let's not forget to add the new port to PR #6479

PR #6492, opened 2026-06-10
feat(csharp): add IDisposable support and dispose the ingestion transporter on replacement

MarioAlexandruDan:
## Summary

Calling `SetTransformationOptions` / `SetTransformationRegion` repeatedly on a `SearchClient` silently replaced the internal `IngestionClient` without disposing the previous one, and clients never released their `HttpClient`. This adds `IDisposable` support across the C# clients and matches the Java/Python behavior of closing the previous ingestion transporter on replacement.

- The **concrete client classes** implement `IDisposable` directly (via `templates/csharp/api.mustache`); `Dispose()` disposes `_transport` and, for `SearchClient`, the `_ingestionTransporter`, with the ingestion side wrapped in `try/finally` so the main transport is always released.
- `AlgoliaHttpRequester` and `HttpTransport` implement `IDisposable` and dispose their owned `HttpClient` / requester.
- `SetTransformationOptions` and the deprecated `SetTransformationRegion` dispose the existing `_ingestionTransporter` before replacing it (capture → assign → dispose previous, so a construction failure keeps the working transporter).

## Non-breaking by design

The **public injectable interfaces are left unchanged** — `IHttpRequester` and the generated `I*Client` interfaces do **not** extend `IDisposable`. Disposal happens on the concrete implementations, and at the one boundary that holds an interface-typed reference we dispose via a runtime cast:

```csharp
(_httpClient as IDisposable)?.Dispose();        // HttpTransport -> IHttpRequester
(_ingestionTransporter as IDisposable)?.Dispose(); // client -> IIngestionClient
```

The default `AlgoliaHttpRequester` / `IngestionClient` are `IDisposable`, so the cast always succeeds on the real path, while a **custom `IHttpRequester` or a mocked `I*Client`** that owns no disposable resources is simply skipped — **no existing implementer or mock has to add `Dispose()`**. Adding `IDisposable` to the concrete client classes is purely additive (callers instantiate these, they don't implement them), so `using` / `Dispose()` works for consumers without any breaking change.

> Considered making the interfaces extend `IDisposable` to drop the casts (per review), but that would force every custom `IHttpRequester` implementer and every `I*Client` mock to implement `Dispose()` — a breaking change. We opted to keep the feature non-breaking; the cast is contained to the disposal paths and always resolves for the shipped implementations.

## Acceptance criteria
- [x] `IngestionClient` implements `IDisposable`
- [x] `SetTransformationOptions` disposes the existing `_ingestionTransporter` (if non-null) before replacing it
- [x] `SearchClient` implements `IDisposable` and disposes `_ingestionTransporter` on teardown
- [x] Behaviour consistent with Java (`IngestionClient.close()` on replacement) and Python (transporter closed on replacement)
- [x] No breaking change for custom `IHttpRequester` implementations or `I*Client` mocks

## References
- PR #6292
- Sibling: Kotlin AutoCloseable (PR #6493) — same non-breaking approach


templates/csharp/api.mustache
```diff
@@ -211,6 +211,18 @@ namespace Algolia.Search.Clients;
...
+      /// Disposes the client and its underlying resources (the HTTP transport{{#isSearchClient}} and, if set, the ingestion transporter{{/isSearchClient}}).
+      /// </summary>
+      public void Dispose()
+      {
+        _transport?.Dispose();
+        {{#isSearchClient}}
+        (_ingestionTransporter as IDisposable)?.Dispose();
+        {{/isSearchClient}}
+        GC.SuppressFinalize(this);
+      }
```
Fluf22, line 235: The main transport is the critical resource, we need to invert the disposing order and wrap the ingestionTransporter one in a try/catch to ensure being able to dispose the main transport one
```suggestion
      public void Dispose()
      {
        {{#isSearchClient}}
        try
        {
          (_ingestionTransporter as IDisposable)?.Dispose();
        }
        finally
        {
        {{/isSearchClient}}
        _transport?.Dispose();
        {{#isSearchClient}}
        }
        {{/isSearchClient}}
        GC.SuppressFinalize(this);
      }
```

templates/csharp/api.mustache
```diff
@@ -211,6 +211,18 @@ namespace Algolia.Search.Clients;
...
       }
 
+      /// <summary>
+      /// Disposes the client and its underlying resources (the HTTP transport{{#isSearchClient}} and, if set, the ingestion transporter{{/isSearchClient}}).
+      /// </summary>
+      public void Dispose()
+      {
+        _transport?.Dispose();
+        {{#isSearchClient}}
+        (_ingestionTransporter as IDisposable)?.Dispose();
```
Fluf22: The explicit cast wouldn't be required if the generated interfaces were extending IDisposable

clients/algoliasearch-client-csharp/algoliasearch/Transport/HttpTransport.cs
```diff
@@ -447,4 +447,13 @@ private TimeSpan GetTimeOut(CallType callType, InternalRequestOptions requestOpt
       _ => Defaults.WriteTimeout,
     };
   }
+
+  /// <summary>
+  /// Disposes the underlying HTTP requester, if it is disposable.
+  /// </summary>
+  public void Dispose()
+  {
+    (_httpClient as IDisposable)?.Dispose();
```
Fluf22, line 458: Same, I believe: the explicit cast wouldn't be required if the generated interfaces were extending IDisposable

templates/csharp/api.mustache
```diff
@@ -242,7 +254,9 @@ namespace Algolia.Search.Clients;
             ingestionConfig.DefaultHeaders[kvp.Key] = kvp.Value;
         }
 
+        var previous = _ingestionTransporter;
         _ingestionTransporter = new IngestionClient(ingestionConfig, _loggerFactory);
+        (previous as IDisposable)?.Dispose();
```
Fluf22, line 270: I hope I'm right: the explicit cast wouldn't be required if the generated interfaces were extending IDisposable

templates/csharp/api.mustache
```diff
@@ -275,6 +290,7 @@ namespace Algolia.Search.Clients;
           },
           factory ?? _loggerFactory
         );
+        (previous as IDisposable)?.Dispose();
```
Fluf22, line 304: :mild-panic-intensifies: The explicit cast wouldn't be required if the generated interfaces were extending IDisposable

Fluf22 (commented):
Most comments are about interface not extending IDisposable: I believe this would be a BC for anyone building on them already...
But I don't think there would be a lot of cases? Tell me what you think, and if you go with the BC, please add a note in the PR desc so we don't forget to warn users in the changelog

Fluf22 (approved):

PR #6493, opened 2026-06-10
feat(kotlin): add AutoCloseable support and close the ingestion transporter on replacement

MarioAlexandruDan:
## Summary

Calling `setTransformationOptions` / `withTransformation` repeatedly on a `SearchClient` silently replaced the internal `IngestionClient` without closing the previous one, leaking its ktor `HttpClient` (connection pool, threads, sockets) until GC finalization. This adds `AutoCloseable` support across the Kotlin client family and matches the Java/Scala/Python behavior of closing the previous ingestion transporter on replacement.

- The **generated client classes** implement `kotlin.AutoCloseable` directly (via `templates/kotlin/api.mustache`); `close()` closes the `ingestionTransporter` (for `SearchClient`) inside a `try`, then always closes the requester in the `finally`, so the critical main transport is released even if closing the ingestion transporter throws (mirrors C#'s `Dispose()` ordering).
- `KtorRequester` implements `kotlin.AutoCloseable` and closes the ktor `HttpClient`.
- `setTransformationOptions` captures the previous transporter, installs the new one, then closes the previous (so a construction failure keeps the working transporter).

## Non-breaking by design

The **public injectable `Requester` interface is left unchanged** — it does **not** extend `kotlin.AutoCloseable`. Closing happens on the concrete implementations, and at the one boundary that holds a `Requester`-typed reference we close via a runtime cast:

```kotlin
(requester as? kotlin.AutoCloseable)?.close()
```

The default `KtorRequester` is `AutoCloseable`, so the cast always succeeds on the real path, while a **custom `Requester`** (passed via `ClientOptions.requester`) that owns no closeable resources is simply skipped — **no existing implementer has to add `close()`**. Adding `AutoCloseable` to the generated client classes is purely additive (callers instantiate these, they don't implement them), so `use {}` / `close()` works for consumers without any breaking change.

> Considered making `Requester` extend `kotlin.AutoCloseable` to drop the cast (to match the C# review discussion), but that would force every custom `Requester` implementer to add `close()` — a breaking change. We opted to keep the feature non-breaking; the cast is contained to `close()` and always resolves for the shipped `KtorRequester`.

## Acceptance criteria
- [x] `IngestionClient` implements `AutoCloseable` (multiplatform-safe — `kotlin.AutoCloseable`, not `java.io.Closeable`)
- [x] `SearchClient` implements `AutoCloseable` and closes `ingestionTransporter` on teardown
- [x] `setTransformationOptions` closes the existing `ingestionTransporter` (if non-null) before replacing it
- [x] Behaviour consistent with Java, Scala, and Python equivalents
- [x] No breaking change for custom `Requester` implementations

## Note on the type
The ticket said `kotlin.io.Closeable`, but that type does not exist in the common stdlib (it's JVM-only). The correct multiplatform common-stdlib interface is **`kotlin.AutoCloseable`** (Kotlin 2.x), which is also what ktor 3's `HttpClient` implements. Using `kotlin.io.Closeable` fails to compile in `commonMain`; `AutoCloseable` satisfies the multiplatform-safe intent.

## References
- PR #6462 — Kotlin port of TransformationOptions
- Sibling: C# IDisposable (PR #6492) — same non-breaking approach


Fluf22 (approved):
Let's go

PR #6499, opened 2026-06-11
fix(ci): fail spread step when pushes to client repos fail

Fluf22:
## Problem

The `spreadGeneration` script caught push errors in a try/catch but only called `console.error()` — never propagating the failure. This caused the `push_and_release` job to report **green** even when **all 11 client repo pushes failed** (see [this run](https://github.com/algolia/api-clients-automation/actions/runs/27349026067/job/80809149635)).

The cascade:
1. Every `git push` fails with `GH006: Protected branch update failed`
2. Errors are caught and logged but silently swallowed
3. `PUSHED_LANGUAGES` output is empty → downstream steps no-op
4. `FAILED_RELEASES` is empty → Slack notification is skipped
5. Job reports success ✅

## Fix

**`scripts/ci/codegen/spreadGeneration.ts`**:
- Track `failed[]` languages alongside `pushed[]`
- Call `core.setFailed()` when any language fails to spread
- Output `FAILED_LANGUAGES` for downstream consumption
- `await` the async entry point (was fire-and-forget)

**`.github/workflows/check.yml`**:
- Wire the notify step to also fire on `failure()` or when `FAILED_LANGUAGES` is non-empty

## Note

The root cause of the actual push failures was missing branch protection bypass for the GitHub App on the 11 client repos — that has been fixed separately.

### DIFF
--- a/scripts/ci/codegen/spreadGeneration.ts
+++ b/scripts/ci/codegen/spreadGeneration.ts
@@ -33,6 +33,8 @@ export function cleanUpCommitMessage(commitMessage: string, version: string): st
   return [`${prCommit[1]} ${text.commitEndMessage}`, `${REPO_URL}/pull/${prCommit[2]}`].join('\n\n');
 }
 
+// Failed languages count as non-skipped even if the script failed before checking the diff.
+// The before/after-diff split is only for reporting where the spread failed.
 export type SpreadGenerationResult =
   | { type: 'skipped'; language: Language }
   | { type: 'pushed'; language: Language }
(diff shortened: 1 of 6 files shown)


scripts/ci/codegen/spreadGeneration.ts
```diff
@@ -118,13 +119,19 @@ async function spreadGeneration(): Promise<void> {
...
       console.error(`Release failed for language ${lang}: ${e}`);
+      failed.push(lang);
     }
   }
 
   core.setOutput('PUSHED_LANGUAGES', pushed.join(' '));
+  core.setOutput('FAILED_LANGUAGES', failed.join(' '));
+
+  if (failed.length > 0) {
+    core.setFailed(`Spread failed for: ${failed.join(', ')}`);
```
eric-zaharia: Do we not want to still execute the remaining workflow steps for successful languages?

Fluf22: Great remark

I believe the best of both world would be to fail if ALL languages have fail, because it highlights a greater issue

Wdyt?

eric-zaharia: Sounds good! And else we still run the entire workflow for the successful langs and notify for failures

scripts/ci/codegen/spreadGeneration.ts
```diff
@@ -33,6 +33,13 @@ export function cleanUpCommitMessage(commitMessage: string, version: string): st
   return [`${prCommit[1]} ${text.commitEndMessage}`, `${REPO_URL}/pull/${prCommit[2]}`].join('\n\n');
 }
 
+export function shouldFailSpreadGeneration(
+  failedLanguages: readonly Language[],
+  languages: readonly Language[] = LANGUAGES,
+): boolean {
+  return languages.length > 0 && failedLanguages.length === languages.length;
```
eric-zaharia: I believe if some clients are skipped (no diff) and all non-skipped ones fail this will not trigger

Fluf22: Agreed, good catch

eric-zaharia (approved):
Looks good!

PR #6507, opened 2026-06-12
fix(dart): gate extension.dart export behind hasExtensions

MarioAlexandruDan:
## Problem

`templates/dart/lib.mustache` always emits `export 'src/extension.dart';`, but `extension.dart` is hand-authored (negated from generation in `config/generation.config.mjs`). As a result, **every Dart package without real helpers needs a manually-created blank `extension.dart`** just to keep the generated lib file compiling. Eight packages carried these placeholders, and it's easy to forget when adding a new client.

## Fix

- **`templates/dart/lib.mustache`** — gate the export behind `{{#hasExtensions}}`, so only clients that actually ship helpers emit the line.
- **`generators/.../AlgoliaDartGenerator.java`** — set `hasExtensions: true` only for `search` and `algoliasearch`, the two packages with real hand-authored extensions.
- **Delete the placeholders** — the 7 blank files (`client_abtesting_v3`, `client_composition`, `client_core`, `client_ingestion`, `client_insights`, `client_realtime_personalization`, `client_recommend`) plus the comment-only `client_agent_studio` stub.

## Notes

- `client_search` and `algoliasearch` keep their `extension.dart` (real helpers); all others are dropped.
- `client_core` is hand-authored (excluded from generation) and its lib never exported `extension.dart`, so its placeholder was already orphaned — safe to delete.
- A future client that needs helpers re-creates `extension.dart` and flips `hasExtensions` for its `client` name.
- Generated client outputs are intentionally **not** regenerated here — they land in the separate generated PR per the usual workflow.


generators/src/main/java/com/algolia/codegen/AlgoliaDartGenerator.java
```diff
@@ -114,6 +114,9 @@ public void processOpts() {
     // Search config
     additionalProperties.put("isSearchClient", client.equals("search"));
     additionalProperties.put("packageVersion", effectiveVersion);
+
+    // Only clients shipping hand-authored helpers export `src/extension.dart`.
+    additionalProperties.put("hasExtensions", client.equals("search") || isAlgoliasearchClient);
```
Fluf22, line 119: The ingestion client doesn't have extensions?
I think they are located in the search extension, but it's a mistake
Let's plan on moving them in a future sprint

Fluf22 (approved):

PR #6570, opened 2026-06-15
chore(deps): Upgrade json4s from 4.1.0 to 4.1.1

dragisak:
## 🧭 What and Why

🎟 JIRA Ticket:

### Changes included:

- List changes


Fluf22:
@dragisak thanks for your contribution! I included this change in the dependency update I'm working on (ref [here](https://github.com/algolia/api-clients-automation/pull/6586))
We should release a new version of the client with this update soon

PR #6581, opened 2026-06-17
fix(swift): return generic type from getObject

Fluf22:
## 🧭 What and Why

🎟 JIRA Ticket: N/A — Fixes [algolia/algoliasearch-client-swift#891](https://github.com/algolia/algoliasearch-client-swift/issues/891)

`SearchClient.getObject(indexName:objectID:attributesToRetrieve:requestOptions:)` returned `AnyCodable`, while every other search/get method returns a generic type (`getObjects<T>`, `searchSingleIndex<T>`, …). With `AnyCodable` there is no typed result: callers had to hand-walk an `Any` tree or do a lossy re-encode/re-decode round-trip, so the reporter fell back to `getObjects` to get a faithful typed decode.

### Root cause

`getObject`'s `200` response is an **inline** `type: object` schema flagged `x-is-generic: true`. For inline-return generics, `GenericPropagator` deliberately sets `x-return-is-generic` (not `x-is-generic`) so only templates that explicitly opt in apply the generic. Only the JavaScript template honored `x-return-is-generic`; the Swift template checked `x-is-generic` only, so `getObject` fell back to `AnyCodable`. (Named-model generics like `getObjects` → `GetObjectsResponse<T>` use `x-is-generic`, which Swift already handled — which is why those were generic and `getObject` was not.)

### Changes included (template-only)

For `x-return-is-generic` operations, the Swift template now generates **four** methods, giving a typed API with **zero breaking change**:

```swift
// new, recommended — faithful single-pass decode via transporter.send<T>
open func getObject<T: Codable>(...) async throws -> T
open func getObjectWithHTTPInfo<T: Codable>(...) async throws -> Response<T>

// kept for back-compat, deprecated
@available(*, deprecated, message: "… use the generic getObject …")
open func getObject(...) async throws -> AnyCodable
@available(*, deprecated, message: "… use the generic getObjectWithHTTPInfo …")
open func getObjectWithHTTPInfo(...) async throws -> Response<AnyCodable>
```

- **`templates/swift/api.mustache`** — handle `x-return-is-generic`: emit the generic pair, plus the deprecated `AnyCodable` pair (a faithful copy of the original methods + `@available(deprecated)`).
- **`templates/swift/tests/{requests,e2e,client}/*.mustache`** and **`templates/swift/snippets/method.mustache`** — annotate `isReturnGeneric` call sites (`Response<[String: AnyCodable]>` / `[String: AnyCodable]`) so generated tests and snippets exercise the **generic** overload (no deprecation warnings in generated code, and the snippet demonstrates the typed form).

The deprecated `AnyCodable` overloads are **generated, not hand-written**, because they're the transport leaf — they must call `self.transporter.send(...)`, and `transporter` is `private` (unreachable from a hand-written extension), and a non-generic overload can't delegate to the generic one without infinite recursion.

```swift
// Typed (new, recommended):
let movie: Movie = try await client.getObject(indexName: "movies", objectID: "1")
// Untyped (still compiles, now deprecated — nudges you to annotate):
let any = try await client.getObject(indexName: "movies", objectID: "1") // -> AnyCodable
```

## ✅ Backward compatibility

**No breaking change.** Swift has no default generic parameters, so a bare `getObject(...)` can't infer `T`; the deprecated non-generic overload covers exactly that case:

- ✅ `let x = try await client.getObject(...)` → resolves to the deprecated `AnyCodable` overload (compiles, works, emits a deprecation warning nudging you to annotate).
- ✅ `let x: AnyCodable = try await client.getObject(...)` → unchanged behavior (deprecated overload).
- ✅ `let x: MyType = try await client.getObject(...)` → the generic overload, faithfully decoded.
- Same for `getObjectWithHTTPInfo`.

The deprecation is cleanly removable in a future major version (the `AnyCodable` overloads are an isolated, gated template block).

## 🧪 Test

Verified locally (Swift 5.9) — all three Swift packages build clean after regeneration:

- `clients/algoliasearch-client-swift` → `swift build` ✅ — the four overloads coexist with **no ambiguity**, and **no internal deprecation warning** (Swift suppresses the deprecated→deprecated call from the AnyCodable `getObject` into the AnyCodable `getObjectWithHTTPInfo`).
- `docs/snippets/swift` → `swift build` ✅ — the annotated snippet resolves to the generic overload (no deprecation warning).
- `tests/output/swift` → `swift build --build-tests` ✅ — the annotated `getObjectWithHTTPInfo` request/e2e sites resolve to the generic overload.

Typed decoding flows through `transporter.send<T>` — the same proven path used by `getObjects<T>` / `searchSingleIndex<T>`.

> Source-only PR per repo convention — the generated client/tests/snippets are produced and pushed by CI. Local generated edits were used only to verify compilation and are not committed.

### DIFF
--- a/clients/algoliasearch-client-swift/Sources/Search/Extra/SearchClientExtension.swift
+++ b/clients/algoliasearch-client-swift/Sources/Search/Extra/SearchClientExtension.swift
@@ -738,36 +738,4 @@ public extension SearchClient {
         )
         return try helper.mergeResponses(responses)
     }
-
-    /// Retrieves one record by its object ID as an untyped `AnyCodable`.
-    ///
-    /// Backward-compatible overload of the generic `getObject`: a bare
-    /// `let record = try await client.getObject(...)` resolves here and returns `AnyCodable`.
-    /// Annotate the result to decode into your own type instead, e.g.
-    /// `let record: MyRecord = try await client.getObject(indexName: ..., objectID: ...)`.
-    ///
-    /// - parameter indexName: (path) Name of the index on which to perform the operation.
-    /// - parameter objectID: (path) Unique record identifier.
-    /// - parameter attributesToRetrieve: (query) Attributes to include with the records in the response. (optional)
-    /// - parameter requestOptions: Request options to merge with the transporter requestOptions.
-    /// - returns: AnyCodable
-    func getObject(
-        indexName: String,
-        objectID: String,
-        attributesToRetrieve: [String]? = nil,
-        requestOptions: RequestOptions? = nil
-    ) async throws -> AnyCodable {
-        let response: Response<AnyCodable> = try await self.getObjectWithHTTPInfo(
-            indexName: indexName,
-            objectID: objectID,
-            attributesToRetrieve: attributesToRetrieve,
-            requestOptions: requestOptions
-        )
-
-        guard let body = response.body else {
-            throw AlgoliaError.missingData
-        }
-
-        return body
-    }
 }
--- a/clients/algoliasearch-client-swift/Sources/Search/Extra/SearchClientExtension.swift
+++ b/clients/algoliasearch-client-swift/Sources/Search/Extra/SearchClientExtension.swift
@@ -738,4 +738,36 @@ public extension SearchClient {
         )
         return try helper.mergeResponses(responses)
     }
+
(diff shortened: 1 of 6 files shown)


eric-zaharia (approved):
Looks good!

PR #6598, opened 2026-06-22
fix(javascript): chunked push polling offset and batchSize forwarding

eric-zaharia:
## 🧭 What and Why

🎟 JIRA Ticket: [API-503](https://algolia.atlassian.net/browse/API-503)

Fixes two independent defects in the JavaScript transformation/push helpers that together caused `waitForTasks` to silently under-poll pushed tasks.

1. **Offset over-advance in `chunkedPush`.** The poll cursor was advanced on every object iteration rather than only when a new batch was pushed. Once `responses.length` reached a multiple of `waitBatchSize`, the `responses.length % waitBatchSize === 0` gate stayed true on subsequent non-push iterations, slicing empty ranges and running `offset` past the real responses — so most pushed tasks were never polled. Slicing an empty range is a no-op, so there was no crash, hence "silent". Only manifests with multiple wait windows (large inputs).

2. **`batchSize` dropped by two wrappers.** `saveObjectsWithTransformation` and `partialUpdateObjectsWithTransformation` destructured their options without `batchSize`, so it never reached `chunkedPush` and chunking never happened through those entry points (`replaceAllObjectsWithTransformation` already forwarded it). A single push left nothing for the cursor to overrun, which masked defect 1 in the existing CTS proof.

The fix mirrors the Python reference ([#6486](https://github.com/algolia/api-clients-automation/pull/6486)): gated wait + bounded cursor + post-loop drain.

### Changes included:

- `chunkedPush` (`templates/javascript/clients/client/api/ingestionHelpers.mustache`): extract the polling loop into a `waitForBatch(start, end)` closure; replace the `% waitBatchSize` gate with `responses.length - offset >= waitBatchSize`, advance the cursor by exactly `waitBatchSize`, and drain the remaining tail after the loop.
- Wrappers (`templates/javascript/clients/algoliasearch/builds/definition.mustache`): forward `batchSize` to `chunkedPush` in `saveObjectsWithTransformation` and `partialUpdateObjectsWithTransformation`.
- Unskip the `chunkedPushWait` CTS proof for `javascript` (`tests/CTS/client/search/saveObjectsWithTransformation.json` and the assertion count in `scripts/cts/runCts.ts`).

## 🧪 Test

Covered by the existing `chunkedPushWait` CTS scenario — `saveObjectsWithTransformation` with 25 objects, `waitForTasks: true`, `batchSize: 10` → 3 pushes. The test server asserts `getEventCount === pushCount` **and** that the set of distinct polled `eventID`s equals `pushCount`, i.e. every pushed task is polled exactly once. Before this PR the scenario was skipped for JS; it now runs. PHP remains skipped for this scenario.

[API-503]: https://algolia.atlassian.net/browse/API-503?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Fluf22 (commented):
Only one comment: can you please gate against weird `batchSize` values (zero or negative)?

Fluf22 (approved):

PR #6599, opened 2026-06-22
fix(javascript): expose Node search helpers on `Algoliasearch` type under `bundler` module resolution

### LINKED ISSUE #1611: [bug]: TS missing type union?
### Description

The type **SearchClientNodeHelpers** does not seem to be used in the searchclient typing.
As a result, the **generateSecuredApiKey** method and others are not available in the client typing.

I had to use this workaround:

```typescript
import { Algoliasearch, algoliasearch, type SearchClientNodeHelpers } from "algoliasearch";

type NodeAlgoliaClient = Algoliasearch & SearchClientNodeHelpers;

let algolia = algoliasearch(appId, apiKey) as NodeAlgoliaClient;
algolia.generateSecuredApiKey({...});
```

 

### Client

Search

### Version

5.55.0

### Relevant log output

```shell

```

Fluf22:
## 🧭 What and Why

🎟 fixes algolia/algoliasearch-client-javascript#1611

The reporter notes that `SearchClientNodeHelpers` is not part of the `algoliasearch` client typing, so `generateSecuredApiKey` (and the other Node-only helpers) are missing from the `Algoliasearch` type, even though they work at runtime. Their workaround was `algoliasearch(...) as (Algoliasearch & SearchClientNodeHelpers)`.

### Root cause

The umbrella `algoliasearch` package declares its public type as:

```ts
import type { SearchClient } from '@algolia/client-search';
export type Algoliasearch = SearchClient & { /* init* methods + *WithTransformation helpers */ };
```

The helper surface of `Algoliasearch` is therefore inherited entirely from however the consumer's TypeScript resolves `@algolia/client-search`. But that package exposes a **different `SearchClient` per build**:

- `client-search` **node**/**fetch** build → `ReturnType<…> & SearchClientNodeHelpers` (has `generateSecuredApiKey`, `getSecuredApiKeyRemainingValidity`, `accountCopyIndex`)
- `client-search` **worker** build → `… & SearchClientWorkerHelpers`
- `client-search` **browser** build → `ReturnType<…>` (no helpers)

Under `moduleResolution: "bundler"` (the modern default for Vite and many TS setups), TypeScript does **not** apply the `node` export condition, so both `algoliasearch` and the nested `@algolia/client-search` resolve through the `default` (browser) condition → `SearchClient` without helpers → `Algoliasearch` without `generateSecuredApiKey`. Classic `node`/`node16`/`nodenext` resolution apply the `node` condition (or fall back to the root `index.d.ts` which re-exports `./dist/node`), so they happen to work — which is why this only bites a subset of users.

I reproduced this in isolation with faithful `exports`/`.d.ts` shapes:

| `moduleResolution` | `generateSecuredApiKey` visible on `Algoliasearch`? |
| --- | --- |
| `node` | ✅ |
| `node16` | ✅ |
| `nodenext` | ✅ |
| `bundler` | ❌ `Property 'generateSecuredApiKey' does not exist on type 'Algoliasearch'` |

### Tradeoff / discussion

This is the minimal change that matches the existing design (all four umbrella builds are intentionally generated from a single, identical template). The one consequence is that the `algoliasearch` **browser** type now also advertises the Node helpers, even though they aren't present in a browser runtime. These are server-only operations (e.g. `generateSecuredApiKey` requires the parent/admin key and `node:crypto`), so this is unlikely to affect real browser apps. If maintainers prefer to keep the browser type clean, the alternative is to split `definition.mustache` into per-build templates (node/fetch → `SearchClientNodeHelpers`, worker → `SearchClientWorkerHelpers`, browser → none) — but note that would **not** fix the reported case, because `bundler` resolution always lands on the browser `.d.ts`. The standalone `@algolia/client-search` package has the same class of `bundler`-resolution gap for direct `SearchClient` consumers; that's intentionally left out of scope here.

<div><a href="https://cursor.com/agents/bc-a0f0256f-757b-4177-9b9e-9fe17b385820"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-a0f0256f-757b-4177-9b9e-9fe17b385820"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/clients/algoliasearch-client-javascript/packages/algoliasearch/builds/browser.ts
+++ b/clients/algoliasearch-client-javascript/packages/algoliasearch/builds/browser.ts
@@ -29,6 +29,7 @@ import type {
   ReplaceAllObjectsOptions,
   ReplaceAllObjectsWithTransformationResponse,
   SaveObjectsOptions,
+  SearchClientNodeHelpers,
 } from '@algolia/client-search';
 import type { WatchResponse } from '@algolia/ingestion';
 
@@ -46,7 +47,10 @@ import type {
 
 export * from './models';
 
-export type Algoliasearch = SearchClient & {
+// `SearchClientNodeHelpers` is intersected explicitly: depending on the consumer's `moduleResolution`
+// (e.g. `bundler`), `@algolia/client-search` can resolve to its browser build where `SearchClient`
+// omits the Node-only helpers, which would otherwise drop `generateSecuredApiKey` & co. from this type.
+export type Algoliasearch = SearchClient & SearchClientNodeHelpers & {
   initAbtesting: (initOptions: InitClientOptions & AbtestingRegionOptions) => AbtestingClient;
   initAbtestingV3: (initOptions: InitClientOptions & AbtestingV3RegionOptions) => AbtestingV3Client;
   initAnalytics: (initOptions: InitClientOptions & AnalyticsRegionOptions) => AnalyticsClient;
--- a/clients/algoliasearch-client-javascript/packages/algoliasearch/builds/fetch.ts
+++ b/clients/algoliasearch-client-javascript/packages/algoliasearch/builds/fetch.ts
@@ -29,6 +29,7 @@ import type {
   ReplaceAllObjectsOptions,
   ReplaceAllObjectsWithTransformationResponse,
   SaveObjectsOptions,
+  SearchClientNodeHelpers,
 } from '@algolia/client-search';
 import type { WatchResponse } from '@algolia/ingestion';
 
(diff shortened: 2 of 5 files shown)


Fluf22:
@shortcuts can I ask for your wiser eye here, please?

Fluf22:
This was not needed: the first thing to do when there is a type/bundler related issue is to check that the bundler actually resolves the right typeset for the given environment

PR #6602, opened 2026-06-23
fix(php): chunked push polling offset

eric-zaharia:
## 🧭 What and Why

🎟 JIRA Ticket: [API-504](https://algolia.atlassian.net/browse/API-504)

Fixes a silent under-poll defect in the PHP transformation push helper that caused `waitForTasks` to under-poll pushed tasks.

**Offset over-advance in `chunkedPush`.** The poll cursor was advanced on every object iteration rather than only when a new batch was pushed. Once `sizeof($responses)` reached a multiple of `$waitBatchSize`, the `sizeof($responses) % $waitBatchSize === 0` gate stayed true on subsequent non-push iterations, slicing empty ranges and running `$offset` past the real responses — so most pushed tasks were never polled. Slicing an empty range is a no-op, so there was no error, hence "silent". Only manifests with multiple wait windows (large inputs).

Unlike JS ([#6598](https://github.com/algolia/api-clients-automation/pull/6598)), PHP's wrappers already forward `batchSize`, so this is a helper-only fix and the bug was never masked in the CTS proof.

The fix mirrors the Python reference ([#6486](https://github.com/algolia/api-clients-automation/pull/6486)): gated wait + bounded cursor + post-loop drain.

### Changes included:

- `chunkedPush` (`templates/php/api.mustache`): extract the polling loop into a `$waitForBatch($start, $end)` closure; replace the `% $waitBatchSize` gate with `sizeof($responses) - $offset >= $waitBatchSize`, advance the cursor by exactly `$waitBatchSize`, and drain the remaining tail after the loop.
- Unskip the `chunkedPushWait` CTS proof for all languages (`tests/CTS/client/search/saveObjectsWithTransformation.json` and the assertion count in `scripts/cts/runCts.ts`).

## 🧪 Test

Covered by the existing `chunkedPushWait` CTS scenario — `saveObjectsWithTransformation` with 25 objects, `waitForTasks: true`, `batchSize: 10` → 3 pushes. The test server asserts `getEventCount === pushCount` **and** that the set of distinct polled `eventID`s equals `pushCount`, i.e. every pushed task is polled exactly once. Before this PR the scenario was skipped for PHP; it now runs for all languages (JS fails until [#6598](https://github.com/algolia/api-clients-automation/pull/6598) merges).

[API-504]: https://algolia.atlassian.net/browse/API-504?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Fluf22 (approved):
Great work

PR #6604, opened 2026-06-23
fix(specs): remove recommend query params overridden by the API

raed667:
## 🧭 What and Why

The Recommend API ignores a few query parameters and applies its own values instead, so listing them as configurable in the spec is misleading. This removes the ones that have no effect.

### Changes included:

- Remove `ranking`, `facetFilters`, and `enableABTest` from `queryParameters` and `fallbackParameters`.
- Remove `enableRules` from `fallbackParameters`. It stays on `queryParameters`, where it's still applied.

## 🧪 Test

Validated with `redocly lint specs/recommend/spec.yml`.


Fluf22:
Hey @raed667 
Do you mean that whatever the API clients will send, the API will override it with another value?

Fluf22 (approved):

PR #6607, opened 2026-06-26
fix(java): add warning in replaceAllObjects for empty objects

eric-zaharia:
## 🧭 What and Why

Stemming from CR-11510: a customer wiped an index by calling `replaceAllObjects` with an empty list. The helper copies the source index's settings/synonyms/rules to a temp index, batches **zero** records into it, then moves the temp index over the source — leaving the original index empty. This happened silently, with no warning.

This adds a warning-level log when `replaceAllObjects` (and its `replaceAllObjectsWithTransformation` sibling) is called with an empty list of objects. **Behavior is unchanged** — the call still proceeds; it only warns.

This is the Java client; the other clients are handled in their own PRs.

### Changes included:

- Log a `WARNING` (via `java.util.logging`) when the `objects` collection is empty, added to the canonical `replaceAllObjects` and `replaceAllObjectsWithTransformation` implementations. All overloads delegate down to these, so every call path is covered: `replaceAllObjects was called with an empty list of objects, which will delete all records currently in the "<indexName>" index.`
- The empty check uses `Collection.isEmpty()` when possible and falls back to `iterator().hasNext()` for non-`Collection` `Iterable`s, so single-use iterables aren't consumed by the check.
- Add a `Warning` note to the `replaceAllObjects` Javadoc so the behavior surfaces in IDE hover/autocomplete at write-time.

### Out of scope

- The public "Replace all records" documentation page lives in the `docs-new` repo and is updated separately.

## 🧪 Test

- Regenerated the Java search client and compiled it (`:algoliasearch:compileJava --rerun-tasks` → BUILD SUCCESSFUL).
- No CTS changes: this only adds a log line and doesn't alter requests or behavior.

### DIFF
--- a/templates/java/api_helpers.mustache
+++ b/templates/java/api_helpers.mustache
@@ -1414,6 +1414,9 @@ public <T> List<BatchResponse> partialUpdateObjects(
  * https://api-clients-automation.netlify.app/docs/custom-helpers/#replaceallobjects for implementation
  * details.
  *
+ * <p><b>Warning:</b> calling this method with an empty {@code objects} list replaces the index with
+ * an empty one, deleting all existing records.
+ *
  * @param indexName The `indexName` to replace `objects` in.
  * @param objects The array of `objects` to store in the given Algolia `indexName`.
  * @throws AlgoliaRetryException When the retry has failed on all hosts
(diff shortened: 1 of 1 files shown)


Fluf22 (approved):
Looks good

PR #6608, opened 2026-06-26
fix(javascript): add warning in replaceAllObjects for empty objects

eric-zaharia:
## 🧭 What and Why

Mirrors #6607 for JavaScript: `replaceAllObjects` and `replaceAllObjectsWithTransformation` now log a warning when called with an empty `objects` list, since the helper would otherwise silently empty the index. Behavior is unchanged — the call still proceeds.

### Changes included:

- Empty-list warning in `replaceAllObjects` (`helpers.mustache`) and `replaceAllObjectsWithTransformation` (`definition.mustache`), plus a `Warning:` JSDoc note on both so it shows on IDE hover.
- The warning goes to the client's `Logger` if it implements the new **optional** `warn` method, and falls back to `console.warn` otherwise (new `logWarning` utility in `client-common`). Existing custom loggers keep compiling; the default null logger doesn't implement `warn`, so out-of-the-box users still see the warning, while anyone with a real logger can route or silence it.
- `createConsoleLogger` implements `warn` (gated at the `Info` threshold), with tests.
- Raised the `client-search` bundlewatch budget 10KB → 11KB (#6682 left it at 9.97KB; the warning code tipped it over).
- Fixed the flaky `with logger` tests in the four `algoliasearch` build variants (browser/node/fetch/worker): each fired an un-awaited real `setSettings` request whose retries raced vitest pool teardown and intermittently failed the `client javascript` CI job with all tests green (it killed this PR's CI twice in a row). The tests now await a mock requester that returns a retryable 500 and assert the transporter's actual `Retryable failure` log — the old inner assertions never executed.

### Usage

```ts
await client.replaceAllObjects({ indexName: "products", objects: [] });
// console.warn: replaceAllObjects was called with an empty list of objects,
// which will delete all records currently in the "products" index.
```


templates/javascript/clients/algoliasearch/builds/definition.mustache
```diff
@@ -156,6 +159,10 @@ export function algoliasearch(
         throw new Error('`transformationOptions` must be set in the client config before calling this method. It defaults to the Ingestion API defaults. See https://www.algolia.com/doc/libraries/sdk/methods/ingestion/');
       }
 
+      if (objects.length === 0) {
+        console.warn(`replaceAllObjectsWithTransformation was called with an empty list of objects, which will delete all records currently in the "${indexName}" index.`);
```
Fluf22: Can we reuse the logger here?

https://github.com/algolia/api-clients-automation/blob/f258a66a9d6d4e46c1ddf732efb8fae98f52ef20/clients/algoliasearch-client-javascript/packages/client-common/src/types/logger.ts#L9

MarioAlexandruDan (approved):
Looks good!

PR #6609, opened 2026-06-26
fix(php): add warning in replaceAllObjects for empty objects

eric-zaharia:
## 🧭 What and Why

Mirrors #6607 for the PHP client: logs a PSR-3 warning when `replaceAllObjects` (or `replaceAllObjectsWithTransformation`) is called with an empty list of objects, which would otherwise copy settings, batch zero records, and move the temp index over the source — silently emptying the index. Behavior is unchanged.

### Changes included:

- Log a PSR-3 `warning` via `Algolia::getLogger()` at the top of `replaceAllObjects` when `$objects` is empty, naming the target index.
- Same warning at the top of `replaceAllObjectsWithTransformation` (after the transformation-transporter guard).
- Add a `Warning` note to the docblock of both methods so the footgun surfaces on IDE hover.

### Usage

```php
$client->replaceAllObjects('my_index', []);
// logs: replaceAllObjects was called with an empty list of objects, which will delete all records currently in the "my_index" index.
```

Existing call sites are unaffected — this only adds a `$logger->warning(...)` when the list is empty.


Fluf22 (approved):
Nice!

PR #6610, opened 2026-06-26
fix(ruby): add warning in replace_all_objects for empty objects

eric-zaharia:
## 🧭 What and Why

Mirrors #6607 for the Ruby client: warns when `replace_all_objects` (or `replace_all_objects_with_transformation`) is called with an empty list of objects, which would otherwise copy settings, batch zero records, and move the temp index over the source — silently emptying the index. Behavior is unchanged.

### Changes included:

- `replace_all_objects` now calls `warn(...)` (stderr) when `objects.empty?`, naming the target index.
- `replace_all_objects_with_transformation` does the same, after the existing `assert_ingestion_transporter!` check.
- Added a YARD `Warning:` note to both method doc blocks so the footgun surfaces on hover/autocomplete.

### Usage

```ruby
# Both of these now emit a stderr warning naming the index, then proceed as before:
client.replace_all_objects("products", [])
client.replace_all_objects_with_transformation("products", [])
```

Existing call sites are unaffected — this only calls `warn` (stderr) when the array is empty.


templates/ruby/search_helpers.mustache
```diff
@@ -118,6 +120,10 @@ def replace_all_objects_with_transformation(
 )
   assert_ingestion_transporter!
 
+  if objects.empty?
+    warn("replace_all_objects_with_transformation was called with an empty list of objects, which will delete all records currently in the \"#{index_name}\" index.")
```
Fluf22: can we reuse the logger somehow?

https://github.com/algolia/api-clients-automation/blob/f258a66a9d6d4e46c1ddf732efb8fae98f52ef20/clients/algoliasearch-client-ruby/lib/algolia/transport/http/http_requester.rb#L4

Fluf22 (approved):

PR #6613, opened 2026-06-26
fix(kotlin): add warning in replaceAllObjects for empty objects

eric-zaharia:
## 🧭 What and Why

Mirrors #6607 for the Kotlin client: warns when `replaceAllObjects` (or `replaceAllObjectsWithTransformation`) is called with an empty list of objects, which would otherwise copy settings, batch zero records, and move the temp index over the source — silently emptying the index. Behavior is unchanged.

### Changes included:

- Added an `objects.isEmpty()` check at the top of `replaceAllObjects` that prints a warning naming the affected index when the list is empty.
- Added the same check to `replaceAllObjectsWithTransformation`, with the message naming that method.
- Added a KDoc `Warning` note to both methods so the footgun surfaces on IDE hover.

### Usage

```kotlin
// Prints a warning, then proceeds (clearing the index as before):
client.replaceAllObjects(indexName = "products", objects = emptyList())
```

Existing call sites are unaffected — this only prints a warning when the list is empty.


clients/algoliasearch-client-kotlin/client/src/commonMain/kotlin/com/algolia/client/extensions/SearchClient.kt
```diff
@@ -519,6 +522,11 @@ public suspend fun SearchClient.replaceAllObjects(
   chunkedOptions: ChunkedHelperOptions =
     ChunkedHelperOptions(maxRetries = DEFAULT_REPLACE_ALL_OBJECTS_MAX_RETRIES),
 ): ReplaceAllObjectsResponse {
+  if (objects.isEmpty()) {
+    println(
+      "replaceAllObjects was called with an empty list of objects, which will delete all records currently in the \"$indexName\" index.",
+    )
```
Fluf22: Can you reuse the logger please?

https://github.com/algolia/api-clients-automation/blob/f258a66a9d6d4e46c1ddf732efb8fae98f52ef20/clients/algoliasearch-client-kotlin/client/src/commonMain/kotlin/com/algolia/client/configuration/ClientOptions.kt#L39

Fluf22 (approved):

PR #6614, opened 2026-06-26
fix(scala): add warning in replaceAllObjects for empty objects

eric-zaharia:
## 🧭 What and Why

Mirrors #6607 for the Scala client: warns when `replaceAllObjects` (or `replaceAllObjectsWithTransformation`) is called with an empty list of objects, which would otherwise copy settings, batch zero records, and move the temp index over the source — silently emptying the index. Behavior is unchanged.

### Changes included:

- `replaceAllObjects`: writes a stderr warning when `objects.isEmpty`, naming the affected index, before the copy/batch/move proceeds; adds a Scaladoc `Warning` note.
- `replaceAllObjectsWithTransformation`: same stderr warning (naming the method) and Scaladoc `Warning` note.

### Usage

```scala
// An empty objects list now logs a warning to stderr before the index is replaced.
client.replaceAllObjects(indexName = "my_index", objects = Seq.empty)
```

Existing call sites are unaffected — this only writes a warning to stderr when the sequence is empty.


clients/algoliasearch-client-scala/src/main/scala/algoliasearch/extension/SearchExtensions.scala
```diff
@@ -429,6 +432,12 @@ trait SearchExtensions extends IngestionExtensions with SecuredApiKeyExtensions
         chunkedOptions: ChunkedHelperOptions =
           ChunkedHelperOptions(maxRetries = ChunkedHelperOptions.DefaultReplaceAllObjectsMaxRetries)
     )(implicit ec: ExecutionContext): Future[ReplaceAllObjectsResponse] = {
+      if (objects.isEmpty) {
+        System.err.println(
+          s"""replaceAllObjects was called with an empty list of objects, which will delete all records currently in the "$indexName" index."""
+        )
+      }
```
Fluf22: Can you reuse the client logger please?

https://github.com/algolia/api-clients-automation/blob/f258a66a9d6d4e46c1ddf732efb8fae98f52ef20/clients/algoliasearch-client-scala/src/main/scala/algoliasearch/config/Logging.scala#L38-L45

Fluf22 (approved):

PR #6615, opened 2026-06-26
fix(swift): add warning in replaceAllObjects for empty objects

eric-zaharia:
## 🧭 What and Why

Mirrors #6607 for the Swift client: warns when `replaceAllObjects` (or `replaceAllObjectsWithTransformation`) is called with an empty list of objects, which would otherwise copy settings, batch zero records, and move the temp index over the source — silently emptying the index. Behavior is unchanged.

### Changes included:

- `replaceAllObjects` now writes a warning to stderr when `objects` is empty, naming the affected index.
- `replaceAllObjectsWithTransformation` does the same, naming itself in the message.
- Both methods gained a `- Warning:` doc-comment note so the footgun surfaces on IDE hover.

The warning uses a non-throwing `FileHandle.standardError.write(Data(...))` so it stays cross-platform (including Linux) and adds no dependency.

Routing through the client's `Logger` facade was tried and reverted: on Apple platforms the facade maps `.warning` to `OSLogType.info`, which is below default Console visibility, and `os.Logger` redacts interpolated strings as `<private>` when no debugger is attached — the warning would never reach the user it targets. It also required making `Logger.warning` public, expanding the Core API surface. Writing to stderr keeps the warning visible on every platform with no API change, matching the deliberate `console.warn` in the JavaScript port (#6608).

### Usage

```swift
// Warns to stderr — the index will be emptied because the list is empty.
try await client.replaceAllObjects(indexName: "products", objects: [Product]())
```

Existing call sites are unaffected — this only writes a warning to stderr when the array is empty.


clients/algoliasearch-client-swift/Sources/Search/Extra/SearchClientExtension.swift
```diff
@@ -586,6 +589,12 @@ public extension SearchClient {
         chunkedOptions: ChunkedHelperOptions = ChunkedHelperOptions(maxRetries: ChunkedHelperOptions
             .defaultReplaceAllObjectsMaxRetries)
     ) async throws -> ReplaceAllObjectsResponse {
+        if objects.isEmpty {
+            let warning =
+                "replaceAllObjects was called with an empty list of objects, which will delete all records currently in the \"\(indexName)\" index.\n"
+            FileHandle.standardError.write(Data(warning.utf8))
```
Fluf22: We have a logger service

eg:
```swift
Logger.loggingService.log(level: LogLevel.info, message: "User-Agent already appended")
```

Fluf22 (approved):

PR #6616, opened 2026-06-26
fix(dart): add warning in replaceAllObjectsWithTransformation for empty objects

eric-zaharia:
## 🧭 What and Why

Mirrors #6607 for the Dart client: warns when `replaceAllObjectsWithTransformation` is called with an empty list of objects, which would otherwise copy settings, batch zero records, and move the temp index over the source — silently emptying the index. Behavior is unchanged. (Dart only exposes the transformation variant.)

### Changes included:

- Adds a runtime warning via `print(...)` when `objects.isEmpty`, naming the affected index.
- Adds a `Warning` doc-comment note to `replaceAllObjectsWithTransformation` so it surfaces on IDE hover.

### Usage

```dart
// Logs a warning naming the index, then proceeds as before:
await client.replaceAllObjectsWithTransformation(
  indexName: 'products',
  objects: [],
);
```

Existing call sites are unaffected — this only prints a warning when the list is empty.


Fluf22 (dismissed):

clients/algoliasearch-client-dart/packages/client_search/lib/src/extension/transformation.dart
```diff
@@ -136,6 +139,12 @@ extension Transformation on SearchClient {
   }) async {
     if (ingestionTransporter == null) throw StateError(_notSetError);
 
+    if (objects.isEmpty) {
+      print(
+        'replaceAllObjectsWithTransformation was called with an empty list of objects, which will delete all records currently in the "$indexName" index.',
+      );
```
Fluf22: It would be better to use the logger if it's available somehow
Can you check this please?

https://github.com/algolia/api-clients-automation/blob/f258a66a9d6d4e46c1ddf732efb8fae98f52ef20/clients/algoliasearch-client-dart/packages/client_core/lib/src/transport/dio/dio_requester.dart#L51-L56

Fluf22 (approved):

PR #6652, opened 2026-07-02
fix(clients): browseSynonyms pagination (CR-11727)

Fluf22:
## What & why

🎟 [CR-11727](https://algolia.atlassian.net/browse/CR-11727)

`SearchClient.browseSynonyms` (Java) incremented an internal page counter but never wrote it back to the request params, so every call re-requested `page=0`. With a full first page (≥1000 synonyms) the stop condition never triggered — the iterator looped and returned duplicate hits until it ran out of memory. Fixed to write the page back before each `searchSynonyms` call (mirrors `browseRules`); the no-arg overload no longer passes `null`.

While auditing the other clients I found a **related bug in PHP**: `SynonymIterator`/`RuleIterator` stopped after the first *full* page (the guard compared the cumulative key to the current page's hit count), silently truncating results to `hitsPerPage`. They now stop only on a non-full page, like every other client. JavaScript, Kotlin, Scala and Swift were already correct.

## Tests

Added a `browseSynonyms` pagination regression test in all six implementing languages (Java, PHP, JavaScript, Kotlin, Scala, Swift): serve a full page 0 then a partial page 1, and assert the client requests pages `0 → 1`, aggregates every hit exactly once, and terminates. All pass locally.

## Notes

- Only the source fix + hand-written tests are included — CI regenerates the client code.
- The Kotlin test lives in `jvmTest` (the client's common test set has no coroutine runner).

[CR-11727]: https://algolia.atlassian.net/browse/CR-11727?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/clients/algoliasearch-client-javascript/packages/algoliasearch/__tests__/algoliasearch.common.test.ts
+++ b/clients/algoliasearch-client-javascript/packages/algoliasearch/__tests__/algoliasearch.common.test.ts
@@ -5,6 +5,7 @@ import {
   DEFAULT_READ_TIMEOUT_BROWSER,
   DEFAULT_WRITE_TIMEOUT_BROWSER,
 } from '@algolia/client-common';
+import type { EndRequest, Requester } from '@algolia/client-common';
 import type { EchoResponse } from '@algolia/requester-testing';
 
 import { browserEchoRequester } from '../../requester-testing/src/browserEchoRequester';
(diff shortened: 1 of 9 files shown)


eric-zaharia (approved):
Looks good!

PR #6722, opened 2026-07-13
fix(swift): try String first when decoding AnyCodable values

MarioAlexandruDan:
## 🧭 What and Why [API-143]

Follow-up to the recurring low Swift numbers in the PR benchmark reports (e.g. #6479, where Swift sits at 378 req/s vs 2361 for JavaScript).

I profiled the benchmark scenario (2000 sequential `search` calls against the CTS mock server) layer by layer with the real client code, in release mode:

| Layer | Per request | Share |
|---|---|---|
| Full Swift client | 0.43 ms | 100% |
| raw `URLSession` round trip alone | 0.24–0.30 ms | ~70% |
| client CPU (encode + transporter + decode) | 0.13 ms | ~30% |
| — of which response decoding | 0.10 ms | ~23% |

Two takeaways:

1. **~70% of each request is `URLSession` itself** — its fixed per-request latency is architectural (the identical request through SwiftNIO's AsyncHTTPClient takes 0.087 ms, matching Node), and can't be fixed without a new dependency. The `macos-14` runner then scales everything down ~6× vs local hardware, which is how Swift lands at 378 req/s.
2. **Almost all of the client CPU cost is response decoding**, dominated by `AnyCodable` for `additionalProperties`: the decode chain tries `Bool` → `Int` → `UInt` → `Double` before `String`, so every string value (by far the most common type in Algolia payloads — hit attributes, highlights, facets) pays for 4 thrown-and-caught `DecodingError`s before decoding successfully.

## 🧪 The changes

1. **Reorder the `AnyDecodable` try-chain**: `String` first, then `Int` before `Bool`. JSON types are mutually exclusive under `JSONDecoder` (a number never decodes as `String` and vice versa, `1`/`0` never decode as `Bool`, `true` never decodes as `Int`), so the order only affects how many attempts fail before the right type is found — no behavior change, on Darwin or on Linux's swift-corelibs-foundation. Verified with a type-preservation suite covering ints/bools/doubles/`UInt.max`/nulls/nested containers.
2. **Drop `.prettyPrinted` from the default `JSONEncoder`**: request bodies shrink ~25%. JSON semantics are identical; CTS body assertions compare re-serialized JSON, so they are unaffected. A custom encoder can still be injected via `CodableHelper.jsonEncoder`.
3. **Remove the dead `timeoutIntervalForResource` write in `URLSessionRequestBuilder.execute`**: `session.configuration` returns a copy, so the write was a no-op (and would have been a data race if it worked). Per-request timeouts still apply through `URLRequest.timeoutInterval`.
4. **Build `nonAdditionalPropertyKeys` as a set literal in `modelObject.mustache`** instead of N `insert` calls per decode.

Measured on the benchmark scenario (release build, M-series macOS, same mock server):

| Metric | Before | After |
|---|---|---|
| decode `SearchResponses<Hit>` | 0.099 ms | 0.058 ms (**-42%**) |
| client CPU without network | 0.127 ms | 0.079 ms (**-38%**) |
| full benchmark loop | ~2 335 req/s | ~2 660 req/s (**+14%**) |

This benefits every response decode in real usage, not just the benchmark.

[API-143]: https://algolia.atlassian.net/browse/API-143?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clients/algoliasearch-client-swift/Sources/Core/Networking/RequestBuilder/URLSessionRequestBuilder.swift
```diff
@@ -29,10 +29,8 @@ open class URLSessionRequestBuilder: RequestBuilder {
     }
 
     @discardableResult
-    open func execute<T: Decodable>(urlRequest: URLRequest, timeout: TimeInterval) async throws
+    open func execute<T: Decodable>(urlRequest: URLRequest, timeout _: TimeInterval) async throws
     -> Response<T> {
-        self.sessionManager.configuration.timeoutIntervalForResource = timeout
-
```
Fluf22, line 35: Are you sure this is unused?
This was supposed to be how we handle request timeout, no?

MarioAlexandruDan: Yeah, it was a no-op, session.configuration returns a copy, so writing to it after the session is created does nothing (the value never changes, I checked with a small repro). The actual request timeout is set via request.timeoutInterval in Transporter.send, so nothing is lost here.

Fluf22 (approved):

PR #6732, opened 2026-07-17
feat(specs): add decision (declared winner) to abtesting-v3 ABTest

cdhawke:
PR Stack:
- algolia/go#26540 (merged) - server returns the `decision` (declared winner) on v3 retrieval
- algolia/api-clients-automation#6732 <- this one - adds `decision` to the abtesting-v3 spec + regenerates the client

## What

Add an optional `decision` object to the **abtesting-v3** `ABTest` schema, so it appears on `getABTest` and `listABTests`:

```yaml
decision:
  winnerVariantId: integer   # declared winning variant (control = 1)
  declaredAt: string         # RFC 3339
```

Present: the test has been decided (a winner declared). Absent: no winner declared. Modelled as an object (not a bare flag) so future decision signals, e.g. Bayesian probability-to-be-better, can join the same container, covering both frequentist and Bayesian outcomes.

## Why

The decided state lives only in the abtests settings table server-side, so the dashboard otherwise needs a separate `/settings` request per test to know it, which is prohibitive on the list view. Surfacing it on the A/B test object lets the dashboard read it from the single retrieval call it already makes.

## How

Only the hand-written spec (`specs/abtesting-v3/common/schemas/ABTest.yml`) is committed here; per this repo's workflow, CI regenerates the clients. Verified locally that `yarn cli generate javascript abtesting-v3` produces the expected `Decision` model and `ABTest.decision?: Decision` field. `yarn cli build specs abtesting-v3` passes.

## Dependencies

- algolia/go#26540 (the server field) has merged, so this is unblocked. Kept as a draft on purpose: it is parked for a possible future winner/decision affordance in the dashboard, not required by any shipping product yet.
- Decoupled from the dashboard Evidence status work (algolia/AlgoliaWeb#29384). That PR is frequentist-only and does not consume `decision`; it answers "is there enough data?", which is separate from "was a winner declared?". A future dashboard winner affordance can bump the regenerated `@algolia/abtesting` and consume `decision` then.

preview:
image
(this has been rolled out and is being returned by the api)


cdhawke:
@Fluf22 @MarioAlexandruDan could I get a review on this one please?

Fluf22 (approved):

stevenMevans (approved):

PR #6747, opened 2026-07-20
feat(javascript): Request-ID and Correlation-ID support

eric-zaharia:
## 🧭 What and Why

🎟 [API-516](https://algolia.atlassian.net/browse/API-516) (epic [API-515](https://algolia.atlassian.net/browse/API-515) · [spec](https://algolia.atlassian.net/wiki/spaces/APIC/pages/7352287252))

JavaScript reference implementation of the search clusters' tracing contract, reviewed by Engines before the 10 language ports.

- **Correlation-ID on errors** — every failed call exposes the server's `Correlation-ID` as `error.correlationId` and appended to the message: `Invalid API key (Correlation-ID: …)`. Customers can quote it to support without inspecting raw HTTP. Header absent → errors byte-identical to today.
- **Request-ID on requests** — the transporter mints an 11-char base62 ID once per call and resends the *same* value on every retry/host-fallback attempt, so one log search finds all attempts of one operation. Node builds send the `Request-ID` header; browser/fetch/worker builds (incl. lite) send the `x-algolia-request-id` query parameter, which never triggers a CORS preflight. A caller-supplied ID always wins, and minting happens after the cache/dedup boundary — cache keys and request deduplication are provably unchanged. Covers `request`, `requestWithHttpInfo` and `requestStream`.

Scoped via a `requestIdSupport` generator flag to **search, recommend, composition, algoliasearch (+lite)**; all other clients' generated code is untouched (their bundles grow ~0.25KB because `client-common` carries the plumbing — abtesting/recommend budgets bumped accordingly). Non-breaking: trailing optional params on the error classes, new optional `requestIdChannel` transporter option (`undefined` = off).

## 📋 Behavioral notes

- **Helpers now forward caller `requestOptions` to their internal calls.** The helper docstrings have always documented that `requestOptions` are forwarded to the `waitForTask`/`getTask` polls, and the batch writes inside `chunkedBatch`/`replaceAllObjects` already forwarded them — but the polls (and `replaceAllObjects`' failure-path `deleteIndex`, plus `indexExists` in `accountCopyIndex`) were called bare. They now receive the caller's `requestOptions`, which is also what lets one helper invocation share a single Request-ID. Side effect to be aware of: caller-supplied headers, query parameters and timeouts now apply to those internal requests too.
- **Error messages gain a suffix when the server returns a Correlation-ID.** Per the Engines contract, the CID is both a structured field *and* appended to `error.message`. Signatures and absent-header behavior are unchanged, but consumers matching exact message strings will see the suffix appear based on **server-side** header rollout, not client version — `error.correlationId` is the stable way to read it.

## 🧪 Test

- Unit tests across `client-common` and all `algoliasearch` builds: cache key unchanged, dedup shares one ID, same ID across retries, fresh per call, caller precedence, per-build channel, CID on 4xx/retry-exhaustion/timeout paths, edge `X-Algolia-RequestID` never read.
- Live check against a cloud-native app: the returned Correlation-ID ends with the exact 11-char `request-id` the client sent (suffix echo), and real API errors surface their CID.
- Drive-by: `withHttpInfo.test.ts` was never registered in `vitest.config.ts` — registered and its assertions fixed.

[API-516]: https://algolia.atlassian.net/browse/API-516?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[API-515]: https://algolia.atlassian.net/browse/API-515?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/clients/algoliasearch-client-javascript/package.json
+++ b/clients/algoliasearch-client-javascript/package.json
@@ -27,7 +27,7 @@
     "files": [
       {
         "path": "packages/algoliasearch/dist/algoliasearch.umd.js",
-        "maxSize": "21KB"
+        "maxSize": "21.25KB"
       },
       {
         "path": "packages/algoliasearch/dist/lite/builds/browser.umd.js",
--- a/clients/algoliasearch-client-javascript/package.json
+++ b/clients/algoliasearch-client-javascript/package.json
@@ -39,7 +39,7 @@
       },
       {
         "path": "packages/client-abtesting/dist/builds/browser.umd.js",
-        "maxSize": "6KB"
+        "maxSize": "6.25KB"
       },
       {
         "path": "packages/client-analytics/dist/builds/browser.umd.js",
--- a/clients/algoliasearch-client-javascript/package.json
+++ b/clients/algoliasearch-client-javascript/package.json
@@ -35,7 +35,7 @@
       },
       {
         "path": "packages/abtesting/dist/builds/browser.umd.js",
-        "maxSize": "6KB"
+        "maxSize": "6.25KB"
       },
       {
         "path": "packages/client-abtesting/dist/builds/browser.umd.js",
@@ -75,7 +75,7 @@
       },
       {
         "path": "packages/recommend/dist/builds/browser.umd.js",
-        "maxSize": "6KB"
+        "maxSize": "6.25KB"
       },
       {
         "path": "packages/agent-studio/dist/builds/browser.umd.js",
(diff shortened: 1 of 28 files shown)


xroche (approved):

MarioAlexandruDan (approved):
Ok!

PR #6778, opened 2026-07-24
feat(specs): extend ABTest data model with Bayesian fields

stevenMevans:
## 🧭 What and Why
Support the new `Bayesian` A/B test type. Users will now be able to select at test creation time between two statistical methods frequentist (default) or Bayesian (new), that will change how results are interpreted. Users can also control which type of metrics are returned in their GET responses.

🎟 JIRA Ticket: [OPTIM-2775](https://algolia.atlassian.net/browse/OPTIM-2775)

### Changes included:

- Adds new fields `method` and `primaryMetric` to `ABTestConfiguration` that specifies a test as frequentist (default) or Bayesian (new).
- Adds new `bayesian` optional field to the Get, List and Timeseries responses that contains additional Bayesian metric data when present.
- Adds new multi-value query parameter `methods` to Get, List and Timeseries requests to control if Bayesian (new), frequentist (existing) or both types of metrics are returned in the response.

## 🧪 Test
- Green CI

[OPTIM-2775]: https://algolia.atlassian.net/browse/OPTIM-2775?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/specs/abtesting-v3/common/schemas/Variant.yml
+++ b/specs/abtesting-v3/common/schemas/Variant.yml
@@ -95,11 +95,12 @@ metricResult:
       type: boolean
       description: |
         Whether the pValue is significant or not based on the critical value and the error correction algorithm used.
+    bayesian:
+      $ref: '#/bayesianMetricResult'
   required:
     - name
     - updatedAt
     - value
-    - pValue
   example:
     - name: addToCartCount
       updatedAt: 2025-06-15T15:06:44.400601Z
@@ -121,6 +122,28 @@ metricResult:
       value: 999.66
       pValue: 0.04
       metadata: {'winsorizedValue': 888.8}
+
+bayesianMetricResult:
+  type: object
+  description: |
+    Bayesian inference results for this variant metric.
+    Absent when Bayesian inference isn't available for this metric.
+  properties:
+    probabilityToBeBetter:
+      type: number
+      format: double
+      description: Probability that this variant is better than the control.
+      example: 0.991
+    relativeEffectCILow:
+      type: number
+      format: double
+      description: Lower bound of the 95% credible interval for the relative effect (variant/control minus 1).
+      example: 0.04
+    relativeEffectCIHigh:
+      type: number
+      format: double
+      description: Upper bound of the 95% credible interval for the relative effect (variant/control minus 1).
+      example: 0.39
 
 variantMetadata:
   type: object
--- a/specs/abtesting-v3/common/schemas/ABTest.yml
+++ b/specs/abtesting-v3/common/schemas/ABTest.yml
@@ -66,6 +66,12 @@ ABTestConfiguration:
       $ref: '#/MetricsFilters'
     errorCorrection:
       $ref: '#/ErrorCorrectionType'
+    method:
+      $ref: '#/AnalysisMethod'
+    primaryMetric:
+      type: string
+      description: Public metric name that drives the Bayesian end-test recommendation. Only relevant when `method` is `bayesian`.
+      example: conversionRate
 
 MinimumDetectableEffect:
   type: object
@@ -137,3 +143,10 @@ ErrorCorrectionType:
   enum:
     - bonferroni
     - benjamini-hochberg
+
+AnalysisMethod:
+  type: string
+  description: A/B test statistical analysis method.
+  enum:
+    - bayesian
+    - frequentist


specs/abtesting-v3/common/schemas/Variant.yml
```diff
@@ -95,6 +95,8 @@ metricResult:
       type: boolean
       description: |
         Whether the pValue is significant or not based on the critical value and the error correction algorithm used.
+    bayesian:
+      $ref: '#/bayesianMetricResult'
   required:
     - name
     - updatedAt
```
leonardogavaudan, line 102: Bayesian-only responses omit `pValue`, but the schema still generates it as required:

```yaml
required:
  - name
  - updatedAt
  - value
  - pValue
```

- TypeScript emits `pValue: number`.
- Kotlin emits a non-null `pValue: Double`, which can fail deserialization when the backend omits the field.

Could we make it optional by removing it from `required`?

```yaml
required:
  - name
  - updatedAt
  - value
```

specs/abtesting-v3/paths/timeseries.yml
```diff
@@ -17,6 +17,7 @@ get:
         type: array
         items:
           $ref: '../common/schemas/Timeseries.yml#/MetricName'
+    - $ref: '../common/parameters.yml#/methods'
```
leonardogavaudan: Adding this optional query parameter changes the generated Java positional API. The currently published overload is:

```java
public Timeseries getTimeseries(
  @Nonnull Integer id,
  String startDate,
  String endDate,
  List<MetricName> metric
) throws AlgoliaRuntimeException {
  return this.getTimeseries(id, startDate, endDate, metric, null);
}
```

This PR instead generates:

```java
public Timeseries getTimeseries(
  @Nonnull Integer id,
  String startDate,
  String endDate,
  List<MetricName> metric,
  List<AnalysisMethod> methods
) throws AlgoliaRuntimeException {
  return this.getTimeseries(id, startDate, endDate, metric, methods, null);
}
```

The compatibility check therefore reports the published overload as removed. Existing source code stops compiling, and previously compiled code can fail with `NoSuchMethodError`. The same applies to the list, async, and `WithHTTPInfo` overloads.

There are two implementation options:

- Preserve backward compatibility by retaining the existing overloads and delegating with `methods = null`:

```java
public Timeseries getTimeseries(
  @Nonnull Integer id,
  String startDate,
  String endDate,
  List<MetricName> metric
) throws AlgoliaRuntimeException {
  return this.getTimeseries(id, startDate, endDate, metric, null, null);
}
```

- Accept the compatibility break, mark the commit as `BREAKING CHANGE`, and coordinate a major Java release (not sure how that works).

specs/abtesting-v3/common/schemas/Variant.yml
stevenMevans: Removed from `required` in [faab196](https://github.com/algolia/api-clients-automation/pull/6778/commits/faab19601cc8a2e3f2872eb516dd119986984a11)

specs/abtesting-v3/paths/timeseries.yml
stevenMevans: Should be resolved with [570d549](https://github.com/algolia/api-clients-automation/pull/6778/commits/570d5490a73d40f30b271eef4b4deb6303003f2e), but I did let Claude handle that so I'll make sure the API clients team takes a good look at this specifically.

leonardogavaudan (dismissed):

specs/abtesting-v3/paths/abtest.yml
```diff
@@ -8,6 +8,7 @@ get:
   description: Retrieves the details for an A/B test by its ID.
   parameters:
     - $ref: '../common/parameters.yml#/ID'
+    - $ref: '../common/parameters.yml#/methods'
```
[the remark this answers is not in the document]
stevenMevans: Backed out the `methods` query param from the commit, so this is no longer relevant.

specs/abtesting-v3/paths/timeseries.yml
stevenMevans: I ended up backing out the `methods` query param completely so this shouldn't be relevant any more. It caused issues with most generated clients and doesn't seem worth the hassle. If user's request this functionality then we can look at doing a major release.

PR #6789, opened 2026-07-28
fix(specs): use periods in description fields consistently

kai687:
## 🧭 What and Why

Many sentences in `description` fields didn't end in periods, although more sentences followed. I believe this was introduced in PR #5125 by accident.

🎟 JIRA Ticket: N/A

### Changes included:

- Make sure every sentence ends in a period
- Use periods in list items consistently
- Fix typo in `specs/common/schemas/Batch.yml`

## 🧪 Test

N/A


eric-zaharia (approved):
This is great, thank you!

xroche (commented):
Posted on the wrong PR by mistake, sorry for the noise. The review belongs on #6790 and is now there.

On this PR: the punctuation pass looks right, and the one non-punctuation edit is correct. `partialUpdateObject` with `createIfNoExists: true` does add the record, so "this action adds the record to the index" is accurate. No period landed mid-sentence, which was the real risk in a `|` block scalar where Markdown joins consecutive lines.

Optional, all pre-existing and inside blocks this PR already edits:

- Four sentences still have no period: `attributesToHighlight` and `attributesToRetrieve` ("Attribute names are case-sensitive"), `minProximity` ("by equally scoring matches that are farther apart"), and two lines in `responseFields`.
- Two grammar errors survive on lines this PR rewrote: "considered an exact matches" in `alternativesAsExact`, and "a more broader search" in `similarQuery`.
- `Batch.yml` leaves `- \`delete\`. Delete an index.` with a period where its siblings use `- \`key\`: description`.

ClaraMuller (approved):

PR #6790, opened 2026-07-28
feat(specs): add attributeComputedBy to spec

kai687:
## 🧭 What and Why

Recommended by the field and used by customers, in 2024, the Search team agreed that we should document this, but then nothing came of it. Another [request to the docs channel](https://algolia.slack.com/archives/C0D3A9E82/p1785178281034399) prompted this activity. This was also [requested for the CLI](https://algolia.slack.com/archives/C2V6Z0YUQ/p1785339597577009?thread_ts=1727077730.956409&cid=C2V6Z0YUQ)

🎟 JIRA Ticket: [DOC-1861](https://algolia.atlassian.net/browse/DOC-1861)

### Changes included:

- Add new index setting and enum `attributeCriteriaComputedBy`.
- Deprecate `attributeCriteriaComputedByMinProximity`.

## 🧪 Test

[DOC-1861]: https://algolia.atlassian.net/browse/DOC-1861?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Fluf22 (dismissed):

kai687:
@Fluf22 any idea how we could get review of a team member from the Search API? I pulled the descriptions from internal documents, but would like a review of this before publishing. I tried in `#help-search` but didn't get a reply.

specs/common/schemas/IndexSettings.yml
```diff
@@ -519,6 +524,35 @@ attributeCriteriaComputedByMinProximity:
...
+    - minProximity
+    - best
+    - sum
+  description: |
+    Strategy for computing the Attribute ranking criterion.
+    This mainly affects multi-word queries where matches occur in and across multiple attributes.
+
+    - `minProximity`.
+      Determine the best matching attribute from those with the best proximity score.
+      This is the default when Attribute follows Proximity in the `ranking` setting.
```
xroche: `best` and `minProximity` are not two defaults that a user picks between. The `ranking` order decides which one the engine uses, and it overrides the setting.

`ProximityCalculator.cpp:115` takes the `minProximity` path when Attribute comes after Proximity, whatever the setting holds:

```cpp
if (attributeCriteriaComputedBy == AttributeCriteriaType::minProximity || posAttribute > posProximity ||
    posAttribute == posProximity) {
    firstMatchedWord = computation.minPosition;
}
```

Attribute comes after Proximity in the default `ranking`. So `best` does nothing on an index that keeps the default order, and that is most indices. Measured on a test cluster with the record from our internal ranking guide:

| `ranking` order | unset | `best` | `minProximity` |
|---|---|---|---|
| default, Proximity first | 2000 | 2000 | 2000 |
| Attribute before Proximity | 0 | 0 | 2000 |

2000 means the engine picked attribute `third`. 0 means it picked `first`.

A reader follows this text, sets `best`, and sees nothing change. State the constraint before the list. There is suggested wording in my summary comment.

specs/common/schemas/IndexSettings.yml
```diff
@@ -519,6 +524,35 @@ attributeCriteriaComputedByMinProximity:
...
+    - `best`.
+      Determine the best matching attribute from all attributes matching any query term.
+      This is the default when Attribute comes before Proximity in the `ranking` setting.
+      Ties on an ordered attribute are broken by the match position.
+    - `sum`.
+      Instead of picking one best attribute, add up a score for every query word.
+      Each word's score depends on which attribute it matched (and, for ordered attributes, where in that attribute).
+      Records with a lower total score rank higher.
+      This stops a record from ranking high just because one word of a multi-word query matched in a top attribute.
+      For most use cases, use `sum` when you place Attribute before Proximity in the `ranking` setting, especially for indices with short, relevant attributes.
```
xroche: `sum` does not depend on the `ranking` order. `ProximityCalculator.cpp:120` applies it after the order-dependent branch and overwrites that result. Measured: `sum` returns the same score under both orders.

This sentence sits next to two bullets whose behavior does depend on the order, so it reads as a requirement. Say that it is a relevance preference, or drop the `ranking` part.

specs/common/schemas/IndexSettings.yml
```diff
@@ -519,6 +524,35 @@ attributeCriteriaComputedByMinProximity:
...
+      Determine the best matching attribute from those with the best proximity score.
+      This is the default when Attribute follows Proximity in the `ranking` setting.
+      Ties on an ordered attribute are broken by the match position.
+    - `best`.
+      Determine the best matching attribute from all attributes matching any query term.
+      This is the default when Attribute comes before Proximity in the `ranking` setting.
+      Ties on an ordered attribute are broken by the match position.
+    - `sum`.
+      Instead of picking one best attribute, add up a score for every query word.
+      Each word's score depends on which attribute it matched (and, for ordered attributes, where in that attribute).
```
xroche: A query word that matches nothing is missing here. `ProximityCalculator.cpp:84` adds 65536 for such a word, and one attribute step is 1000. One unmatched word therefore outweighs about 65 attributes and dominates the total.

This matters because `sum` gets used together with `optionalWords` and `removeWordsIfNoResults`. One line covers it: "A query word that matches nothing adds a large penalty."

specs/common/schemas/IndexSettings.yml
```diff
@@ -510,7 +512,10 @@ alternativesAsExact:
 
 attributeCriteriaComputedByMinProximity:
   type: boolean
+  deprecated: true
```
xroche: The description change is good. The `deprecated: true` flag needs a decision from the Search team first.

The engine does not treat this setting as deprecated. Its deprecated-settings list (`IndexSettingsEnums.h:70`) holds 11 entries and this setting is not one of them. The setting still changes behavior and returns no warning. I checked both on a test cluster.

This flag makes every SDK warn on a parameter that the engine fully supports. Who owns that call?

specs/common/schemas/IndexSettings.yml
```diff
@@ -439,6 +439,8 @@ indexSettingsAsSearchParams:
       $ref: '#/sortFacetValuesBy'
     attributeCriteriaComputedByMinProximity:
       $ref: '#/attributeCriteriaComputedByMinProximity'
+    attributeCriteriaComputedBy:
```
xroche, line 442: Three more spec files declare `attributeCriteriaComputedByMinProximity` and need the new parameter too:

- `specs/agent-studio/common/schemas/SearchParameters.yml:174`
- `specs/composition/common/params/Search.yml:34`
- `specs/composition/common/schemas/components/injection/SearchQueryParams.yml:14`

Without them, Composition and Agent Studio expose the deprecated parameter and not its replacement. The engine accepts the new one on both surfaces: `Parameters.h:178` carries the `SearchQuery` and `SubQuery` flags.

xroche (commented):
Thanks for writing this up. I checked every claim against the engine source, then measured each one on a test cluster.

The three bullet definitions are correct, and so are the tie-break rule, the `sum` arithmetic, and the precedence over the old boolean. Three statements around them are wrong. One of them points readers at a setting that does nothing.

Blocking:

1. `best` is inert on an index that keeps the default `ranking` order, and the text presents it as a selectable default (line 540 and line 544).
2. `sum` does not need Attribute before Proximity (line 551).
3. `sum` says nothing about a query word that matches nothing, which dominates the score (line 548).

Also worth settling: the `deprecated: true` flag needs an owner's decision (line 515), and three more spec files need the parameter (line 442).

Suggested description that fixes the three points:

```yaml
  description: |
    Strategy for computing the Attribute ranking criterion.
    This mainly affects multi-word queries with matches in more than one attribute.

    The `ranking` setting decides whether `best` takes effect.
    When Attribute comes after Proximity, which is the default order, the engine always uses the `minProximity` strategy and ignores `best`.
    To select `best`, move Attribute before Proximity in `ranking`.
    The `sum` strategy applies in both orders.

    - `minProximity`.
      Pick the best matching attribute from the attributes that form the best proximity score.
      On an ordered attribute, the match position breaks ties.
    - `best`.
      Pick the best matching attribute from all attributes that match any query word.
      On an ordered attribute, the match position breaks ties.
    - `sum`.
      Add up a score for every query word instead of picking one attribute.
      Each word's score comes from the attribute it matched, and from its position in that attribute when the attribute is ordered.
      A query word that matches nothing adds a large penalty.
      A record with a lower total ranks higher.
      A record therefore cannot rank high only because one word of a multi-word query matched a top attribute.
      Use `sum` with short, relevant attributes, and set long-text attributes to unordered.
```

Two small points. This is the only string enum in the file with no `default:`; `queryType`, `mode`, `removeWordsIfNoResults` and `exactOnSingleWordQuery` all have one, and the engine default is `best`. Fix point 1 first, because `default: best` on its own makes the trap worse. Separately, `title: attribute criteria computed by` uses lowercase words with spaces, which only two other titles in the spec tree do.

CI is red on all 24 client jobs and it is not caused by this PR. They all fail on `withQueryCategorization`. `main` landed `fix(cts): make query categorization e2e assertion data-indep` at 12:37 UTC today and this run was 08:38. A rebase clears it.

specs/common/schemas/IndexSettings.yml
kai687: @Fluf22 what do you think? For the docs it would be clearer to mark the `attributeCriteriaComptuedByMinProximity` as deprecated. It would only affect the docs and the SDKs, as the parameter would continue to be valid in the API itself...

@xroche @Fluf22 I'd say if we're not sure who can make that call, it's safer to just _not_ mark this param as deprecated.

Fluf22: I think if it's not deprecated in the engine, it's best if we don't deprecate it officially in the SDKs/docs, and maybe add some words in the description to incentivise users in using something else?

kai687: Undeprecated the param and updated it's description in [7acbbdf](https://github.com/algolia/api-clients-automation/pull/6790/commits/7acbbdf9a79e1896c3137e81b86b05fede001b12)

Fluf22 (dismissed):

specs/composition/common/params/Search.yml
```diff
@@ -34,6 +34,9 @@ aroundRadius:
 attributeCriteriaComputedByMinProximity:
   $ref: '../../../common/schemas/IndexSettings.yml#/attributeCriteriaComputedByMinProximity'
 
+attributeCriteriaComputedBy:
```
ClaraMuller: This query parameter is not allowed in Composition API.
See official internal documentation if needed: https://algolia.atlassian.net/wiki/spaces/CMP/pages/5294981278/Composition+API+-+Query+Parameters

specs/composition/common/schemas/components/injection/SearchQueryParams.yml
```diff
@@ -13,6 +13,8 @@ baseInjectionQueryParameters:
       $ref: '../../../params/Search.yml#/alternativesAsExact'
     attributeCriteriaComputedByMinProximity:
       $ref: '../../../params/Search.yml#/attributeCriteriaComputedByMinProximity'
+    attributeCriteriaComputedBy:
```
ClaraMuller: Not allowed

ClaraMuller (dismissed):
Invalid documentation for Composition API

kai687:
As a start, I'll limit this PR to the Search API. Can add it to Agent Studio/Recommend later if needed.

Fluf22 (dismissed):

Fluf22 (dismissed):

Fluf22:
@kai687 are we good to merge the whole stack?

Fluf22 (dismissed):

PR #6795, opened 2026-07-30
feat(specs): add analytics v3 pattern endpoints

Niamsy:
## 🧭 What and Why

Adds the semantic-patterns framework (`/3/patterns/*`) endpoints to the existing **analytics** client spec so it gets public clients generated. It's additive, no changes to `/2/*`.
Since these endpoints aren't fully stable yet, they're tagged beta (via the Patterns (Beta) tag + (Beta) summaries). They live in the existing analytics client rather than a dedicated v3 one, since they're additive.

### Changes included:

- `specs/analytics/common/schemas/patterns.yml`: payload/response/catalog schemas (+ `limit`/`offset`)
- `specs/analytics/paths/patterns/`: `getPatternsFields` (GET) + `queryPatterns{Timeseries,Scalar,Table,Distribution}` (POST)
- `specs/analytics/spec.yml`: wire the 5 paths, add `pattern` tag + group, note in version prose

## 🧪 Test
- Green CI


specs/analytics/paths/patterns/queryPatternsScalar.yml
```diff
@@ -0,0 +1,41 @@
...
+
+    Aggregates the requested `metrics` over the whole period and returns a single object keyed by metric
+    kind. `metrics` and `parameters` are required; `filters` is optional. Discover valid field kinds per
+    domain with `/3/patterns/fields`.
+  requestBody:
+    required: true
+    content:
+      application/json:
+        schema:
+          $ref: '../../common/schemas/patterns.yml#/ScalarPayload'
```
leonardogavaudan, line 21: Looks good to me overall, and I don’t think this needs to block the PR. One thing to note:

- These generated methods receive the indices through the request body.
- The current Analytics backend performs index authorization earlier, using the `index` URL query parameter.
- Unrestricted API keys work without that URL parameter.
- With an index-restricted key, the consumer currently also needs to provide `?index=indexA,indexB` through the generic request options.
- For scalar, table, and timeseries, the backend then verifies that the URL and body indices match.

The generated client shape here looks reasonable, but restricted-key callers still have this non-obvious backend constraint.

specs/analytics/paths/patterns/getPatternsFields.yml
```diff
@@ -0,0 +1,43 @@
...
+        x-ratelimit-remaining:
+          $ref: '../../../common/responses/rateLimit.yml#/x-ratelimit-remaining'
+        x-ratelimit-reset:
+          $ref: '../../../common/responses/rateLimit.yml#/x-ratelimit-reset'
+      content:
+        application/json:
+          schema:
+            $ref: '../../common/schemas/patterns.yml#/Catalog'
+    '304':
+      description: Not Modified — the catalog matches the supplied `If-None-Match`.
```
leonardogavaudan: I was looking into the caching behavior here because `If-None-Match` and `304` are different from how the other Analytics endpoints work. I think there may be a gap between what this method exposes and what the generated clients currently support:

- The four query endpoints use backend Redis caching. Consumers always receive `200` and do not manage caching themselves.
- This endpoint’s ETag flow requires the consumer to retrieve the ETag, store both the ETag and catalog, explicitly pass `ifNoneMatch` later, and reuse the stored catalog after a `304`.
- The generated client does not manage those steps automatically.
- The generated JavaScript client does send `If-None-Match` when requested, but its transporter only treats `2xx` responses as successful. It classifies a valid `304` as retryable and eventually throws, including from the HTTP-info method.

Given that the manual ETag flow is not currently supported end to end, would it be better to avoid exposing `If-None-Match` and `304` in the generated method for now?

specs/analytics/common/schemas/patterns.yml
```diff
@@ -0,0 +1,416 @@
...
+    roots:
+      type: array
+      description: Root stages the field depends on. Two fields are combinable in one query only when their roots intersect.
+      items:
+        type: string
+    requires:
+      type: array
+      description: Public ACL identifiers required to read the field, for example `clickAnalyticsEnabled`.
+      items:
+        type: string
```
leonardogavaudan, line 407: I noticed that the requiredness of these response models does not currently match what the Analytics backend serializes:

- The backend always includes `series`, `totals`, `dates`, `date`, `metrics`, `rows`, `domains`, and the catalog arrays because those Go fields do not use `omitempty`.
- `key` is the exception and is genuinely optional.
- The current schemas do not mark the other fields as required, so generated clients model them as optional. For example, JavaScript generates `rows?: ...` and `series?: ...`, even though the backend does not omit those properties.
- There is also a nullability difference for an empty timeseries. The backend currently leaves `series` as a nil slice and serializes `"series": null`, while this schema describes an optional, non-null array.

Could we align the response contract before publishing these models, including deciding whether an empty timeseries should return `[]` or `null` and marking the fields that are always present as required?

specs/analytics/paths/patterns/getPatternsFields.yml
```diff
@@ -0,0 +1,43 @@
...
+    '400':
+      $ref: '../../../common/responses/BadRequest.yml'
+    '402':
+      $ref: '../../../common/responses/FeatureNotEnabled.yml'
+    '403':
+      $ref: '../../../common/responses/MethodNotAllowed.yml'
```
leonardogavaudan: I noticed a difference between the contract described here and how the current Analytics backend registers this endpoint:

- The backend registers `/3/patterns/fields` without the normal authentication or rate-limit middleware. It requires no application ID, API key, or Analytics ACL.
- The handler returns the complete static catalog. Each catalog entry’s `requires` field explains the permissions needed when that field is later used in a query.
- This specification inherits the Analytics API’s global application ID and API-key security and declares `x-acl: analytics`. The generated JavaScript documentation consequently says that the `analytics` ACL is required.
- The response here declares the normal Analytics rate-limit headers, while the backend instead emits `ETag` and `Cache-Control` for this route.

Authenticated consumers can still call the endpoint because the additional credentials are harmless. The difference is that the backend presents it as an anonymous, cache-oriented discovery endpoint, while the generated contract presents it as an authenticated, rate-limited Analytics operation.

leonardogavaudan (dismissed):

specs/analytics/paths/patterns/queryPatternsScalar.yml
Niamsy: Updated the code to add the indices in the query param too (but not required)

specs/analytics/common/schemas/patterns.yml
Niamsy: I updated to add the required to the field responses + made timeseries response nullable (I prefer matching the contract to the go code for now)

specs/analytics/paths/patterns/getPatternsFields.yml
Niamsy: I removed the auth check for this route

leonardogavaudan (approved):

eric-zaharia (approved):
LGTM!

PR #3339, opened 2024-07-09
fix(cts): split requests and e2e tests

millotp:
## 🧭 What and Why

🎟 JIRA Ticket: [DI-2573](https://algolia.atlassian.net/browse/DI-2573)

Split e2e to be able to retry them, or exclude them when running the test suite offline.

[DI-2573]: https://algolia.atlassian.net/browse/DI-2573?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #6878, opened 2026-08-11
feat(dart): Request-ID and Correlation-ID support

MarioAlexandruDan:
## 🧭 What and Why

🎟 JIRA Ticket: [API-520](https://algolia.atlassian.net/browse/API-520)

Dart port of the JS Request-ID/Correlation-ID reference implementation (#6747), completing the ports alongside Go (#6868), C# (#6869), Swift (#6874) and Ruby (#6877).

### Changes included:

- The retry strategy mints an 11-char base62 Request-ID once per execution, before the host loop, and reuses it across retries; a caller-supplied ID wins case-insensitively over request options, operation headers and client default headers (`client_core/lib/src/transport/retry_strategy.dart`, new `request_id.dart`). The default-headers check is a construction-time flag because the default requester snapshots its headers then, and it is gated on the default requester since a custom one never receives `ClientOptions.headers`. On the web the ID is minted as the `x-algolia-request-id` query parameter instead of the header, which is not covered by the API's CORS `Access-Control-Allow-Headers` and would fail the preflight; the channel is a conditional import like the existing platform seam for the Algolia agent, defaulting to the query parameter and selecting the header only on `dart.library.io`, so wasm web builds (where `dart.library.html` is false) stay on the safe channel, and the JS reference does the same through `requestIdChannel` (`request_id_channel_web.dart`, `retry_strategy.dart`, `wait_task.dart`).
- `HttpResponse` gains an optional `headers` field, extending the core transport response type that only carried status and body; equality is untouched (`client_core/lib/src/transport/requester.dart`, populated in `dio_requester.dart`, for error responses only so the success path stays allocation-free). The transport reads it too: a custom requester that returns an error response instead of throwing now gets the Correlation-ID surfaced, with 4xx terminal and every other non-2xx status retried, the same classification as thrown API errors and the JS reference (`retry_strategy.dart`). Behavior note for custom requester authors: on main a returned error response's body was passed through as the success payload; it now throws `AlgoliaApiException`.
- Failed responses expose the Correlation-ID header as `AlgoliaApiException.correlationId` and in its description, read from Dio's response headers at the throw site (`algolia_exception.dart`, `dio_requester.dart`).
- The waitTask/waitAppTask/waitKey* helpers derive one shared Request-ID per invocation, and replaceAllObjectsWithTransformation shares one across its search-side calls (the three operationIndex calls and their waitTask polls) while chunkedPush keeps the caller's options untouched, so the Ingestion API never receives the ID, matching the Go and C# ports (`client_search/lib/src/extension/request_id_options.dart`, `wait_task.dart`, `transformation.dart`). The pure-ingestion helpers (saveObjects/partialUpdateObjects WithTransformation) stay ID-free, and the rescue deleteIndex carries the shared ID scrubbed of the caller's timeouts and body, like the Go, C# and Swift rescue paths.
- Enabled for search, recommend, composition and the algoliasearch umbrella through the shared `Helpers.requestIdSupport` predicate (`AlgoliaDartGenerator.java`, `templates/dart/api.mustache`).
- Minting is caller-overridable through the new `ClientOptions.requestIdEnabled` flag, mirroring the Go and C# ports: null keeps the generated client's own setting, false disables minting on the transport and in the waitTask helpers (`client_options.dart`, `retry_strategy.dart`, `wait_task.dart`). The generated constructors rebuild ClientOptions for the retry strategy, so the template forwards the flag explicitly (`templates/dart/api.mustache`).
- CTS: dart unskipped on the client tests, expected error suffix registered, `REQUEST_ID_LANGUAGES` gains dart, and the recorder assertions get a separate helper count since dart has no saveObjects helper to run the helper test with (`scripts/cts/testServer/requestId.ts`, `scripts/cts/runCts.ts`).
- Fixes two latent gaps in the dart test templates: the client-suite method partial silently dropped request option headers and query parameters (`templates/dart/tests/client/method.mustache`), and the requests-suite header assertion is exhaustive deep-equality so it now accepts a well-formed minted Request-ID when none is pinned (`templates/dart/tests/requests/requests.mustache`, `tests/output/dart/lib/src/expect.dart`).

The requests half of the shared getComposition caller-precedence tests now runs on dart (the skip only had an e2e justification), with the requests template gaining `absentHeaders` rendering so the query-parameter test keeps its no-minted-header assertion like the csharp, javascript, kotlin and scala templates (`templates/dart/tests/requests/requests.mustache`, `tests/output/dart/lib/src/expect.dart`).

Not covered on purpose: the e2e CID-suffix-echo assertion. Dart has no e2e suite at all — a gap left by the 2024 requests/e2e split (#3339) — so the acceptance criterion cannot be met without building that infrastructure first, and the codegen tripwire that protects other languages is a no-op for dart. The e2e tests stay skipped for dart and the deviation is called out here for the ticket owner.

## 🧪 Test

- `dart test test/manual/request_id_test.dart` passes with 24 new tests (format and strict uniqueness, retry reuse, freshness, caller/operation/default-header precedence, disabled clients, the requestIdEnabled opt-out at transport and helper level plus the opt-in on a non-supporting client, waitTask sharing and caller precedence, returned-error Correlation-ID surfacing, the web query-parameter channel with caller precedence, the transformation helper sharing one search-side ID with ingestion staying ID-free and its rescue delete keeping that ID while dropping the caller timeouts, exception and response surface). The dart CTS expected error now pins the full message like the other languages.
- `yarn cli cts run dart --no-e2e` green with the requestId client tests enabled, including the split recorder assertions and the ingestion no-leak check.

**Changelog note:** custom requesters that return a non-2xx `HttpResponse` instead of throwing now get an `AlgoliaApiException` (4xx terminal, other statuses retried) rather than the error body as the success payload; this only affects the `custom*` methods behind a user-implemented `Requester` or an interceptor that resolves error responses.

[API-520]: https://algolia.atlassian.net/browse/API-520?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MarioAlexandruDan:
Sign-off on the behavior change flagged in review: a custom Requester that returns a non-2xx HttpResponse instead of throwing now raises AlgoliaApiException (4xx terminal, other statuses retried) rather than having the error body passed through as the success payload. Keeping it on purpose: the old pass-through was undocumented, fed error bodies straight into deserialization for typed operations anyway, and the new path is what surfaces the Correlation-ID for custom requesters. It is called out in the PR description under the HttpResponse bullet.

eric-zaharia (approved):

PR #6883, opened 2026-08-11
feat(specs): remove migratedAbTestID from abtesting-v3

cdhawke:
## 🧭 What and Why

🎟 JIRA Ticket: [OPTIM-2329](https://algolia.atlassian.net/browse/OPTIM-2329)

 [algolia/go#26972](https://github.com/algolia/go/pull/26972) is merged and deployed

**Breaking:** removes `migratedAbTestID` from the `abtesting-v3` `ABTest` schema, so it disappears from all eight generated clients.

The field exists only for the one-off US→EU A/B test region migration, which finished in 2024. Measured on `abtests-prod-europe-west3`:

| Measure | Value |
|---|---|
| Rows carrying a value | 1,408 across 226 apps |
| Newest `created_at` | **2024-03-05** |
| Still `active` | 0 |
| Same on `abtests-prod-us-east1` | 0 |

Nothing has been written for 2.4 years, and nothing reads it: the dashboard's last reference went in [AlgoliaWeb#29783](https://github.com/algolia/AlgoliaWeb/pull/29783), and the API stops emitting it in [algolia/go#26972](https://github.com/algolia/go/pull/26972).

### Changes included:

- Removes the `migratedAbTestID` property from `ABTest` and the `MigratedABTestId` schema it referenced.

Deliberately a removal rather than a deprecation. A deprecation marker promises removal, and there is no migration path to signal — the field has no successor and no remaining consumer, so the notice would only add noise to eight clients' typedefs.

## 🧪 Test

Green CI. Spec-only, so no generated code is included, per the CI note that generated code is pushed at the end of the run.

Locally verified with `yaml.safe_load`: the `MigratedABTestId` schema and the `migratedAbTestID` property are both gone, no dangling `$ref` remains, `ABTest.required` is unchanged, and the property count drops from 10 to 9.

Needs a release note calling out the removal.

[OPTIM-2329]: https://algolia.atlassian.net/browse/OPTIM-2329?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Niamsy (approved):

cdhawke:
@Fluf22 @MarioAlexandruDan could you have a look at this when you get some time? 🙏🏻

Fluf22 (approved):

PR #6949, opened 2026-08-25
fix(csharp): restore ReadTimeout enforcement dropped in 7.37.0

MarioAlexandruDan:
## 🧭 What and Why

Since 7.37.0 the whole request lifetime is capped by the connect timeout alone, so ReadTimeout and WriteTimeout are silently ignored. This came from #5886, which changed `SetTimeout(requestTimeout + connectTimeout)` to `SetTimeout(connectTimeout)` in AlgoliaHttpRequester. Reported by a customer in CR-12049.

🎟 JIRA Ticket: [CR-12049]

### Changes included:

- The phase up to the response headers now gets the connect budget plus the read budget (AlgoliaHttpRequester.cs:83). HttpClient on netstandard cannot bound the connect phase alone, so this is the closest match to the other clients.
- The body read gets its own fresh read budget via ResponseHeadersRead and a linked cancellation source (AlgoliaHttpRequester.cs:89-99), like the JS requester does once headers arrive.
- Mid-body connection drops map to a network error for the retry strategy (AlgoliaHttpRequester.cs:178), and the retry log prints the measured elapsed time (HttpTransport.cs:275).
- New test server routes /1/test/delayed-headers/:ms and /1/test/stalled-body, plus updated and new timeout integration tests including the customer scenario.

## 🧪 Test

- TimeoutIntegrationTests: 5/5 pass. ReadTimeoutHonoredForSlowResponses fails on the previous requester (times out at ~140ms with ConnectTimeout 100ms and ReadTimeout 2s) and passes with this change.
- Handwritten suite minus e2e: 1116/1116 pass. `CI=true yarn cli cts run csharp --no-e2e --no-requests` exits 0, including the server-side assertValidTimeouts checks.

[CR-12049]: https://algolia.atlassian.net/browse/CR-12049


clients/algoliasearch-client-csharp/algoliasearch/Http/AlgoliaHttpRequester.cs
```diff
@@ -40,8 +42,8 @@ public AlgoliaHttpRequester(ILoggerFactory loggerFactory)
   /// Send request to the REST API
   /// </summary>
   /// <param name="request">Request</param>
-  /// <param name="requestTimeout">Request timeout</param>
-  /// <param name="connectTimeout">Connect timeout</param>
+  /// <param name="requestTimeout">Read/write timeout: budget to receive the response once a connection is possible, applied again to the body read</param>
+  /// <param name="connectTimeout">Connect timeout: extra budget granted on top of <paramref name="requestTimeout"/> to establish the connection</param>
```
Fluf22: It's the other way around: first you grant some time for connection, then for the read/write

clients/algoliasearch-client-csharp/algoliasearch/Http/AlgoliaHttpRequester.cs
```diff
@@ -76,7 +76,9 @@ public async Task<AlgoliaHttpResponse> SendRequestAsync(
 
     httpRequestMessage.Headers.Fill(request.Headers);
 
-    httpRequestMessage.SetTimeout(connectTimeout);
+    // TimeoutHandler already cancels the request via a linked token. HttpClient cannot
+    // split connect vs read, so the whole request gets both budgets.
```
Fluf22: Comment should be clearer about why we do this so we don't revert it again in the future

```suggestion
    // TimeoutHandler is a single overall deadline. We can't split connect vs
    // request timeouts because we target netstandard2.0 (no SocketsHttpHandler),
    // so we sum both.
```

clients/algoliasearch-client-csharp/algoliasearch/Transport/HttpTransport.cs
```diff
@@ -272,11 +272,11 @@ private async Task<TResult> ExecuteRequestAsync<TResult, TData>(
           if (_logger.IsEnabled(LogLevel.Information))
           {
             _logger.LogInformation(
-              "Retry {RetryCount}/{MaxRetries}: Timeout on {Host} after {ConnectTimeout}ms",
+              "Retry {RetryCount}/{MaxRetries}: Retryable failure on {Host} after {Elapsed}ms",
               attemptNumber,
               maxAttempts - 1,
               host.Url,
-              (int)connectTimeout.TotalMilliseconds
+              requestStopwatch.ElapsedMilliseconds
```
Fluf22: Unnecessary change, this is not the goal of this log, nor even the goal of the PR

Here, we want report on the different retry configs for debug purposes

tests/output/csharp/src/LoggingTests.cs
```diff
@@ -235,8 +235,8 @@ await CreateTransportWithHosts(NonRoutableIp, NonRoutableIp2)
 
     AssertLogMatches(
       LogLevel.Information,
-      @"Retry \d+/\d+: Timeout on .+ after \d+ms",
-      @"INFO log should match ""Retry {N}/{MAX}: Timeout on {HOST} after {TIMEOUT}ms"""
+      @"Retry \d+/\d+: Retryable failure on .+ after \d+ms",
+      @"INFO log should match ""Retry {N}/{MAX}: Retryable failure on {HOST} after {ELAPSED}ms"""
```
Fluf22: Revert this too

Fluf22 (approved):

PR #6957, opened 2026-08-28
docs(specs): declare the Correlation-ID response header

xroche:
## 🧭 What and Why

🎟 JIRA Ticket: n/a

Search, Recommend and Composition return a `Correlation-ID` header on every response the engine handles, and no spec declared it, so the REST API reference omits it. #6820 added the `cid` log field, which points at that header, and @Fluf22 asked for the declaration there.

`specs/common/responses/correlationId.yml` holds the definition and copies the shape of `rateLimit.yml`. 56 responses reference it directly, and five shared response files carry it for another 22.

## Scope

Only these three APIs answer on `{appId}.algolia.net`, so my "every API" claim on #6820 was too broad. Left out:

- Error responses, because the engine sends no header when it refuses a request before a handler runs. AlgoliaSaaS #14953 fixes that.
- Client helpers, because one helper call spans several requests.
- `customRequest.yml`, because all 13 APIs share it.
- Agent Studio, because it answers on the same host but returned nothing I could check.

## How

No lint rule checks that a new engine operation declares the header, so that is worth a follow-up.

No test can assert the header today. The CTS `correlationIdSuffix` key asserts that the response repeats the `Request-ID` the client sent, and Search clusters do not repeat it. Codegen produces no client changes.

### DIFF
--- a/specs/common/responses/CreatedAt.yml
+++ b/specs/common/responses/CreatedAt.yml
@@ -1,4 +1,8 @@
+# Reused by Search and Recommend only. Drop the Correlation-ID header below before reusing this response for an API that does not send it.
 description: OK
+headers:
+  Correlation-ID:
+    $ref: 'correlationId.yml#/Correlation-ID'
 content:
   application/json:
     schema:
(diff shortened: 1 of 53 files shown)


ClaraMuller (changes requested):
Can you create a common interface like we do for search query parameter here: https://github.com/algolia/api-clients-automation/blob/main/specs/composition/common/params/Search.yml

Ideally a new file with an interface linked to correlation id that all composition files are referencing.

ClaraMuller (approved):
Thanks for the update on Comp API side

Fluf22 (approved):

PR #6965, opened 2026-09-01
feat(javascript): wait and retry on 429 with Retry-After

MarioAlexandruDan:
## 🧭 What and Why

🎟 JIRA Ticket: [API-545](https://algolia.atlassian.net/browse/API-545)

HTTP 429 is not a host-failover case. JavaScript now waits on the **same host** and retries: `Retry-After` as a positive whole number of seconds, waited in full per the spec (no client-side max wait), otherwise 1s; `maxRateLimitRetries` defaults to 3 (`0` fails on the first 429). 5xx / timeout / network retries are unchanged. Other languages skip the new CTS until sibling tickets.

### Changes included:

- Hand-written JS transporter (`client-common`) intercepts 429 before the 4xx-not-retryable path, waits with `setTimeout`, and re-queues the same host
- `requestStream` waits and retries on 429 using `StreamRequestError` (Retry-After from node-http and fetch requesters)
- Shared CTS mock on ports 6697 / 6698 plus `createClient` plumbing for `maxRateLimitRetries`; the retry-after route replies with a 2s wait so the CTS can tell header parsing apart from the 1s fallback
- JS-only CTS via `onlyLanguages: ["javascript"]` (not a denylist of current languages)
- JSDoc on `maxRateLimitRetries` and a retry-strategy note pointing at the JavaScript transporter

Deferred: docs-new customize page (other repo), live e2e (no staging rate limit)

## 🧪 Test

- `yarn test` in `clients/algoliasearch-client-javascript/packages/client-common` — 20 files, 152 tests
- Requester tests: node-http 17 passed, fetch 14 passed
- `CI=1 yarn cli cts generate javascript`
- `CI=1 yarn cli cts run javascript --no-e2e` — 27 files, 1123 tests

[API-545]: https://algolia.atlassian.net/browse/API-545?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


eric-zaharia (approved):
LGTM!

PR #7029, opened 2026-09-11
fix(specs): clarify addItemKey description in injectedItem metadata

charlottetaylor-oss:
## 🧭 What and Why

Clarify the `addItemKey` behavior

## 🧪 Test

Green CI


gavinwade12 (approved):
LGTM - thanks!

PR #7066, opened 2026-09-17
fix(specs): BREAKING CHANGE - make AB test pValue optional

stevenMevans:
## 🧭 What and Why

The A/B Testing v3 API has allowed `metricResult.pValue` to be omitted for some time when no Frequentist result is available. The OpenAPI spec still marks it as required, so generated clients do not accurately represent the existing API behavior and may reject valid responses.

This PR brings the spec and generated public model types back in line with the runtime contract. Bayesian tests exposed the mismatch because Bayesian-only responses omit Frequentist inference fields, but the mismatch predates that work and is not specific to Bayesian testing.

### Changes included:

- Makes `metricResult.pValue` optional.
- Documents when the API omits `pValue`.

This changes generated public model types from required to optional or nullable across client languages and should be handled as a breaking client release.

## Changelog

`metricResult.pValue` is now optional in generated A/B Testing v3 clients, matching API responses where no p-value is available for a metric. This is a breaking type change: code that assumes `pValue` is always present must handle an absent or nullable value.

## 🧪 Tests

Green CI


Fluf22 (dismissed):

stevenMevans:
@Fluf22 thank you for all the reviews! :pray: Because the Github stack required rebasing after each PR merge, I've merged the stack into this single feature branch that will need your sign-off one more time.

Fluf22 (approved):

PR #7073, opened 2026-09-22
feat(specs): add `externalProvider` source for CompAPI client

ClaraMuller:
## 🧭 What and Why

🎟 JIRA Ticket: https://algolia.atlassian.net/browse/CMP-897

Adds public API support for referencing external provider (e.g. Retail Media Network) configurations from Composition API sources, per the RFC [External Provider Sources in Composition API](https://algolia.atlassian.net/wiki/spaces/CMP/pages/7358120173/RFC+-+External+Provider+Sources+in+Composition+API).

Previously, compositions could only retrieve results via Algolia `search` & `recommend`. This lets customers delegate result retrieval to a configured external provider instead.

### Changes included:

- Add a new `externalProvider` source type, usable as both the main source and an injected-item source (`index`, `configurationID`, `configurationParams`, `params`, `ordering`)
- Add `externalProvider.configurationParams` to the Run request body, to override provider parameters at query time
- Add an `errors` field to the Run response, to surface non-critical processing errors (e.g. a provider failure that fell back to another result set)

## 🧪 Test

- `yarn cli build specs composition` bundles and lints cleanly with the new schemas
- Added CTS fixtures in `tests/CTS/requests/composition/{putComposition,putCompositionRule,search}.json`.


specs/composition/common/schemas/components/injection/ExternalProviderSource.yml
```diff
@@ -0,0 +1,5 @@
+externalProviderOrdering:
+  description: |
```
Fluf22, line 3: We are missing a `type: string` here, right?

Fluf22:
Don't worry about the breaking change warning in the comment: it's coming from an older commit being included in the code generation commit

specs/composition/common/schemas/components/injection/ExternalProviderSource.yml
ClaraMuller: Good catch 👍

specs/composition/common/schemas/components/injection/ExternalProviderSource.yml
```diff
@@ -0,0 +1,6 @@
+externalProviderOrdering:
+  type: string
+  description: |
+    Ordering to apply on the items retrieved from the external provider. 'default' uses the relevance ranking from the Algolia retrieval step. 'providerDefined' uses the ordering returned by the external provider.
+  enum: ['default', 'providerDefined']
+  default: 'providerDefined'
```
ben-kalmus: Should we also add `userDefined` as it's technically allowed?

ben-kalmus (approved):

specs/composition/common/schemas/components/injection/ExternalProviderSource.yml
ClaraMuller: TBC with @aiordache but for the externalProvider source, `providerDefined` replace `userDefined`.
So it is not allowed

ClaraMuller: > `providerDefined`: Uses the ordering returned by the external provider. This corresponds to the `userDefined` ordering for `external` source types.

PR #7074, opened 2026-09-23
refactor(specs): extract source object from behaviour object

ClaraMuller:
## 🧭 What and Why

🎟 JIRA Ticket:

### Changes included:

- List changes


specs/composition/common/schemas/components/behaviour/injection/InjectedItemSource.yml
```diff
@@ -0,0 +1,109 @@
...
+    search:
+      $ref: '#/injectedItemSearch'
+  required:
+    - search
+
+# Split out from injectedItemSearchSource because the custom
+# automation-custom/out-of-line-all-of lint rule requires allOf to be a
+# top-level schema, not nested inside `properties`.
+injectedItemSearch:
+  title: injectedItemSearch
```
gavinwade12: Isn't the `title` redundant when it matches the component's name?
```suggestion
```

specs/composition/common/schemas/components/behaviour/injection/InjectedItemSource.yml
```diff
@@ -0,0 +1,109 @@
...
+    external:
+      $ref: '#/injectedItemExternal'
+  required:
+    - external
+
+# Split out from injectedItemExternalSource because the custom
+# automation-custom/out-of-line-all-of lint rule requires allOf to be a
+# top-level schema, not nested inside `properties`.
+injectedItemExternal:
+  title: injectedItemExternal
```
gavinwade12: same for this + the other sources in this file

gavinwade12 (commented):
I don't quite understand the breaking changes message - are there actually any, and if so, are they just for `java@21`?

specs/composition/common/schemas/components/behaviour/injection/InjectedItemSource.yml
ClaraMuller: Good question 🤔

